PR #1, opened 2017-09-01
Historical queries

cloudhead:
The historical queries feature is separated in two new datatypes: `VersionedTree` and `orphaningTree`.
The former keeps track of multiple of the later, and is the API for working with historical queries. The later is not exposed, and is a wrapper around `IAVLTree` which has the additional feature of keeping track of orphaned nodes.

This implementation of historical queries allows for gaps in history: it's possible to delete intermediary versions. At all times, we attempt to keep the minimal set of orphans required, since version `N` might still be depending on nodes from multiple versions `N - M`.

There is currently no automatic pruning of the tree, since the pruning strategy is to be determined by the user of the tree, ex: keep every 10 trees on disk (10, 20, 30...), or every 100 trees (100, 200, 300), or only keep the last 20 (1, 2, 3, 4...20).

The main new operations are: `SaveVersion(uint64)` and `DeleteVersion(uint64)`. There are some deliberate constraints mostly for simplicity reasons:

* You can't delete the last saved version.
* You can't save the same version twice.
* Version numbers must be > 0.
* You can't save with an empty tree (nil root).

When a new version is saved, the previous one is copied, and acts as the "working tree". When `SaveVersion` is called, it is that working tree that is saved, and the new nodes are marked with the given version.

I've added a test flag: `-test.leveldb` to run certain tests with leveldb, since there is leveldb-specific code for prefix iteration.

I've also added a `-test.fuzz-iterations` flag which takes a number of iterations to run for the fuzz test. A good number would be `-test.fuzz-iterations 2000000` which completes in 15s on my computer.

I haven't benchmarked the code yet, but the original set of operations should perform similarly. Save and Delete might be slower currently, as very little optimization has been done.

An example database which had 3 versions (`1 2 3`), where the 2nd was deleted looks like this:

```
roots/1: a7923cdb7d6b680ad4b9278959dc98d8389fe2c2
roots/3: b388298b512229ed1c1285f18a3ecfdad62dda4c

orphans/1/1/1d950b27a03b2cdff72a588ad38819755f28a3d8: 1d950b27a03b2cdff72a588ad38819755f28a3d8
orphans/1/1/a7923cdb7d6b680ad4b9278959dc98d8389fe2c2: a7923cdb7d6b680ad4b9278959dc98d8389fe2c2
orphans/1/1/8c255924ea9620ccab7247d46752bb52bf0d8ab3: 8c255924ea9620ccab7247d46752bb52bf0d8ab3
orphans/1/1/cc9d4d0a99d84dd89364811d9bfb7681a04209c0: cc9d4d0a99d84dd89364811d9bfb7681a04209c0
orphans/1/1/202d2c0f930fc88496282dd1d69cfb0473e01fd8: 202d2c0f930fc88496282dd1d69cfb0473e01fd8

nodes/1d950b27a03b2cdff72a588ad38819755f28a3d8: key0 = val0             h=0 version=1
nodes/2f348898cd0411816c89aa7c81d0e74196dc14c5: key0 = val2             h=0 version=3
nodes/8c255924ea9620ccab7247d46752bb52bf0d8ab3: key1 = val0             h=0 version=1
nodes/a7923cdb7d6b680ad4b9278959dc98d8389fe2c2: key1                    h=2 version=1
nodes/cc9d4d0a99d84dd89364811d9bfb7681a04209c0: key2                    h=1 version=1
nodes/202d2c0f930fc88496282dd1d69cfb0473e01fd8: key2 = val0             h=0 version=1
nodes/b388298b512229ed1c1285f18a3ecfdad62dda4c: key2                    h=2 version=3
nodes/339726bcc7814d2e983b7e799b2983bffa56e240: key2 = val2             h=0 version=3
nodes/2ffa1c0916cd58cbfc3a05b0a0c5f4b1150cb105: key3 = val1             h=0 version=2
nodes/c4fb4590d73dc65505d4859f3ff6bf43ff61d109: key3                    h=1 version=3
```

I implemented a fuzz tester to catch more bugs. This enabled me to fix a few edge cases I hadn't foreseen regarding orphan functionality. The output shows you the steps required to reproduce the panic:

```
iavl_tree_fuzz_test.go:121: Error after 107227 iterations (size 13):
Program panicked with: Panicked on a Sanity Check: Value missing for key e214af93bcf3e3dbda8569421e55b2436a6fdbeb

                   SET      L        c
                   SET      J        s
                   SAVE                       1
                   REMOVE   8
                   SET      i        D
                   SAVE                       2
                   SET      U        J
                   SAVE                       3
                   DELETE                     1
                >> SET      L        f
                   DELETE                     1
                   SAVE                       4
                   SET      2        D
```

### DIFF
--- a/iavl_node.go
+++ b/iavl_node.go
@@ -115,7 +115,6 @@ func (node *IAVLNode) clone() *IAVLNode {
 	}
 	return &IAVLNode{
 		key:       node.key,
-		value:     node.value,
 		height:    node.height,
 		version:   node.version,
 		size:      node.size,
--- a/iavl_node.go
+++ b/iavl_node.go
@@ -110,6 +110,9 @@ func (node *IAVLNode) debugString() string {
 
 // clone creates a shallow copy of a node with its hash set to nil.
 func (node *IAVLNode) clone() *IAVLNode {
+	if node.isLeaf() {
+		cmn.PanicSanity("Attempt to copy a leaf node")
+	}
 	return &IAVLNode{
 		key:       node.key,
 		value:     node.value,
--- a/iavl_node.go
+++ b/iavl_node.go
@@ -351,7 +351,7 @@ func (node *IAVLNode) remove(t *IAVLTree, key []byte) (
 		if bytes.Equal(key, node.key) {
 			return nil, nil, nil, node.value, []*IAVLNode{node}
 		}
-		return node.hash, node, nil, nil, []*IAVLNode{}
+		return node.hash, node, nil, nil, orphaned
 	}
 
 	if bytes.Compare(key, node.key) < 0 {
@@ -362,7 +362,7 @@ func (node *IAVLNode) remove(t *IAVLTree, key []byte) (
 			node.getLeftNode(t).remove(t, key)
 
 		if len(orphaned) == 0 {
-			return node.hash, node, nil, value, []*IAVLNode{}
+			return node.hash, node, nil, value, orphaned
 		} else if newLeftHash == nil && newLeftNode == nil { // left node held value, was removed
 			return node.rightHash, node.rightNode, node.key, value, orphaned
 		}
@@ -382,7 +382,7 @@ func (node *IAVLNode) remove(t *IAVLTree, key []byte) (
 			node.getRightNode(t).remove(t, key)
 
 		if len(orphaned) == 0 {
-			return node.hash, node, nil, value, []*IAVLNode{}
+			return node.hash, node, nil, value, orphaned
 		} else if newRightHash == nil && newRightNode == nil { // right node held value, was removed
 			return node.leftHash, node.leftNode, nil, value, orphaned
 		}
--- a/iavl_node.go
+++ b/iavl_node.go
@@ -414,9 +414,9 @@ func (node *IAVLNode) getRightNode(t *IAVLTree) *IAVLNode {
 	return t.ndb.GetNode(node.rightHash)
 }
 
-// NOTE: overwrites node
-// TODO: optimize balance & rotate
-func (node *IAVLNode) rotateRight(t *IAVLTree) (*IAVLNode, *IAVLNode) {
+// Rotate right and return the new node and orphan.
+func (node *IAVLNode) rotateRight(t *IAVLTree) (newNode *IAVLNode, orphan *IAVLNode) {
+	// TODO: optimize balance & rotate.
 	node = node.clone()
 	l := node.getLeftNode(t)
 	_l := l.clone()
@@ -431,9 +431,9 @@ func (node *IAVLNode) rotateRight(t *IAVLTree) (*IAVLNode, *IAVLNode) {
 	return _l, l
 }
 
-// NOTE: overwrites node
-// TODO: optimize balance & rotate
-func (node *IAVLNode) rotateLeft(t *IAVLTree) (*IAVLNode, *IAVLNode) {
+// Rotate left and return the new node and orphan.
+func (node *IAVLNode) rotateLeft(t *IAVLTree) (newNode *IAVLNode, orphan *IAVLNode) {
+	// TODO: optimize balance & rotate.
 	node = node.clone()
 	r := node.getRightNode(t)
 	_r := r.clone()
--- a/iavl_node.go
+++ b/iavl_node.go
@@ -335,8 +335,8 @@ func (node *IAVLNode) set(t *IAVLTree, key []byte, value []byte) (
 			return node, updated, orphaned
 		} else {
 			node.calcHeightAndSize(t)
-			new, balanceOrphaned := node.balance(t)
-			return new, updated, append(orphaned, balanceOrphaned...)
+			newNode, balanceOrphaned := node.balance(t)
+			return newNode, updated, append(orphaned, balanceOrphaned...)
 		}
 	}
 }
@@ -469,8 +469,8 @@ func (node *IAVLNode) balance(t *IAVLTree) (newSelf *IAVLNode, orphaned []*IAVLN
 	if balance > 1 {
 		if node.getLeftNode(t).calcBalance(t) >= 0 {
 			// Left Left Case
-			new, orphaned := node.rotateRight(t)
-			return new, []*IAVLNode{orphaned}
+			newNode, orphaned := node.rotateRight(t)
+			return newNode, []*IAVLNode{orphaned}
 		} else {
 			// Left Right Case
 			var leftOrphaned *IAVLNode
@@ -478,16 +478,16 @@ func (node *IAVLNode) balance(t *IAVLTree) (newSelf *IAVLNode, orphaned []*IAVLN
 			left := node.getLeftNode(t)
 			node.leftHash = nil
 			node.leftNode, leftOrphaned = left.rotateLeft(t)
-			new, rightOrphaned := node.rotateRight(t)
+			newNode, rightOrphaned := node.rotateRight(t)
 
-			return new, []*IAVLNode{left, leftOrphaned, rightOrphaned}
+			return newNode, []*IAVLNode{left, leftOrphaned, rightOrphaned}
 		}
 	}
 	if balance < -1 {
 		if node.getRightNode(t).calcBalance(t) <= 0 {
 			// Right Right Case
-			new, orphaned := node.rotateLeft(t)
-			return new, []*IAVLNode{orphaned}
+			newNode, orphaned := node.rotateLeft(t)
+			return newNode, []*IAVLNode{orphaned}
 		} else {
 			// Right Left Case
 			var rightOrphaned *IAVLNode
@@ -495,9 +495,9 @@ func (node *IAVLNode) balance(t *IAVLTree) (newSelf *IAVLNode, orphaned []*IAVLN
 			right := node.getRightNode(t)
 			node.rightHash = nil
 			node.rightNode, rightOrphaned = right.rotateRight(t)
-			new, leftOrphaned := node.rotateLeft(t)
+			newNode, leftOrphaned := node.rotateLeft(t)
 
-			return new, []*IAVLNode{right, leftOrphaned, rightOrphaned}
+			return newNode, []*IAVLNode{right, leftOrphaned, rightOrphaned}
 		}
 	}
 	// Nothing changed
--- a/iavl_node.go
+++ b/iavl_node.go
@@ -190,6 +190,24 @@ func (node *IAVLNode) getByIndex(t *IAVLTree, index int) (key []byte, value []by
 	}
 }
 
+// Computes the hash of the node without computing its descendants. Must be
+// called on nodes which have descendant node hashes already computed.
+func (node *IAVLNode) _hash() []byte {
+	if node.hash != nil {
+		return node.hash
+	}
+
+	hasher := ripemd160.New()
+	buf := new(bytes.Buffer)
+	if _, err := node.writeHashBytes(buf); err != nil {
+		cmn.PanicCrisis(err)
+	}
+	hasher.Write(buf.Bytes())
+	node.hash = hasher.Sum(nil)
+
+	return node.hash
+}
+
 // Hash the node and its descendants recursively. This usually mutates all
 // descendant nodes. Returns the node hash and number of nodes hashed.
 func (node *IAVLNode) hashWithCount() ([]byte, int) {
@@ -199,7 +217,7 @@ func (node *IAVLNode) hashWithCount() ([]byte, int) {
 
 	hasher := ripemd160.New()
 	buf := new(bytes.Buffer)
-	_, hashCount, err := node.writeHashBytes(buf)
+	_, hashCount, err := node.writeHashBytesRecursively(buf)
 	if err != nil {
 		cmn.PanicCrisis(err)
 	}
@@ -209,9 +227,9 @@ func (node *IAVLNode) hashWithCount() ([]byte, int) {
 	return node.hash, hashCount + 1
 }
 
-// Writes the node's hash to the given io.Writer.
-// This function has the side-effect of calling hashWithCount.
-func (node *IAVLNode) writeHashBytes(w io.Writer) (n int, hashCount int, err error) {
+// Writes the node's hash to the given io.Writer. This function expects
+// child hashes to be already set.
+func (node *IAVLNode) writeHashBytes(w io.Writer) (n int, err error) {
 	wire.WriteInt8(node.height, w, &n, &err)
 	wire.WriteVarint(node.size, w, &n, &err)
 
@@ -222,26 +240,30 @@ func (node *IAVLNode) writeHashBytes(w io.Writer) (n int, hashCount int, err err
 		wire.WriteByteSlice(node.value, w, &n, &err)
 		wire.WriteUint64(node.version, w, &n, &err)
 	} else {
-		if node.leftNode != nil {
-			leftHash, leftCount := node.leftNode.hashWithCount()
-			node.leftHash = leftHash
-			hashCount += leftCount
-		}
-		if node.leftHash == nil {
-			cmn.PanicSanity("node.leftHash was nil in writeHashBytes")
+		if node.leftHash == nil || node.rightHash == nil {
+			cmn.PanicSanity("Found an empty child hash")
 		}
 		wire.WriteByteSlice(node.leftHash, w, &n, &err)
-
-		if node.rightNode != nil {
-			rightHash, rightCount := node.rightNode.hashWithCount()
-			node.rightHash = rightHash
-			hashCount += rightCount
-		}
-		if node.rightHash == nil {
-			cmn.PanicSanity("node.rightHash was nil in writeHashBytes")
-		}
 		wire.WriteByteSlice(node.rightHash, w, &n, &err)
 	}
+	return
+}
+
+// Writes the node's hash to the given io.Writer.
+// This function has the side-effect of calling hashWithCount.
+func (node *IAVLNode) writeHashBytesRecursively(w io.Writer) (n int, hashCount int, err error) {
+	if node.leftNode != nil {
+		leftHash, leftCount := node.leftNode.hashWithCount()
+		node.leftHash = leftHash
+		hashCount += leftCount
+	}
+	if node.rightNode != nil {
+		rightHash, rightCount := node.rightNode.hashWithCount()
+		node.rightHash = rightHash
+		hashCount += rightCount
+	}
+	n, err = node.writeHashBytes(w)
+
 	return
 }
 
--- a/iavl_node.go
+++ b/iavl_node.go
@@ -110,11 +110,9 @@ func (node *IAVLNode) debugString() string {
 
 // clone creates a shallow copy of a node with its hash set to nil.
 func (node *IAVLNode) clone() *IAVLNode {
-	if node.isLeaf() {
-		cmn.PanicSanity("Why are you copying a value node?")
-	}
 	return &IAVLNode{
 		key:       node.key,
+		value:     node.value,
 		height:    node.height,
 		version:   node.version,
 		size:      node.size,
--- a/iavl_nodedb.go
+++ b/iavl_nodedb.go
@@ -28,25 +28,26 @@ var (
 )
 
 type nodeDB struct {
-	mtx        sync.Mutex               // Read/write lock.
-	cache      map[string]*list.Element // Node cache.
-	cacheSize  int                      // Node cache size limit in elements.
-	cacheQueue *list.List               // LRU queue of cache elements. Used for deletion.
-	db         dbm.DB                   // Persistent node storage.
-	batch      dbm.Batch                // Batched writing buffer.
-	versions   map[uint64]bool
-	latest     uint64
+	mtx   sync.Mutex // Read/write lock.
+	db    dbm.DB     // Persistent node storage.
+	batch dbm.Batch  // Batched writing buffer.
+
+	versionCache  map[uint64]bool // Cache of tree (root) versions.
+	latestVersion uint64          // Latest root version.
+
+	nodeCache      map[string]*list.Element // Node cache.
+	nodeCacheSize  int                      // Node cache size limit in elements.
+	nodeCacheQueue *list.List               // LRU queue of cache elements. Used for deletion.
 }
 
 func newNodeDB(cacheSize int, db dbm.DB) *nodeDB {
 	ndb := &nodeDB{
-		cache:      make(map[string]*list.Element),
-		cacheSize:  cacheSize,
-		cacheQueue: list.New(),
-		db:         db,
-		batch:      db.NewBatch(),
-		versions:   map[uint64]bool{},
-		latest:     0,
+		nodeCache:      make(map[string]*list.Element),
+		nodeCacheSize:  cacheSize,
+		nodeCacheQueue: list.New(),
+		db:             db,
+		batch:          db.NewBatch(),
+		versionCache:   map[uint64]bool{},
 	}
 	return ndb
 }
@@ -58,21 +59,21 @@ func (ndb *nodeDB) GetNode(hash []byte) *IAVLNode {
 	defer ndb.mtx.Unlock()
 
 	// Check the cache.
-	if elem, ok := ndb.cache[string(hash)]; ok {
-		// Already exists. Move to back of cacheQueue.
-		ndb.cacheQueue.MoveToBack(elem)
+	if elem, ok := ndb.nodeCache[string(hash)]; ok {
+		// Already exists. Move to back of nodeCacheQueue.
+		ndb.nodeCacheQueue.MoveToBack(elem)
 		return elem.Value.(*IAVLNode)
 	}
 
 	// Doesn't exist, load.
 	buf := ndb.db.Get(hash)
 	if len(buf) == 0 {
-		cmn.PanicSanity(cmn.Fmt("Value missing for key %X", hash))
+		cmn.PanicSanity(cmn.Fmt("Value missing for key %x", hash))
 	}
 
 	node, err := MakeIAVLNode(buf)
 	if err != nil {
-		cmn.PanicCrisis(cmn.Fmt("Error reading IAVLNode. bytes: %X, error: %v", buf, err))
+		cmn.PanicCrisis(cmn.Fmt("Error reading IAVLNode. bytes: %x, error: %v", buf, err))
 	}
 
 	node.hash = hash
@@ -94,7 +95,7 @@ func (ndb *nodeDB) SaveNode(node *IAVLNode) {
 		cmn.PanicSanity("Shouldn't be calling save on an already persisted node.")
 	}
 
-	// Save node bytes to db
+	// Save node bytes to db.
 	buf := new(bytes.Buffer)
 	if _, err := node.writeBytes(buf); err != nil {
 		cmn.PanicCrisis(err)
@@ -118,7 +119,6 @@ func (ndb *nodeDB) SaveBranch(node *IAVLNode, cb func(*IAVLNode) *IAVLNode) {
 		return
 	}
 
-	// save children
 	if node.leftNode != nil {
 		ndb.SaveBranch(node.leftNode, cb)
 		node.leftNode = nil
@@ -135,6 +135,7 @@ func (ndb *nodeDB) SaveBranch(node *IAVLNode, cb func(*IAVLNode) *IAVLNode) {
 	}
 }
 
+// DeleteVersion deletes a tree version from disk.
 func (ndb *nodeDB) DeleteVersion(version uint64) {
 	ndb.mtx.Lock()
 	defer ndb.mtx.Unlock()
@@ -142,6 +143,15 @@ func (ndb *nodeDB) DeleteVersion(version uint64) {
 	ndb.deleteOrphans(version)
 	ndb.deleteRoot(version)
 	ndb.cleanup()
+}
+
+// Unorphan deletes the orphan entry from disk, but not the node it points to.
+func (ndb *nodeDB) Unorphan(hash []byte, version uint64) {
+	ndb.mtx.Lock()
+	defer ndb.mtx.Unlock()
+
+	key := fmt.Sprintf(orphansKeyFmt, version, hash)
+	ndb.batch.Delete([]byte(key))
 }
 
 // Saves orphaned nodes to disk under a special prefix.
@@ -154,15 +164,17 @@ func (ndb *nodeDB) SaveOrphans(orphans map[string]uint64) {
 	}
 }
 
+// Saves a single orphan to disk.
 func (ndb *nodeDB) saveOrphan(hash []byte, version uint64) {
 	key := fmt.Sprintf(orphansKeyFmt, version, hash)
 	ndb.batch.Set([]byte(key), hash)
 }
 
+// Return whether a version is the earliest available.
 func (ndb *nodeDB) isEarliestVersion(version uint64) bool {
-	earliest := ndb.latest
+	earliest := ndb.getLatestVersion()
 
-	for v, _ := range ndb.versions {
+	for v, _ := range ndb.getVersions() {
 		if v < earliest {
 			earliest = v
 		}
@@ -170,9 +182,12 @@ func (ndb *nodeDB) isEarliestVersion(version uint64) bool {
 	return version == earliest
 }
 
+// Cleanup is called after a version is deleted.
 func (ndb *nodeDB) cleanup() {
-	if len(ndb.versions) == 1 {
-		ndb.traverseOrphansVersion(ndb.latest, func(key, value []byte) {
+	// If only one version is left, we can cleanup all orphans, they aren't
+	// needed anymore.
+	if len(ndb.getVersions()) == 1 {
+		ndb.traverseOrphansVersion(ndb.getLatestVersion(), func(key, value []byte) {
 			ndb.batch.Delete(key)
 			ndb.batch.Delete(value)
 		})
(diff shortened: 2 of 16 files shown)


adrianbrink:
Could you create a develop branch and then PR this against develop.

It's just so that all our repos are in the same state.

cloudhead:
@ebuchman @ethanfrey @jaekwon this is ready to review.

ethanfrey:
Very nice description of the PR. And I assume you fixed that case the fuzz-tester found. Could you add the fuzz-tester command above to the PR.

Also, I remember you showing me some commands you ran to make sure that after saving and deleting multiple versions, the total size of the tree was the same as if there was no versioning, just one tree (eg. no lost orphans). Can you reference that somewhere?

Now going into the code.

cloudhead:
Added the fuzz command to the pr description. The test that checks size and checks that on every delete, the database is shrinking is: https://github.com/tendermint/iavl/blob/feature/historical-queries/iavl_tree_test.go#L764

Also: https://github.com/tendermint/iavl/blob/feature/historical-queries/iavl_tree_test.go#L328 and https://github.com/tendermint/iavl/blob/feature/historical-queries/iavl_tree_test.go#L129

iavl_nodedb.go
```diff
@@ -73,68 +107,416 @@ func (ndb *nodeDB) SaveNode(node *IAVLNode) {
...
-	delete(ndb.orphans, string(node.hash))
-	delete(ndb.orphansPrev, string(node.hash))
+// SaveBranch saves the given node and all of its descendants. For each node
+// about to be saved, the supplied callback is called and the returned node is
+// is saved. You may pass nil as the callback as a pass-through.
+//
+// Note that this function clears leftNode/rigthNode recursively and calls
+// hashWithCount on the given node.
+func (ndb *nodeDB) SaveBranch(node *IAVLNode, cb func(*IAVLNode)) {
+	if node.hash == nil {
```
ethanfrey: This is my issue with version.

version is set in https://github.com/tendermint/iavl/blob/feature/historical-queries/iavl_orphaning_tree.go#L75-L80 that is only in the callback

the first time a node is saved, we calculate the hashes recursively here. (with version=0)

iavl_nodedb.go
```diff
@@ -73,68 +107,416 @@ func (ndb *nodeDB) SaveNode(node *IAVLNode) {
...
+	if node.leftNode != nil {
+		ndb.SaveBranch(node.leftNode, cb)
+		node.leftNode = nil
+	}
+	if node.rightNode != nil {
+		ndb.SaveBranch(node.rightNode, cb)
+		node.rightNode = nil
+	}
+
+	if cb != nil {
```
ethanfrey: Then we set the version of the node here. Which would have modified the hash, except that it was already calculated above.

iavl_nodedb.go
```diff
@@ -73,68 +107,416 @@ func (ndb *nodeDB) SaveNode(node *IAVLNode) {
...
+// hashWithCount on the given node.
+func (ndb *nodeDB) SaveBranch(node *IAVLNode, cb func(*IAVLNode)) {
+	if node.hash == nil {
+		node.hash, _ = node.hashWithCount()
+	}
+	if node.persisted {
+		return
+	}
+
+	if node.leftNode != nil {
```
ethanfrey: We also have two recursions, one with hashWithCount and another with SaveBranch and I think they kind of do double work. I would look for a cleaner way to combine the two.

iavl_nodedb.go
ethanfrey: This is not a problem if they are in memory, but if we reload a tree, would they still match? Maybe we just take the version out of the hash and there is no more problem. It is still defined by the parent node, so I don't know that this adds security,.

iavl_tree.go
```diff
@@ -73,41 +71,52 @@ func (t *IAVLTree) Copy() *IAVLTree {
...
 	return t.root.has(t, key)
 }
 
+// Set a key.
 func (t *IAVLTree) Set(key []byte, value []byte) (updated bool) {
+	_, updated = t.set(key, value)
+	return updated
+}
+
+func (t *IAVLTree) set(key []byte, value []byte) (orphaned []*IAVLNode, updated bool) {
```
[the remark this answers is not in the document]
cloudhead: Yeah, was just wondering the same thing. Will do.

iavl_orphaning_tree.go
```diff
@@ -0,0 +1,105 @@
...
+// Remove a key from the underlying tree while storing the orphaned nodes.
+func (tree *orphaningTree) Remove(key []byte) ([]byte, bool) {
+	val, orphaned, removed := tree.IAVLTree.Remove(key)
+	tree.addOrphans(orphaned)
+	return val, removed
+}
+
+// Clone creates a clone of the tree.
+func (tree *orphaningTree) Clone() *orphaningTree {
+	inner := &IAVLTree{
```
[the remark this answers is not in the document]
cloudhead: Yeah the problem is that IAVLTree.Copy() panics when the tree is not persisted. But there are legitimate cases when I need to copy an unpersisted tree, for example if you `tree.SaveVersion(10)` then `tree.Remove(k)` then `tree.SaveVersion(11)` - the Remove just creates an orphan, but doesn't modify the tree.

iavl_nodedb.go
```diff
@@ -148,26 +148,26 @@ func (ndb *nodeDB) Has(hash []byte) bool {
...
 	if node.rightNode != nil {
 		ndb.SaveBranch(node.rightNode, cb)
-		node.rightNode = nil
 	}
 
 	if cb != nil {
 		cb(node)
 	}
+
+	node.hashWithCount()
```
ethanfrey: The callback is now executed *before* the hash, which is much better :)

iavl_orphaning_tree.go
```diff
@@ -73,10 +73,11 @@ func (tree *orphaningTree) SaveVersion(version uint64) {
 	// incorrectly marked as orphaned, since tree patterns after a re-balance
 	// may mirror previous tree patterns, with matching hashes.
 	tree.ndb.SaveBranch(tree.root, func(node *IAVLNode) {
-		tree.Unorphan(node.hash)
-
 		// The node version is set here since it isn't known until we save.
 		node.version = version
+		node.hashWithCount()
```
ethanfrey: but what?
we call hashWithCount in the callback, when it is called right after the callback in SaveBranch. twice?

iavl_proof.go
```diff
@@ -67,17 +67,18 @@ func (branch IAVLProofInnerNode) Hash(childHash []byte) []byte {
 type IAVLProofLeafNode struct {
 	KeyBytes   data.Bytes `json:"key"`
 	ValueBytes data.Bytes `json:"value"`
+	Version    uint64     `json:"version"`
```
ethanfrey: much cleaner

ethanfrey:
Definitely an improvement, but it seems we call hashWithCount now twice for every node in SaveBranch. hashWithCount then does it's own recursion...

To be fair, I think it aborts if the hash is calculated already, but two algorithms both decending down the tree recursively, calling each other... well, it seems ugly to me. Can we merge those two algorithms? as they have basically the same set of guards

iavl_orphaning_tree.go
cloudhead: They're both necessary - this one is needed because Unorphan needs the updated hash. The one in SaveBranch is necessary because it can't rely on the caller to hash. In any case work is only done once, as hashWithCount is a noop if the node is already hashed.

cloudhead:
Ok let me see if I can somehow merge them. The thing is as you say it only goes through the tree once, since it's hashing from the bottom (leafs) up, and hashWithCount doesn't do work if the nodes are already hashed.

iavl_tree.go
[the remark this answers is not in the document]
cloudhead: Done.

iavl_versioned_tree.go
```diff
@@ -0,0 +1,161 @@
...
+// DeleteVersion deletes a tree version from disk. The version can then no
+// longer be accessed.
+func (tree *VersionedTree) DeleteVersion(version uint64) error {
+	if version == 0 {
+		return errors.New("invalid version")
+	}
+	if version == tree.latest {
+		return errors.New("cannot delete latest saved version")
+	}
+	if _, ok := tree.versions[version]; ok {
```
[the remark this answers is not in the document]
cloudhead: You're right, done!

iavl_node.go
```diff
@@ -14,6 +15,7 @@ import (
 type IAVLNode struct {
 	key       []byte
 	value     []byte
+	version   uint64
```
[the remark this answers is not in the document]
cloudhead: Yeah I'm not super happy about it - it makes sense when you load a versioned tree, but when saving it does feel like bad coupling. hmmm.

iavl_nodedb.go
cloudhead: :+1:

iavl_proof.go
cloudhead: Agreed.

iavl_tree.go
```diff
@@ -116,19 +125,21 @@ func (t *IAVLTree) Hash() []byte {
...
+// HashWithCount returns the root hash and hash count.
 func (t *IAVLTree) HashWithCount() ([]byte, int) {
 	if t.root == nil {
 		return nil, 0
 	}
 	return t.root.hashWithCount()
 }
 
+// Save writes the tree to disk, if it was created with a datastore.
 func (t *IAVLTree) Save() []byte {
```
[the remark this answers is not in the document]
cloudhead: I'd rather not expose it actually. I think it's easier if we only have to expose VersionedTree. IAVLTree might be nice if it's actually immutable and non-persistent, but that requires a bunch of changes. Let me see what I can do.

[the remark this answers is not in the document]
cloudhead: Ok, removes Save() and Load(). This also makes the Tree() method on VersionedTree safer.

iavl_versioned_tree.go
```diff
@@ -0,0 +1,161 @@
+package iavl
+
+import (
+	"github.com/pkg/errors"
+
+	dbm "github.com/tendermint/tmlibs/db"
+)
+
+var ErrVersionDoesNotExist = errors.New("version does not exist")
```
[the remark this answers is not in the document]
cloudhead: Ah I forgot. How's this 751ac62?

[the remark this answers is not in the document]
cloudhead: Actually that still adds useless items to the trace, this is better: 868ca0f

iavl_node.go
```diff
@@ -209,9 +227,9 @@ func (node *IAVLNode) hashWithCount() ([]byte, int) {
 	return node.hash, hashCount + 1
 }
 
-// Writes the node's hash to the given io.Writer.
-// This function has the side-effect of calling hashWithCount.
-func (node *IAVLNode) writeHashBytes(w io.Writer) (n int, hashCount int, err error) {
+// Writes the node's hash to the given io.Writer. This function expects
```
ethanfrey: Very good to document when panics can ensue :+1:

iavl_node.go
```diff
@@ -222,29 +240,33 @@ func (node *IAVLNode) writeHashBytes(w io.Writer) (n int, hashCount int, err err
...
-			cmn.PanicSanity("node.rightHash was nil in writeHashBytes")
-		}
 		wire.WriteByteSlice(node.rightHash, w, &n, &err)
 	}
 	return
 }
 
+// Writes the node's hash to the given io.Writer.
+// This function has the side-effect of calling hashWithCount.
+func (node *IAVLNode) writeHashBytesRecursively(w io.Writer) (n int, hashCount int, err error) {
```
ethanfrey: ahh... for backwards compatibility

iavl_nodedb.go
```diff
@@ -147,27 +147,29 @@ func (ndb *nodeDB) Has(hash []byte) bool {
 //
 // Note that this function clears leftNode/rigthNode recursively and calls
 // hashWithCount on the given node.
-func (ndb *nodeDB) SaveBranch(node *IAVLNode, cb func(*IAVLNode)) {
+func (ndb *nodeDB) SaveBranch(node *IAVLNode, cb func(*IAVLNode)) []byte {
```
ethanfrey: Nice. I like this method.
And I think the return value is quite useful, as that hash is our persistent identifier

iavl_orphaning_tree.go
```diff
@@ -75,9 +79,7 @@ func (tree *orphaningTree) SaveVersion(version uint64) {
 	tree.ndb.SaveBranch(tree.root, func(node *IAVLNode) {
 		// The node version is set here since it isn't known until we save.
 		node.version = version
-		node.hashWithCount()
-
-		tree.Unorphan(node.hash)
+		tree.Unorphan(node._hash())
```
ethanfrey: okay, so _hash() is still called twice.
but it is not recursive and it caches the results, so that is fine with me

iavl_nodedb.go
ethanfrey: And one recursive tree traversal is better

iavl_tree.go
```diff
@@ -133,28 +133,6 @@ func (t *IAVLTree) HashWithCount() ([]byte, int) {
 	return t.root.hashWithCount()
 }
 
-// Save writes the tree to disk, if it was created with a datastore.
```
ethanfrey: :)

IAVLTree now just works in memory by itself, right?
Must use VersionedTree for persistence

iavl_tree.go
```diff
@@ -206,8 +184,15 @@ func (t *IAVLTree) GetLastInRangeWithProof(startKey, endKey []byte) ([]byte, []b
...
+// value, and 'true'.
+func (t *IAVLTree) Remove(key []byte) ([]byte, bool) {
+	value, _, removed := t.remove(key)
+	return value, removed
+}
+
+// remove tries to remove a key from the tree and if removed, returns its
 // value, nodes orphaned and 'true'.
-func (t *IAVLTree) Remove(key []byte) (value []byte, orphans []*IAVLNode, removed bool) {
+func (t *IAVLTree) remove(key []byte) (value []byte, orphans []*IAVLNode, removed bool) {
```
ethanfrey: :)

iavl_tree_test.go
```diff
@@ -60,25 +60,32 @@ func TestVersionedRandomTree(t *testing.T) {
...
 	}
+	singleVersionTree.SaveVersion(1)
 
 	for i := 1; i < versions; i++ {
 		tree.DeleteVersion(uint64(i))
 	}
 
 	// After cleaning up all previous versions, we should have as many nodes
-	// in the db as in the current tree version.
+	// in the db as in the current tree version. The simple tree must be equal
```
ethanfrey: i like this test a lot.
can you make a variant that deletes the versions in a semi-random order rather than just in-order? to shake out any edge cases.

`for _, i := range rand.Perm(1,versions) { .. }`
(https://golang.org/pkg/math/rand/#Perm) can give you a random ordering

iavl_versioned_tree.go
```diff
@@ -110,15 +110,16 @@ func (tree *VersionedTree) DeleteVersion(version uint64) error {
 	if version == tree.latest {
 		return errors.New("cannot delete latest saved version")
 	}
-	if _, ok := tree.versions[version]; ok {
-		tree.ndb.DeleteVersion(version)
-		tree.ndb.Commit()
+	if _, ok := tree.versions[version]; !ok {
+		return errors.WithStack(ErrVersionDoesNotExist)
```
ethanfrey: Thanks for the cleanup

iavl_nodedb.go
cloudhead: :tada:

iavl_orphaning_tree.go
cloudhead: Cool.

iavl_tree.go
cloudhead: Yeah exactly. I still don't think it's perfect because IAVLTree still has an ndb in its struct, but taking that out is somewhat complicating because of the ability for the tree to lazy-load.

ethanfrey:
I have yet to go through `iavl_nodedb.go` thoroughly. Otherwise, it looks good to me.

iavl_tree_test.go
cloudhead: Yes, good idea - also that is a cool function I didn't know about.

iavl_tree_test.go
```diff
@@ -95,6 +96,44 @@ func TestVersionedRandomTreeSmallKeys(t *testing.T) {
...
+			k := []byte(cmn.RandStr(1))
+			v := []byte(cmn.RandStr(8))
+			tree.Set(k, v)
+			singleVersionTree.Set(k, v)
+		}
+		tree.SaveVersion(uint64(i))
+	}
+	singleVersionTree.SaveVersion(1)
+
+	for _, i := range rand.Perm(versions) {
```
ethanfrey: This will cover [1, versions]
The non-random test deleted [1, versions)

I think you must do `rand.Perm(versions-1)`

cloudhead: Yeah I'm aware, but the delete of `version` will just silently fail so shouldn't matter? Or am I missing something?

cloudhead: Fixed anyway, it is more correct.

iavl_node.go
```diff
@@ -312,49 +346,58 @@ func (node *IAVLNode) set(t *IAVLTree, key []byte, value []byte) (newSelf *IAVLN
...
+
+		if len(orphaned) == 0 {
+			return node.hash, node, nil, value, []*IAVLNode{}
+		} else if newLeftHash == nil && newLeftNode == nil { // left node held value, was removed
+			return node.rightHash, node.rightNode, node.key, value, orphaned
+		}
+		orphaned = append(orphaned, node)
+
+		node = node.clone()
+		node.leftHash, node.leftNode = newLeftHash, newLeftNode
```
melekes: Do not see where you're setting the node.key to newKey for the left node. https://github.com/tendermint/iavl/pull/1/files#diff-a5ef36e3990bde5c8239a4c18ec3d25dR395

iavl_tree.go
```diff
@@ -251,3 +252,13 @@ func (t *IAVLTree) IterateRangeInclusive(start, end []byte, ascending bool, fn f
 		}
 	})
 }
+
+// nodeSize is like Size, but includes inner nodes too.
+func (t *IAVLTree) nodeSize() int {
+	size := 0
+	t.root.traverse(t, true, func(n *IAVLNode) bool {
+		size++
+		return false
```
melekes: hah? did not know one can return a bool as int

iavl_versioned_tree.go
```diff
@@ -0,0 +1,167 @@
...
+	dbm "github.com/tendermint/tmlibs/db"
+)
+
+var ErrVersionDoesNotExist = fmt.Errorf("version does not exist")
+
+// VersionedTree is a persistent tree which keeps track of versions.
+type VersionedTree struct {
+	*orphaningTree                           // The current, latest version of the tree.
+	versions       map[uint64]*orphaningTree // The previous, saved versions of the tree.
+	latest         uint64                    // The latest saved version.
```
melekes: maybe call it `latestSaved` than

cloudhead: I changed the comment on *orphaningTree to be less confusing and renamed latest to latestVerison. 411c361

Effectively, a version can only exist if it is saved.

iavl_node.go
cloudhead: Hmm, so this code was adapted from the old version of iavl, which has that peculiarity too, if you look at the deleted code. I'll take a deeper look, but didn't want to break anything by changing the logic. I'm not sure why it works this way, and whether or not it's correct. I've also tried removing that `if` statement, and the tests still pass :grimacing:

melekes: LOL, ok

iavl_versioned_tree.go
```diff
@@ -10,9 +10,9 @@ var ErrVersionDoesNotExist = fmt.Errorf("version does not exist")
 
 // VersionedTree is a persistent tree which keeps track of versions.
 type VersionedTree struct {
-	*orphaningTree                           // The current, latest version of the tree.
+	*orphaningTree                           // The current, working tree.
 	versions       map[uint64]*orphaningTree // The previous, saved versions of the tree.
-	latest         uint64                    // The latest saved version.
+	latestVersion  uint64                    // The latest saved version.
```
melekes: dope 👍

iavl_nodedb.go
```diff
@@ -73,68 +107,418 @@ func (ndb *nodeDB) SaveNode(node *IAVLNode) {
...
+
+	if version > ndb.getLatestVersion() {
+		ndb.latestVersion = version
+	}
+}
+
+func (ndb *nodeDB) getPreviousVersion(version uint64) uint64 {
+	var result uint64 = 0
+	for v, _ := range ndb.getVersions() {
+		if v < version && (result == 0 || v > result) {
```
ebuchman: is `result == 0` necessary ?

ebuchman: Also maybe there's a more efficient way to keep the previous version around so we dont have to iterate a map?

iavl_nodedb.go
```diff
@@ -73,68 +107,418 @@ func (ndb *nodeDB) SaveNode(node *IAVLNode) {
...
+
+// Traverse all keys with a certain prefix.
+func (ndb *nodeDB) traversePrefix(prefix []byte, fn func(k, v []byte)) {
+	if ldb, ok := ndb.db.(*dbm.GoLevelDB); ok {
+		it := ldb.DB().NewIterator(util.BytesPrefix([]byte(prefix)), nil)
+		for it.Next() {
+			k := make([]byte, len(it.Key()))
+			v := make([]byte, len(it.Value()))
+
+			// Leveldb reuses the memory, we are forced to copy.
```
ebuchman: Hmmm. Can we not have the individual `fn` that need to keep the values around do the copying? Seems there are cases where we don't need copies. 

Of course that wouldn't be the safest thing and would require clear comments everywhere, but just checking ...

iavl_nodedb.go
```diff
@@ -73,68 +107,418 @@ func (ndb *nodeDB) SaveNode(node *IAVLNode) {
...
+	// because we might be saving an old root at a new version in the case
+	// where the tree wasn't modified between versions.
+	key := fmt.Sprintf(rootsPrefixFmt, version)
+	ndb.batch.Set([]byte(key), root.hash)
+	ndb.cacheVersion(version, root.hash)
+
+	return nil
+}
+
+///////////////////////////////////////////////////////////////////////////////
```
ebuchman: Let's note that these functions are only used for testing. 

If any are used outside testing, we should move them, or make it clear.

iavl_orphaning_tree.go
```diff
@@ -0,0 +1,108 @@
...
+// Load the tree from disk, from the given root hash, including all orphans.
+func (tree *orphaningTree) Load(root []byte) {
+	if len(root) == 0 {
+		tree.root = nil
+	} else {
+		tree.root = tree.ndb.GetNode(root)
+	}
+
+	// Load orphans.
+	tree.ndb.traverseOrphansVersion(tree.root.version, func(k, v []byte) {
```
ebuchman: won't this panic if `tree.root == nil` ?

iavl_test.go
```diff
@@ -522,16 +518,18 @@ func testProof(t *testing.T, proof *KeyExistsProof, keyBytes, valueBytes, rootHa
 }
 
 func TestIAVLProof(t *testing.T) {
+	t.Skipf("This test has a race condition causing it to occasionally panic.")
```
ebuchman: ?!

iavl_tree.go
```diff
@@ -43,6 +38,9 @@ func (t *IAVLTree) String() string {
 	return "IAVLTree{" + strings.Join(leaves, ", ") + "}"
 }
 
+// DEPRECATED.
```
ebuchman: Can we note what it's been replaced with

iavl_versioned_tree.go
```diff
@@ -0,0 +1,167 @@
...
+		}
+	}
+	// Set the working tree to a copy of the latest.
+	tree.orphaningTree = tree.versions[tree.latestVersion].Clone()
+
+	return nil
+}
+
+// GetVersioned gets the value at the specified key and version.
+func (tree *VersionedTree) GetVersioned(key []byte, version uint64) (
```
ebuchman: I think we could just call this `Get`. Alternatively, we could use a `tree.Get(version).Get(key)` pattern. Thoughts ?

iavl_versioned_tree.go
```diff
@@ -0,0 +1,167 @@
...
+		return nil, errors.Errorf("version %d was already saved", version)
+	}
+	if tree.root == nil {
+		return nil, ErrNilRoot
+	}
+	if version == 0 {
+		return nil, errors.New("version must be greater than zero")
+	}
+	if version <= tree.latestVersion {
+		return nil, errors.New("version must be greater than latest")
```
ebuchman: lets give the values in this error

iavl_versioned_tree.go
```diff
@@ -0,0 +1,167 @@
...
+	tree.ndb.Commit()
+
+	return tree.root.hash, nil
+}
+
+// DeleteVersion deletes a tree version from disk. The version can then no
+// longer be accessed.
+func (tree *VersionedTree) DeleteVersion(version uint64) error {
+	if version == 0 {
+		return errors.New("invalid version")
```
ebuchman: "version must be greater than zero".

Below errors could include values too

iavl_node.go
```diff
@@ -57,6 +63,9 @@ func MakeIAVLNode(buf []byte) (node *IAVLNode, err error) {
 	}
 	buf = buf[n:]
 
+	node.version = wire.GetUint64(buf)
+	buf = buf[8:]
```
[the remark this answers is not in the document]
cloudhead: I wondered that myself - I just went with the style that was already there. Could also be due to the special case for leaf nodes vs inner nodes?

iavl_node.go
```diff
@@ -81,20 +90,41 @@ func MakeIAVLNode(buf []byte) (node *IAVLNode, err error) {
...
+
+// clone creates a shallow copy of a node with its hash set to nil.
+func (node *IAVLNode) clone() *IAVLNode {
 	if node.isLeaf() {
-		cmn.PanicSanity("Why are you copying a value node?")
+		cmn.PanicSanity("Attempt to copy a leaf node")
 	}
 	return &IAVLNode{
 		key:       node.key,
+		value:     node.value,
```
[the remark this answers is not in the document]
cloudhead: Woops. 2d295aff2eaf0ab64e28de0bb6b4f57902891651

iavl_node.go
```diff
@@ -276,34 +304,40 @@ func (node *IAVLNode) set(t *IAVLTree, key []byte, value []byte) (newSelf *IAVLN
...
+			orphaned = append(orphaned, rightOrphaned...)
 		}
+
 		if updated {
-			return node, updated
+			return node, updated, orphaned
 		} else {
 			node.calcHeightAndSize(t)
-			return node.balance(t), updated
+			new, balanceOrphaned := node.balance(t)
```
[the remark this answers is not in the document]
cloudhead: Yeah, good point. 6bd02c0

iavl_node.go
```diff
@@ -420,41 +461,48 @@ func (node *IAVLNode) calcBalance(t *IAVLTree) int {
...
 	if node.persisted {
 		panic("Unexpected balance() call on persisted node")
 	}
 	balance := node.calcBalance(t)
+
 	if balance > 1 {
 		if node.getLeftNode(t).calcBalance(t) >= 0 {
 			// Left Left Case
-			return node.rotateRight(t)
+			new, orphaned := node.rotateRight(t)
```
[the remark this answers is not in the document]
cloudhead: Yup. 6bd02c0

iavl_node.go
```diff
@@ -312,49 +346,58 @@ func (node *IAVLNode) set(t *IAVLTree, key []byte, value []byte) (newSelf *IAVLN
...
+) {
 	if node.isLeaf() {
 		if bytes.Equal(key, node.key) {
-			removeOrphan(t, node)
-			return nil, nil, nil, node.value, true
-		} else {
-			return node.hash, node, nil, nil, false
+			return nil, nil, nil, node.value, []*IAVLNode{node}
 		}
+		return node.hash, node, nil, nil, []*IAVLNode{}
```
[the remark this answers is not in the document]
cloudhead: Good call. 6dff6b2

iavl_nodedb.go
```diff
@@ -3,56 +3,89 @@ package iavl
...
-		db:          db,
-		batch:       db.NewBatch(),
-		orphans:     make(map[string]struct{}),
-		orphansPrev: make(map[string]struct{}),
+		nodeCache:      make(map[string]*list.Element),
+		nodeCacheSize:  cacheSize,
+		nodeCacheQueue: list.New(),
+		db:             db,
+		batch:          db.NewBatch(),
+		versionCache:   map[uint64][]byte{},
```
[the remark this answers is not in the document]
cloudhead: The only case I would use `make` is with channels or when using the extra len/cap parameters. In this case there's no reason, as the literal version is simpler. If it's a code-style thing though, I can change it, but I think this looks cleaner.

iavl_nodedb.go
```diff
@@ -3,56 +3,89 @@ package iavl
...
+		ndb.nodeCacheQueue.MoveToBack(elem)
 		return elem.Value.(*IAVLNode)
 	}
 
 	// Doesn't exist, load.
-	buf := ndb.db.Get(hash)
+	buf := ndb.db.Get(ndb.nodeKey(hash))
 	if len(buf) == 0 {
-		cmn.PanicSanity(cmn.Fmt("Value missing for key %X", hash))
+		cmn.PanicSanity(cmn.Fmt("Value missing for key %x", hash))
```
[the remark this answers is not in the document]
cloudhead: Ah, no I don't mind changing it, I was just making it consistent with the rest of the library.

iavl_nodedb.go
```diff
@@ -73,68 +107,418 @@ func (ndb *nodeDB) SaveNode(node *IAVLNode) {
...
+	key := ndb.nodeKey(hash)
+
+	if ldb, ok := ndb.db.(*dbm.GoLevelDB); ok {
+		exists, err := ldb.DB().Has(key, nil)
+		if err != nil {
+			cmn.PanicSanity("Got error from leveldb: " + err.Error())
+		}
+		return exists
+	}
+	return len(ndb.db.Get(key)) != 0
```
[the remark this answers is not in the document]
cloudhead: Looking at `tmlibs` - we return `nil` if `err == ErrNotFound`, so I guess we can change this to a nil check, and also the one in `GetNode`.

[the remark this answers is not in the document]
cloudhead: d83437b

iavl_nodedb.go
cloudhead: Yeah I think this behavior is so unexpected and low-level that I wouldn't want to expose it. I wasted a good 15 minutes trying to figure out what was wrong.

iavl_nodedb.go
```diff
@@ -3,56 +3,89 @@ package iavl
...
+
 	cmn "github.com/tendermint/tmlibs/common"
 	dbm "github.com/tendermint/tmlibs/db"
 )
 
+var (
+	// All node keys are prefixed with this. This ensures no collision is
+	// possible with the other keys, and makes them easier to traverse.
+	nodesPrefix = "n/"
+	nodesKeyFmt = "n/%x"
```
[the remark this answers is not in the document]
cloudhead: I think it's fine for now, but an obvious optimization when we need to. It makes things a lot easier at the moment for the keys to be easily printable and readable, but eventually we might want to use binary prefixes and hashes for sure.

iavl_node.go
```diff
@@ -374,11 +417,10 @@ func (node *IAVLNode) getRightNode(t *IAVLTree) *IAVLNode {
 
 // NOTE: overwrites node
 // TODO: optimize balance & rotate
-func (node *IAVLNode) rotateRight(t *IAVLTree) *IAVLNode {
-	node = node._copy()
+func (node *IAVLNode) rotateRight(t *IAVLTree) (*IAVLNode, *IAVLNode) {
```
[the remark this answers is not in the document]
cloudhead: Done. 8a1918c

iavl_nodedb.go
```diff
@@ -73,68 +107,418 @@ func (ndb *nodeDB) SaveNode(node *IAVLNode) {
...
+		ndb.saveOrphan([]byte(hash), fromVersion, toVersion)
+	}
+}
+
+// Saves a single orphan to disk.
+func (ndb *nodeDB) saveOrphan(hash []byte, fromVersion, toVersion uint64) {
+	if fromVersion > toVersion {
+		cmn.PanicSanity("Orphan expires before it comes alive")
+	}
+	key := fmt.Sprintf(orphansKeyFmt, toVersion, fromVersion, hash)
```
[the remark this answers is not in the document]
cloudhead: 04e2482

iavl_test.go
cloudhead: Yeah, this panics one out of 20 times on average, I guess it wasn't detected when it was written. I could just delete the test since I've written many new tests for proofs. It's actually not a race condition now that I look at it, but probably just due to some non-deterministic bug in the test code.

iavl_versioned_tree.go
cloudhead: That's one option for sure, then we'd have to rename `Get` to `GetLatest`. The reason I have it this way is that `Get` has the same type for both versioned and non-versioned trees. I'm not a fan of the double get, I'd rather expose only one interface (VersionedTree).

iavl_nodedb.go
cloudhead: Done. 60d9992

iavl_orphaning_tree.go
cloudhead: Good point. 1126fcc

iavl_tree.go
cloudhead: Yup. 27b5237

iavl_versioned_tree.go
cloudhead: 0215259

iavl_versioned_tree.go
cloudhead: 0215259

cloudhead:
Updated README.

cloudhead:
Merging this now, feel free to open issues or comment post-merge.

PR #2, opened 2017-09-06
Serialize proofs

ethanfrey:
We need to have solid binary reader and writers for the proofs.  At least the key proofs for now, until we expose more to the outside worlds.


ethanfrey:
Tests fail, as a I can deserialize a KeyAbsentProof into a KeyExistsProof

PR #5, opened 2017-09-27
linting: add gometalinter to Makefile

zramsay:
- introduces `make metalinter` and `make metalinter_test`. The latter should be phased out eventually when the code complies with `--enable-all`
- only fixed the typos because I'm not sure what the status is on the code. (e.g., deadcode - will it be used?)


zramsay:
will re-implement with v2 when best practices are worked out

PR #9, opened 2017-10-18
Use linked-list to store versions

cloudhead:
Proposal by @jaekwon to store versions in a linked-list. The main benefit here is we don't need to store a version cache in memory, and it performs roughly the same.

### DIFF
--- a/nodedb.go
+++ b/nodedb.go
@@ -8,6 +8,7 @@ import (
 	"sort"
 	"sync"
 
+	wire "github.com/tendermint/go-wire"
 	cmn "github.com/tendermint/tmlibs/common"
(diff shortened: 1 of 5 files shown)


marbar3778:
This PR has been open for about two years, does not seem like there is interest to include it, Will keep the branch, but will close this PR.

cloudhead:
Yeah I think the benefits are kind of a matter of preference :man_shrugging:

marbar3778:
@cloudhead if you are interested in helping https://github.com/tendermint/tendermint/pull/3769 was just opened about state-sync with a link to your proposal, your comments would be valuable to the design process

PR #15, opened 2017-11-29
Implement LoadVersion and ResetToLatest

cloudhead:


ethanfrey:
Looks good.
But what happens when LoadVersion(v) gets a version that is not stored in the db?
It should detect this and return an error

cloudhead:
@ethanfrey https://github.com/tendermint/iavl/pull/15/files#diff-06e77bf1f34b6f5d0fe30e83bf1f7861R88 isn't enough?

ethanfrey:
Yes, you are right, didn't see that... prefer to review code in my editor

ethanfrey:
then imo you can merge to sdk2

PR #25, opened 2017-12-27
WIP: Improve benchmarks

ethanfrey:
I'm going through the benchmarks to make them a bit more realistic along with the versionedtree. In doing so, I found some issues, which I will link to this pr.

This branch is broken, and until it is fixed stands mainly to demonstrate the issues.


cloudhead:
I would base this off of sdk2, or it'll break when we merge. The main thing is `SaveVersion` doesn't take arguments anymore.

PR #49, opened 2018-05-03
Minimal changes to resolve #38

### LINKED ISSUE #38: Switch the internal hash function to SHA2 from RIPEMD160
I'm expecting that the auditors will eventually make the same recommendation.

But I don't think we should go to 1.0 with RIPEMD160 as our hash function.

RIPEMD160 is a bad choice for a few reasons.

1. The birthday bound on RIPEMD160 is 2^80 for a second preimage. A 2nd preimage would at very least allow for state corruption in our database and possibly allow for forgeries

2. RIPEMD160 is roughly the same speed/throughput as SHA256 on current hardware. Future generations of Intel and ARM server hardware have will have native SHA256 instructions that allow massive speed ups. Here is how fast SHA156 is on Ryzen: https://crypto.stackexchange.com/questions/50618/how-fast-can-a-sha-256-implementation-go/50620

3. RIPEMD160 is 10x more expensive to verify on the EVM than sha256.
https://github.com/ethereum/go-ethereum/blob/master/params/protocol_params.go#L69-L72

**Bikeshed**

It's a reasonable choice to ask if we should adopt a modern hash function. The consensus among hash function experts is that SHA2 will never be broken and will likely be the most widely supported hash function for the rest of human existence. 

BLAKE2b is almost 5x faster in software than RIPE160MD but doesn't currently have ethereum ecosystem compatibility and is unlikely ever to be implemented in widely available hardware. Tho hardware sha256 is only 2x faster blake2b.

Shake128 is kinda slow. Farfalle is very fast but too new.

Ethereum KECCAK is indefensible.

Liamsi:
Change from RIPEMD160 to SHA256
resolves #38 

If #53 is merged we can additionally remove ripemd160 from the glide file (or dep file if #52 get's merged first)

### DIFF
--- a/sha256truncated/sha256truncated.go
+++ b/sha256truncated/sha256truncated.go
@@ -1,0 +1,44 @@
+// Package sha256truncated provides a sha256 hash.Hash whose output is truncated to 20 bytes (160 bits).
+//
+// This is the default hashing algorithm used by IAVL+ trees.
+//
+//   s256 := sha256.New() // crypto/sha256
+//   s256Truncated := New() // this package
+//
+//   // Use like any other hash.Hash ...
+//   // Contract:
+//   s256Trunc.Sum(nil) == s256.Sum(nil)[:20]
+package sha256truncated
+
+import (
+	"crypto/sha256"
+	"hash"
+)
+
+const Size = 20
+
+// New returns a new hash.Hash computing the truncated to the first 20 bytes SHA256 checksum.
+func New() hash.Hash {
+	return &digest{sha256.New()}
+}
+
+func (d *digest) Sum(in []byte) []byte {
+	return d.Hash.Sum(in)[:20]
+}
+
+func (d *digest) Reset() {
+	d.Hash.Reset()
+}
+
+func (d *digest) Size() int {
+	return Size
+}
+
+func (d *digest) BlockSize() int {
+	return d.Hash.BlockSize()
+}
+
+// digest is just a wrapper around sha256
+type digest struct {
+	hash.Hash
+}
--- a/sha256truncated/sha256truncated.go
+++ b/sha256truncated/sha256truncated.go
@@ -23,7 +23,7 @@ func New() hash.Hash {
 }
 
 func (d *digest) Sum(in []byte) []byte {
-	return d.Hash.Sum(in)[:20]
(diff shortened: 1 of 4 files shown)


ebuchman:
Maybe it would be better if we have a `sha256truncated` utility function that we can just call with the `buf`, instead of having to create and write to the hasher and then slice it every time. At least that way if we change it again we just change it in one place

Liamsi:
Good point. It's a bit messy with all the `[:20]` right. My intention was to submit this PR as is (with the least changes possible) and further improve this in a followup PR shortly after (see https://github.com/tendermint/iavl/issues/38#issuecomment-386290979 and https://github.com/tendermint/iavl/issues/11). Will do as you suggested and submit another PR later.

Liamsi:
> Maybe it would be better if we have a sha256truncated utility function that we can just call with the buf, instead of having to create and write to the hasher and then slice it every time.

I took a slightly different approach and added a tiny package for the hasher. It makes the hasher a bit more visible in generated godoc and adds a bit more flexibility if we switched to another hasher. 

PTAL @ebuchman.

sha256truncated/sha256truncated.go
```diff
@@ -0,0 +1,44 @@
...
+
+const Size = 20
+
+// New returns a new hash.Hash computing the truncated to the first 20 bytes SHA256 checksum.
+func New() hash.Hash {
+	return &digest{sha256.New()}
+}
+
+func (d *digest) Sum(in []byte) []byte {
+	return d.Hash.Sum(in)[:20]
```
melekes: Size

sha256truncated/sha256truncated_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+	s256.Write(testStr2)
+	s256Trunc.Write(testStr2)
+
+	require.Equal(t, s256Trunc.Sum(nil), s256.Sum(nil)[:20])
+}
+
+func TestSize(t *testing.T) {
+	got := New().Size()
+	want := 20
```
melekes: Size

sha256truncated/sha256truncated.go
Liamsi: Done. Thanks!

Issue #51, opened 2018-05-03
Remove dependency on go-wire


Liamsi:
iavl uses/vendors go-wire (which depends on ripemd160)
Make it either indpendendent from go-wire / amino, or update to amino.

Remove ripemd dependency from gilde file(s).

zmanian:
We need to make some decision here.

We clearly need to remove the go-wire dependency.

Based on how wire is invoked right now, it is only used to encode fixed length structures.

If we switch to amino, compatible implementations will need to have some awareness of amino. To me this looks like the `typ3` table and `varint` encoding maybe but not the`disfix` identifier.

Are we comfortable requiring this?

One use case for compatible implementations will implementing IBC in lotion and in our partners stacks.

zmanian:
I think this is semi relevant to 
https://github.com/cosmos/cosmos-sdk/pull/454/files

Liamsi:
Thanks for your input @zmanian. There is a related PR #53. I'm not sure yet if I found a bug [here](https://github.com/tendermint/iavl/pull/53#issuecomment-386603451) or if the way iavl uses amino needs to be updated.

ebuchman:
Looks like IAVL never actually encodes a struct, so I dont think it needs to know the typ3 table - it only calls `EncodeIntXxx` and  `EncodeByteSlice`, which AFAIK are just purely proto3 encodings of ints and byte slices. 

That sounds pretty ideal for this.

zmanian:
`EncodeIntXxx` is just big endian fixed size int encoding

`EncodeByteSlice` is varint LEB128 encoded length prefixed bytes

ebuchman:
Int32 and Int64 are the fixed-size big-endian, but Int8 is apparently encoded as a Varint ...

In any case, there's nothing amino specific about these, right ?

zmanian:
Yeah 8 and 18 are varints

zmanian:
Yeah these aren't amino things. But LEB128 varints and bigendian fixed with ints are specifications the protobufs and amino incorporate

Liamsi:
Closing this. The go-wire (golang package) dependency was removed with #53. I'll update the documentation to clarify encoding specific things. 

linking to #56

PR #53, opened 2018-05-04
Switch from go-wire to go-amino

Liamsi:
Use latest amino instead of go-wire (see #51)

WIP / Notes: After switching to amino e.g. `TestVersionedRandomTreeSmallKeys`
currently fails/panics (at least most of the times) with: 

panic: Error reading Node. bytes: [...] error: insufficient bytes decoding []byte of length [...]

`TestTreeKeyRangeProof` fails with "invalid proof"

### DIFF
--- a/amino.go
+++ b/amino.go
@@ -1,0 +1,5 @@
+package iavl
+
+import "github.com/tendermint/go-amino"
+
+var cdc = amino.NewCodec()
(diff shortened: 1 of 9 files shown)


Liamsi:
Looks like https://github.com/tendermint/go-amino/commit/f144b939e22210d4219f2d84039eab67db0663de
breaks decoding (or encoding) of nodes (prev. commit 5d7845f24b843c914cf571dad2ca13c91cf70f0d still works)

ebuchman:
Oh no, this was already done in https://github.com/tendermint/iavl/pull/45/files ! I'll try to reconcile ...

PR #54, opened 2018-05-04
Switch to dep from glide

### LINKED ISSUE #52: switch from glide to dep
most of the stack has made this switch

zmanian:
Closes #52


Gopkg.lock
```diff
@@ -0,0 +1,125 @@
...
+    "leveldb/memdb",
+    "leveldb/opt",
+    "leveldb/storage",
+    "leveldb/table",
+    "leveldb/util"
+  ]
+  revision = "34011bf325bce385408353a30b101fe5e923eb6e"
+
+[[projects]]
+  branch = "master"
```
Liamsi: Nitpick: for other tendermint projects we use the `develop` branch. I wonder if we should stay consistent here.

Gopkg.toml
```diff
@@ -0,0 +1,42 @@
...
+#   source = "github.com/myfork/project2"
+#
+# [[override]]
+#   name = "github.com/x/y"
+#   version = "2.4.0"
+#
+# [prune]
+#   non-go = false
+#   go-tests = true
+#   unused-packages = true
```
Liamsi: Nitpick: maybe delete this example doc?

Gopkg.toml
```diff
@@ -0,0 +1,42 @@
...
+
+
+[[constraint]]
+  name = "github.com/stretchr/testify"
+  version = "1.1.4"
+
+[[constraint]]
+  branch = "master"
```
Liamsi: like above: s/master/develop ?

zmanian: Fix this after we figure how how we want integrate amino.

ebuchman:
rebased and merged to develop

Gopkg.toml
ebuchman: actually best to use versions until/unless we need something from a branch, and then in that situation ideally use a tag

PR #58, opened 2018-05-07
WIP: some lints

Liamsi:
Some nitpicks I've stumbled upon while reading through the code (and while using golint / gometalinter).

- consistently named receiver
- remove obsolete if blocks <sup>1)</sup>

Low priority PR ...

---
<sup>1)</sup> "In the Go libraries, you'll find that when an if statement doesn't flow into the next statement—that is, the body ends in break, continue, goto, or return—the unnecessary else is omitted." [source](https://golang.org/doc/effective_go.html#if)


Liamsi:
Superseded by #80

PR #61, opened 2018-06-06
Add GetVersion method to VersionedTree

aastein:
Added a GetVersion method to get a single version of the tree. The existing methods for accessing a version did not provide enough flexibility for my use case.


Liamsi:
Hey, any reason you just closed the issue? Did you find an alternative way to use the API to do what you tried to acchieve?

aastein:
Hi, I closed the PR because I ended up implementing a solution that did not depend on this change.

PR #72, opened 2018-06-29
Fix linting error on the sdk

ValarDragon:
For an unknown reason, this line is causing a linting error in the sdk. (Even though we have it set to ignore the vendor directories, and it succesfully ignores all other linting errors in other vendor directories)


Liamsi:
Sounds like a linter bug (surprising as this is in the vendored dir). Linking to #58 as it would be nice if `nolint` wasn't necessary here anyways.

PR #80, opened 2018-07-03
Various lints

Liamsi:
As it is planned to merge iavl into cosmos-sdk, we need to conform to the lints used there:
https://github.com/cosmos/cosmos-sdk/blob/3f438cdbc2ed6b462e3e146403f836c8bb680317/Makefile#L104-L107

This PR contains most changes necessary for this.

It does not contain changes which would silence warnings a la:
`comment on exported method Node.PathToLeaf should be of the form "PathToLeaf ..." (golint)` 
I think in tendermint we do not lint for these, too.

It also merges master into develop (maybe should be a separate change though).

Supersedes #58


proof.go
```diff
@@ -144,10 +144,14 @@ func (pln proofLeafNode) Hash() []byte {
 // a path to the least item.
 func (node *Node) PathToLeaf(t *Tree, key []byte) (PathToLeaf, *Node, error) {
 	path := new(PathToLeaf)
-	val, err := node._PathToLeaf(t, key, path)
+	val, err := node._pathToLeaf(t, key, path)
```
melekes: why underscore in the beginning?

node.go
```diff
@@ -531,34 +525,32 @@ func (node *Node) balance(t *Tree) (newSelf *Node, orphaned []*Node) {
...
 	if balance < -1 {
 		if node.getRightNode(t).calcBalance(t) <= 0 {
 			// Right Right Case
 			newNode, orphaned := node.rotateLeft(t)
 			return newNode, []*Node{orphaned}
-		} else {
-			// Right Left Case
-			var rightOrphaned *Node
+		}
+		// Right Left Case
```
melekes: I'd advocate that sometimes `if then else` does make perfect sense and this is one of those cases

```
if left {
} else {
  (right)
}
```

NOT

```
if left {
}

(right)
```

proof_range.go
```diff
@@ -190,9 +188,8 @@ func (proof *RangeProof) verify(root []byte) (err error) {
 	}
 	if !bytes.Equal(rootHash, root) {
 		return cmn.ErrorWrap(ErrInvalidRoot, "root hash doesn't match")
-	} else {
-		proof.rootVerified = true
 	}
+	proof.rootVerified = true
```
melekes: note: this is the good case where `else` is not needed indeed

proof_range.go
```diff
@@ -179,7 +177,7 @@ func (proof *RangeProof) Verify(root []byte) error {
 	return err
 }
 
-func (proof *RangeProof) verify(root []byte) (err error) {
+func (proof *RangeProof) verify(root []byte) (error) {
```
melekes: brackets around `error` can be removed

proof.go
Liamsi: Was there before. I think Jae uses them to mark recursive helpers. I can remove it here but there are many other places.

node.go
Liamsi: I kinda agree. It makes more sense here then in many other places. I think the point is that there is a return in the first case (which makes the else obsolete semantically).  Can reintroduce.

PR #86, opened 2018-07-12
Add substore proof

HaoyangLiu:


Liamsi:
The changes look non-critical / good to me. Can you explain in the PR description why this is necessary and how you intend to use the sub-store commit-id?

proof_range.go
```diff
@@ -10,7 +10,15 @@ import (
...
 
+type SubstoreCommitID struct {
+	Name string `json:"name"`
+	Version int64 `json:"version"`
+	CommitHash    cmn.HexBytes `json:"commit_hash"`
+}
+
 type RangeProof struct {
+	MultiStoreCommitInfo []SubstoreCommitID `json:"multi_store_commit_info"`
+	StoreName string `json:"store_name"`
```
Liamsi: Will this be used written from outside? Both new fields are never used here.

HaoyangLiu:
@Liamsi 
The referenced code:
https://github.com/irisnet/cosmos-sdk/blob/d68616d92583940e11d73ff129dd73a86d2ee1b0/store/rootmultistore.go#L317
and 
https://github.com/irisnet/cosmos-sdk/blob/d68616d92583940e11d73ff129dd73a86d2ee1b0/client/context/helpers.go#L132

Liamsi:
Thanks for the explanation @HaoyangLiu! That's very helpful. The only thing I'm wondering about: if this is only read / written from outside the iavl code, couldn't a wrapper around rangeproof do the trick for you?

HaoyangLiu:
Hi @Liamsi,

Thanks for your feedback. If I wrapper it, I will have to change ResponseQuery in ABCI query which will affect more code. So how about the latest version? I have commit a new change. In this change, I only add a new member: Appendix. Thus arbitrary additional data can be appended. Do you think this is acceptable?

Liamsi:
Yes, that is better. On one hand, it might be interesting for other use cases to add associated data in that new field. On the other hand, I still do not understand why the data needs to be added to the range proof (if it is not really used there). While in you use-case it seems to make sense (and seems to save you a lot of work writing a wrapper) it might be confusing in general. 
cc @jaekwon  @jlandrews

PR #95, opened 2018-08-06
Fix `go vet` failures

misha-tectonic:


misha-tectonic:
Go 1.11beta2, beta3, rc1.

`go vet` test modules recursively, and `go test` runs `go vet` in 1.11, so any module that explicitly or implicitly includes `iavl` fails in the testsuite.

Liamsi:
Thanks!

Liamsi:
Interestingly, only https://github.com/tendermint/iavl/pull/95/commits/c917ed2043a0e0cc01425c73d16e83571d5c76b5 seems necessary to make `go vet` happy:
```
go1.11rc1 test
# github.com/tendermint/iavl
./proof_range.go:368:6: lastDepth declared but not used
vet: typecheck failures
FAIL    github.com/tendermint/iavl [build failed]
```

In any case the unreachable code doesn't make sense. If there is an overflow `append` will panic and the second return is still unreachable.

misha-tectonic:
`go test` runs a subset of `go vet` tests. Running `go vet` standalone will emit the error about unreachable statement.

Liamsi:
I also ran `go1.11rc1 vet` and it only complained about
```
go1.11rc1 vet
# github.com/tendermint/iavl
./proof_range.go:368:6: lastDepth declared but not used
vet: typecheck failures
```

I was surprised about this too. I do remember vet complaining about the unreachable statement (a while ago though).

misha-tectonic:
Oh, I see. `go vet` stops early: once you remove unused variables it starts complaining about unreachable code:

```
% git describe --tags
v0.9.2-1-gc917ed2
% go vet
# github.com/tendermint/iavl
./util.go:75: unreachable code
```

PR #96, opened 2018-08-08
Remove architecture dependent getter functions

### LINKED ISSUE #93: Do we need separate getters for cast return types?
We currently have multiple similar getter methods, such as Get64, Has64, Version64, etc. with corresponding methods Get, Has, Version, etc. which return regular ints.

Are these needed for some specific reason? or is it just to avoid user side casting to int?

cc @Liamsi

jlandrews:
Closes #93

I've checked using this branch as a dependency for the SDK and it passes tests with just minor changes. (SDK currently calls `Version64` which in this branch can be accessed as `Version`)

cc @Liamsi


Liamsi:
I've merged #92 (thanks again @jlandrews). Please rebase on (or merge in develop) and will merge this too.

Liamsi:
I've resolved the merge conflicts and would merge into develop.

Liamsi:
@jlandrews can you add an entry to the Changelog?

PR #106, opened 2018-08-29
Merge back master to develop

Liamsi:


Liamsi:
I do not fully understand why the diff shows that many changes although the files are identical on master/develop. I'll close this and create a branch with a merge commit instead.

PR #107, opened 2018-08-30
Use 8 bytes to store int64 components of database keys 

### LINKED ISSUE #104: Use 64-bit wide prefix rather than 10 decimal digits
It would take a few hundred years to cause my use case (one version per Tendermint block) any problems, but it currently we overflow various of our nodedb prefixes well before we overflow the underyling `int64` type used for versions:

https://github.com/tendermint/iavl/blob/master/nodedb.go#L24-L30 

IAVL used to load every node from previous versions:

https://github.com/tendermint/iavl/compare/v0.9.2...v0.10.0#diff-6d87913bee304887177dad726b816faeR256

Once we get past version 9999999999 then we will not be able to load newer versions because of the behaviour of `Sscanf`: https://github.com/tendermint/iavl/blob/master/nodedb.go#L344.

We'll actually still save them because the corresponding `Sprintf` will include digits outside the padding. But we'll hit this error when loading a large version: https://github.com/tendermint/iavl/blob/master/mutable_tree.go#L248 

It would make sense to use a prefix that is 8 bytes - we could use raw big-endian padded bytes or we could use 16 chars of hex so that we maintain the lexicographical sorting we need for iteration. At least things blow up at the same time (and a significantly larger number) as well as it being neater to concede the same version bound.

Note the lazy-loading added by 0.10.0 makes it cheap to store a large number of versions, which is nice.

silasdavis:
This fixes #104 by introducing `KeyFormat` that uses a fixed-width key format over the full 8 bytes provided by int64 versions in orphan, and root keys. It also relies on the fixed width format to provide more efficient code for formatting into and scanning values from keys.

As a result of this it seems to lead to a ~10% speedup in `BenchmarkTreeLoadAndDelete` by avoiding the relatively expensive `fmt.Sprintf` and `fmt.Sscanf` functions, see below.

#### Before `KeyFormat` introduction
```shell
go test -bench=.
goos: linux
goarch: amd64
pkg: github.com/tendermint/iavl
BenchmarkNodeKey-8                 	 5000000	       316 ns/op
BenchmarkOrphanKey-8               	 2000000	       613 ns/op
ok, starting
BenchmarkImmutableAvlTreeMemDB-8   	200	   5783362 ns/op	 1104190 B/op	   22488 allocs/op
BenchmarkTreeLoadAndDelete/LoadAndDelete-8         	       2	 707800784 ns/op
PASS
ok  	github.com/tendermint/iavl	129.478s
```

#### After `KeyFormat` introduction
```shell
go test -bench=.
goos: linux
goarch: amd64
pkg: github.com/tendermint/iavl
BenchmarkNodeKey-8                 	30000000	        50.9 ns/op
BenchmarkOrphanKey-8               	 5000000	       256 ns/op
BenchmarkImmutableAvlTreeMemDB-8   	     500	   4048465 ns/op	   82591 B/op	    1546 allocs/op
BenchmarkTreeLoadAndDelete/LoadAndDelete-8         	       5	 641415439 ns/op
PASS
ok  	github.com/tendermint/iavl	121.982s
```

For `BenchmarkTreeLoadAndDelete/LoadAndDelete-8` we have 1 - 641415439/707800784 ~= 9.3%

### DIFF
--- a/key_format.go
+++ b/key_format.go
@@ -13,9 +13,22 @@ type KeyFormat struct {
 }
 
 // Create a []byte key format based on a single byte prefix and fixed width key segments each of whose length is
-// specified by by the corresponding element of layout
+// specified by by the corresponding element of layout.
+//
+// For example, to store keys that could index some objects by a version number and their SHA256 hash using the form:
+// 'c<version uint64><hash [32]byte>' then you would define the KeyFormat with:
+//
+//  var keyFormat = NewKeyFormat('c', 8, 32)
+//
+// Then you can create a key with:
+//
+//  func ObjectKey(version uint64, objectBytes []byte) []byte {
+//  	hasher := sha256.New()
+//  	hasher.Sum(nil)
+//  	return keyFormat.Key(version, hasher.Sum(nil))
+//  }
 func NewKeyFormat(prefix byte, layout ...int) *KeyFormat {
-	// For prefix
+	// For prefix byte
 	length := 1
 	for _, l := range layout {
 		length += int(l)
@@ -27,7 +40,7 @@ func NewKeyFormat(prefix byte, layout ...int) *KeyFormat {
 	}
 }
 
-// Format the byte segments into the key format - will panic if the segment lengths to do match the layout.
+// Format the byte segments into the key format - will panic if the segment lengths do not match the layout.
 func (kf *KeyFormat) KeyBytes(segments ...[]byte) []byte {
 	key := make([]byte, kf.length)
 	key[0] = kf.prefix
@@ -45,9 +58,9 @@ func (kf *KeyFormat) KeyBytes(segments ...[]byte) []byte {
 	return key[:n]
 }
 
-// Format the args passed into the key format - will panic if the arguments passed to not match the length
+// Format the args passed into the key format - will panic if the arguments passed do not match the length
 // of the segment to which they correspond. When called with no arguments returns the raw prefix (useful as a start
-// element of the entire keys space when sorted lexicographically)
+// element of the entire keys space when sorted lexicographically).
 func (kf *KeyFormat) Key(args ...interface{}) []byte {
 	if len(args) > len(kf.layout) {
 		panic(fmt.Errorf("KeyFormat.Key() is provided with %d args but format only has %d segments",
@@ -60,7 +73,7 @@ func (kf *KeyFormat) Key(args ...interface{}) []byte {
 	return kf.KeyBytes(segments...)
 }
 
-// Reads out the bytes associated with each segment of the key format from key
+// Reads out the bytes associated with each segment of the key format from key.
 func (kf *KeyFormat) ScanBytes(key []byte) [][]byte {
 	segments := make([][]byte, len(kf.layout))
 	n := 1
@@ -87,7 +100,7 @@ func (kf *KeyFormat) Scan(key []byte, args ...interface{}) {
 	}
 }
 
-// Return the prefix as a string
+// Return the prefix as a string.
 func (kf *KeyFormat) Prefix() string {
 	return string([]byte{kf.prefix})
 }
@@ -95,6 +108,7 @@ func (kf *KeyFormat) Prefix() string {
 func scan(a interface{}, value []byte) {
 	switch v := a.(type) {
 	case *int64:
+		// Negative values will be mapped correctly when read in as uint64 and then type converted
 		*v = int64(binary.BigEndian.Uint64(value))
 	case *uint64:
 		*v = binary.BigEndian.Uint64(value)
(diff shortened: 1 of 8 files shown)


ValarDragon:
This is an incredible performance win!
Does the key being negative matter?

proof_test.go
```diff
@@ -17,7 +17,7 @@ func TestTreeGetWithProof(t *testing.T) {
 	require := require.New(t)
 	for _, ikey := range []byte{0x11, 0x32, 0x50, 0x72, 0x99} {
 		key := []byte{ikey}
-		tree.Set(key, []byte(rand.Str(8)))
+		tree.Set(key, []byte(random.Str(8)))
```
silasdavis: I made this rename because `rand` clashes with standard `math/rand` which I wanted to use. I could of aliased by I thought it was better to avoid clashing with stdlib (maybe better still to not use a global...? (even better to perhaps do without tendermint/common :o - but I should tread lightly for now :) )).

testutils_test.go
```diff
@@ -115,23 +115,22 @@ func benchmarkImmutableAvlTreeWithDB(b *testing.B, db db.DB) {
 	b.StopTimer()
 
 	t := NewMutableTree(db, 100000)
+	value := []byte{}
 	for i := 0; i < 1000000; i++ {
-		t.Set(i2b(int(cmn.RandInt32())), nil)
+		t.Set(i2b(int(cmn.RandInt32())), value)
```
silasdavis: `go test -bench=.` was panicking on develop without this change

silasdavis:
@ValarDragon negative gives should be fine - I've added a test.

ValarDragon:
Just want to note, this performance increase is actually super helpful for our state machine as well. Some benchmarks I was doing had the sprintf calls from GetNode and SaveAs as taking up 1.5% of the entire state machines time. (Includes goleveldb compaction)

key_format.go
```diff
@@ -0,0 +1,127 @@
...
+
+// Provides a fixed-width lexicographically sortable []byte key format
+type KeyFormat struct {
+	prefix byte
+	layout []int
+	length int
+}
+
+// Create a []byte key format based on a single byte prefix and fixed width key segments each of whose length is
+// specified by by the corresponding element of layout
```
Liamsi: Maybe, a little example would be great here?

testutils_test.go
Liamsi: linking https://github.com/tendermint/iavl/issues/76

key_format.go
```diff
@@ -0,0 +1,127 @@
...
+		n += l
+		// Big endian so pad on left if not given the full width for this segment
+		copy(key[n-len(s):n], s)
+	}
+	return key[:n]
+}
+
+// Format the args passed into the key format - will panic if the arguments passed to not match the length
+// of the segment to which they correspond. When called with no arguments returns the raw prefix (useful as a start
+// element of the entire keys space when sorted lexicographically)
```
Liamsi: Nit: comments should end with a full stop (here and everywhere else). Don't bother, I'll add these.

key_format.go
```diff
@@ -0,0 +1,127 @@
...
+func (kf *KeyFormat) Prefix() string {
+	return string([]byte{kf.prefix})
+}
+
+func scan(a interface{}, value []byte) {
+	switch v := a.(type) {
+	case *int64:
+		*v = int64(binary.BigEndian.Uint64(value))
+	case *uint64:
+		*v = binary.BigEndian.Uint64(value)
```
Liamsi: Question: Maybe I'm overlooking sth obvious here but why are there only two cases (`int64` and `uint64`) when reading in the value? When formatting the we allow for `uint64`, `uint`, `int64`, `int`.

key_format.go
```diff
@@ -0,0 +1,127 @@
...
+
+// Return the prefix as a string
+func (kf *KeyFormat) Prefix() string {
+	return string([]byte{kf.prefix})
+}
+
+func scan(a interface{}, value []byte) {
+	switch v := a.(type) {
+	case *int64:
+		*v = int64(binary.BigEndian.Uint64(value))
```
Liamsi: Will this still behave as intended when this overflows?

silasdavis:
@Liamsi I'll make these changes shortly - thanks.

key_format.go
silasdavis: It's convenient to be able to call `kf.Key(0, 1, 2)` and not have to cast each argument, `kf.Key(0)` is the example that exists, and in tests.

key_format.go
```diff
@@ -0,0 +1,127 @@
...
+		binary.BigEndian.PutUint64(bs, v)
+		return bs
+	case uint:
+		return format(uint64(v))
+	case int64:
+		bs := make([]byte, 8)
+		binary.BigEndian.PutUint64(bs, uint64(v))
+		return bs
+	case int:
+		return format(int64(v))
```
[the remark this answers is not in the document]
silasdavis: When I read that I thought, oh really, how did I not know you could do that, but sadly you cannot. The compile-time type of `v` becomes `interface{}` as soon as you combine types which cannot be type converted.

You must have been confusing this with a language that has type system :)

I can make it a bit more consistent though:
```go
func format(a interface{}) []byte {
	switch v := a.(type) {
	case uint64:
		bs := make([]byte, 8)
		binary.BigEndian.PutUint64(bs, v)
		return bs
	case uint:
		return format(uint64(v))
	case int64:
		return format(uint64(v))
	case int:
		return format(uint64(v))
	case []byte:
		return v
	default:
		panic(fmt.Errorf("KeyFormat format() does not support formatting value of type %T: %v", a, a))
	}
}
```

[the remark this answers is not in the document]
ValarDragon: I'd prefer not to do recursive calls, but instead have a second "formatuint64" function which all of these sub-cases call. A second interface conversions seems like an unnecessary performance hit, given that this may be called a ton.

key_format.go
silasdavis: Yeah it's fine the MSB just becomes the sign bit - I've added a test to demonstrate. Provided the user is consistent with using int64 for signed numbers.

silasdavis:
@Liamsi changes done

Liamsi:
Awesome! Merging and will prep a release this week.

PR #108, opened 2018-08-30
Update to CircleCI 2.0

silasdavis:


Makefile
```diff
@@ -1,11 +1,11 @@
-GOTOOLS := github.com/mitchellh/gox \
```
silasdavis: gox not used anywhere

PR #118, opened 2018-11-08
update tendermint dependency and update changelog

Liamsi:
update to tendermint v0.26.0


Liamsi:
Tendermint was updated in some other PR to https://github.com/tendermint/iavl/blob/de0740903a67b624d887f9055d4c60175dcfa758/Gopkg.toml#L11

PR #123, opened 2019-02-02
expose versions field to support pruning

ackratos:
to support https://github.com/cosmos/cosmos-sdk/issues/3480

### DIFF
--- a/mutable_tree.go
+++ b/mutable_tree.go
@@ -270,7 +270,7 @@ func (tree *MutableTree) LoadVersionForOverwriting(targetVersion int64) (int64, 
 	if err != nil {
 		return latestVersion, err
 	}
-	tree.deleteVersionsFrom(targetVersion+1)
+	tree.deleteVersionsFrom(targetVersion + 1)
 	return targetVersion, nil
 }
 
@@ -500,4 +500,8 @@ func (tree *MutableTree) addOrphans(orphans []*Node) {
 		}
 		tree.orphans[string(node.hash)] = node.version
 	}
+}
+
+func (tree *MutableTree) GetVersions() map[int64]bool {
+	return tree.versions
 }


Liamsi:
Thanks a lot for looking into this @ackratos! Do you mind sharing your implementation of the PruningStrategy to better understand this requirement?  

Also, shouldn't the interface in cosmos/cosmos-sdk#3480 rather be:
```go
func (tree *MutableTree) Prune(ps PruningStrategy) {/* as in cosmos-sdk#3480 */}
```
and require changes here (and not in the sdk)?

ackratos:
@Liamsi sorry for the delayed reply, I have had a whole week vacation.

Apart from https://github.com/cosmos/cosmos-sdk/issues/3480, this is mainly mean to support state sync (https://github.com/tendermint/tendermint/issues/828)

I haven't realise cosmos recently refactored the prune strategy part :( (https://github.com/cosmos/cosmos-sdk/pull/2985/). I already finished change on my earlier cosmos fork (will raise a PR to latest cosmos within this week):
```
// PruningStrategy specfies how old states will be deleted over time
type PruningStrategy interface {
	Prune(version, latestVersion int64) bool
}

// PruneSyncable means only those states not needed for state syncing will be deleted (keeps last 100 + every 10000th)
type PruneSyncable struct {
	// How many old versions we hold onto.
	// A value of 0 means keep no recent states.
	NumRecent int64

	// This is the distance between state-sync waypoint states to be stored.
	// See https://github.com/tendermint/tendermint/issues/828
	// A value of 1 means store every state.
	// A value of 0 means store no waypoints. (node cannot assist in state-sync)
	// By default this value should be set the same across all nodes,
	// so that nodes can know the waypoints their peers store.
	StoreEvery int64
}

func (strategy PruneSyncable) Prune(version, latestVersion int64) bool {
	return (latestVersion-version > strategy.NumRecent) && (version%strategy.StoreEvery != 0)
}

// PruneEverything means all saved states will be deleted, storing only the current state
type PruneEverything struct{}

func (strategy PruneEverything) Prune(version, latestVersion int64) bool {
	return true
}

// PruneNothing means all historic states will be saved, nothing will be deleted
type PruneNothing struct{}

func (strategy PruneNothing) Prune(version, latestVersion int64) bool {
	return false
}

// Implements Committer.
func (st *iavlStore) Commit() CommitID {
	// Save a new version.
	hash, version, err := st.tree.SaveVersion()
	if err != nil {
		// TODO: Do we want to extend Commit to allow returning errors?
		panic(err)
	}

	// Release an old version of history, if not a sync waypoint.
	for v, _ := range st.tree.GetVersions() {
		if st.pruningStrategy.Prune(v, version) {
			st.tree.DeleteVersion(v)
		}
	}

	return CommitID{
		Version: version,
		Hash:    hash,
	}
}
```

Our usage is a little more complicated than cosmos syncable way point. (which related to num of versions recent and fixed number of blocks). We introduced `breathe block` - the first block of each UTC day and snapshot tx/msg execution result in memory between two breathe block (to ease disk r/w pressure).  We want keep all breathe block versions. Below is strategy implemented in our app level.
```
type KeepRecentAndBreatheBlock struct {
	breatheBlockInterval int64

 	// Keep recent number blocks in case of rollback
	numRecent int64

 	blockStore *blockchain.BlockStore

 	blockStoreInitializer sync.Once
}

func NewKeepRecentAndBreatheBlock(breatheBlockInterval, numRecent int64) *KeepRecentAndBreatheBlock {
	return &KeepRecentAndBreatheBlock{
		breatheBlockInterval: breatheBlockInterval,
		numRecent:            numRecent,
	}
}

func (strategy KeepRecentAndBreatheBlock) Prune(version, latestVersion int64) bool {

 	// only at this time block store is initialized!
	strategy.blockStoreInitializer.Do(func() {
		strategy.blockStore = cosmossrv.BlockStore
	})

 	if version == 1 {
		return true
	} else {
		lastBlock := strategy.blockStore.LoadBlock(version - 1)
		block := strategy.blockStore.LoadBlock(version)

 		if strategy.breatheBlockInterval > 0 {
			return version%strategy.breatheBlockInterval != 0
		} else {
			return utils.SameDayInUTC(lastBlock.Time, block.Time)
		}
	}
}
```

marbar3778:
closing this PR as we introduced pruning into IAVL via #158

PR #125, opened 2019-02-12
update tendermint dependency to v0.30

Liamsi:
update to tendermint v0.30 and bump version to 0.12.1 (for immediate release)

Dependency to  tendermint v0.27.0-dev1 was causing some inconvenience when developing an app which has the sdk and iavl as a dependency.

cc @alessio

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -8,7 +8,7 @@
 
 [[constraint]]
   name = "github.com/tendermint/tendermint"
-  version = "v0.27.0-dev1"
+  version = "v0.30.0"
 
 [prune]
   go-tests = true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.12.1 (February 12, 2019)
+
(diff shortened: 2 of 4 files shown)


Gopkg.toml
```diff
@@ -8,7 +8,7 @@
 
 [[constraint]]
   name = "github.com/tendermint/tendermint"
-  version = "v0.27.0-dev1"
+  version = "v0.30.0"
```
melekes: why iavl depends on Tendermint again?

Liamsi: Because it uses some parts of what previously tmlibs: https://github.com/tendermint/iavl/commit/c8114f186987fa8c0489afd00ae40846eee9fb34#diff-836546cc53507f6b2d581088903b1785
more dependencies introduced due to:
https://github.com/tendermint/iavl/pull/105 (https://github.com/tendermint/tendermint/pull/1912, https://github.com/tendermint/tendermint/pull/2298)

Liamsi: also ref: https://github.com/tendermint/iavl/issues/46

PR #127, opened 2019-02-15
Iaviewer command to visualize iavl database from leveldb

ethanfrey:
Our nodes fell out of consensus due to inconsistent app hashes, but after locating the transaction that caused it and finding no obvious issues in the execution path, we had no real tools to figure out what the problem was.

I add a new command `cmd/iaviewer` that can also be installed via `make install` to allow forensic analysis and comparison of leveldb-backed iavl trees

We also allow an optional third argument of a version number to specify an older version of the state to use, rather than the latest one. This is interesting if you want to see what things looked like right _before_ it all went wrong.

Done:

`iaviewer data <path to db> [version]`

Prints out version, hash, size as well as every key along with a hash of the data (to keep sizes small). We try to convert as much of the key to ascii if possible, in case the app is using human-readble strings. The data is hashed, as the only thing we cared about at this point was if one key had a different value on two different nodes.

`iaviewer shape <path to db> [version]`

It is possible for two trees to have a different hash even with the same data if the insert order is different. This attempts to visualize the tree structure to locate the difference.

`iaviewer versions <path to db>`

Lists all available versions in the db, so we can compare snapshots from different points in time.


ethanfrey:
@Liamsi Now that cosmos mainnet is live and the launch party is over, maybe you have time to review this?

I'm happy to make changes as needed, but some feedback would be nice

Liamsi:
Will do tomorrow!

cmd/iaviewer/main.go
```diff
@@ -0,0 +1,164 @@
...
+		fmt.Println("Usage: iaviewer <data|shape|versions> <leveldb dir> [version number]")
+		os.Exit(1)
+	}
+
+	version := 0
+	if len(args) == 3 {
+		var err error
+		version, err = strconv.Atoi(args[2])
+		if err != nil {
+			fmt.Printf("Invalid version number: %s\n", err)
```
Liamsi: Same as above: this should probably go to `os.Stderr`.

cmd/iaviewer/main.go
```diff
@@ -0,0 +1,164 @@
...
+			os.Exit(3)
```
Liamsi: Nit: Besides indicating an error, is it useful to exit with a different non-zero code per failure? Here 2, below 3? 
I would suggest to exit with `os.Exit(1)` in all failure cases.

mutable_tree.go
```diff
@@ -45,6 +46,18 @@ func (tree *MutableTree) VersionExists(version int64) bool {
 	return tree.versions[version]
 }
 
+// AvailableVersions returns all available versions
+func (tree *MutableTree) AvailableVersions() []int {
```
Liamsi: It would be great to have a little test for `AvailableVersions`. I can add it it you do not have time.

cmd/iaviewer/main.go
```diff
@@ -0,0 +1,164 @@
...
+	for _, b := range id {
+		if b < 0x20 || b >= 0x80 {
+			return strings.ToUpper(hex.EncodeToString(id))
+		}
+	}
+	return string(id)
+}
+
+func PrintShape(tree *iavl.MutableTree) {
+	// shape := tree.RenderShape("  ", nil)
```
Liamsi: Delete commented out code. 

Oh wait, shouldn't this be the default and `nodeEncoder` below makes some assumptions about the key?

cmd/iaviewer/main.go
```diff
@@ -0,0 +1,164 @@
...
+	if err != nil {
+		return nil, err
+	}
+	// PrintDbStats(db)
+	return db, nil
+}
+
+func PrintDbStats(db dbm.DB) {
+	// stats, _ := json.MarshalIndent(db.Stats(), "", "  ")
+	// fmt.Println(string(stats))
```
Liamsi: deleted commented out code

cmd/iaviewer/main.go
```diff
@@ -0,0 +1,164 @@
...
+	cut := strings.LastIndex(dir, "/")
+	if cut == -1 {
+		return nil, fmt.Errorf("Cannot cut paths on %s", dir)
+	}
+	name := dir[cut+1:]
+	db, err := dbm.NewGoLevelDB(name, dir[:cut])
+	if err != nil {
+		return nil, err
+	}
+	// PrintDbStats(db)
```
Liamsi: delete

cmd/iaviewer/main.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+}
+
+func OpenDb(dir string) (dbm.DB, error) {
+	if strings.HasSuffix(dir, ".db") {
+		dir = dir[:len(dir)-3]
+	} else if strings.HasSuffix(dir, ".db/") {
+		dir = dir[:len(dir)-4]
+	} else {
+		return nil, fmt.Errorf("Database directory must end with .db")
```
Liamsi: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

cmd/iaviewer/main.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	// TODO: doesn't work on windows!
+	cut := strings.LastIndex(dir, "/")
+	if cut == -1 {
+		return nil, fmt.Errorf("Cannot cut paths on %s", dir)
```
Liamsi: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

immutable_tree.go
```diff
@@ -38,6 +38,50 @@ func (t *ImmutableTree) String() string {
...
+// RenderShape provides a nested tree shape, ident is prepended in each level
+// Returns an array of strings, one per line, to join with "\n" or display otherwise
+func (t *ImmutableTree) RenderShape(indent string, encoder func([]byte, int, bool) string) []string {
+	if encoder == nil {
+		encoder = defaultNodeEncoder
+	}
+	return t.renderNode(t.root, indent, 0, encoder)
+}
+
+// defaultNodeEncoder takes an id (hash, or key for leaf notes) and whether or not this is a leaf node
```
Liamsi: s/notes/nodes

immutable_tree.go
```diff
@@ -38,6 +38,50 @@ func (t *ImmutableTree) String() string {
 	return "Tree{" + strings.Join(leaves, ", ") + "}"
 }
 
+// RenderShape provides a nested tree shape, ident is prepended in each level
+// Returns an array of strings, one per line, to join with "\n" or display otherwise
+func (t *ImmutableTree) RenderShape(indent string, encoder func([]byte, int, bool) string) []string {
```
Liamsi: At least add a comment about the `encoder`. Alternatively, introduce a commented type alias. Something along the lines of: `type NodeEncoder func(id []byte, depth int, isLeaf bool) string` or an interface:
```go
type NodeEncoder interface {
	Encode(id []byte, depth int, isLeaf bool) string
}
```

mutable_tree.go
```diff
@@ -45,6 +46,18 @@ func (tree *MutableTree) VersionExists(version int64) bool {
...
+	res := make([]int, 0, len(tree.versions))
+	for i, v := range tree.versions {
+		if v {
+			res = append(res, int(i))
+		}
+	}
+	sort.Sort(sort.IntSlice(res))
```
Liamsi: Shouldn't one sort be enough here?

mutable_tree.go
```diff
@@ -45,6 +46,18 @@ func (tree *MutableTree) VersionExists(version int64) bool {
 	return tree.versions[version]
 }
 
+// AvailableVersions returns all available versions
```
Liamsi: Nit: could add that the versions are sorted.

cmd/iaviewer/main.go
```diff
@@ -0,0 +1,164 @@
...
+		return false
+	})
+}
+
+// parseWeaveKey assumes a separating : where all in front should be ascii,
+// and all afterwards may be ascii or binary
+func parseWeaveKey(key []byte) string {
+	cut := bytes.IndexRune(key, ':')
+	if cut == -1 {
+		return encodeId(key)
```
Liamsi: OK. I see the fallback here works even if the key is not in that particular format.

cmd/iaviewer/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+```shell
+iaviewer data ./bns-a.db 190258 > a-cur.data
+iaviewer data ./bns-b.db 190258 > b-cur.data
+
+diff a-cur.data b-cur.data
+```
+
+Hmmm... everything is the same, except the hash. Odd...
+So odd that I [wrote an article about it](https://medium.com/@ethan.frey)
```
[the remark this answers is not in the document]
ethanfrey: Yup, I have a draft. Meant to publish yesterday. Will update this.

cmd/iaviewer/README.md
```diff
@@ -0,0 +1,136 @@
...
+Once this is merged into the offical repo, master, you should be able to do:
+
+```shell
+go get github.com/tendermint/iavl
+cd ${GOPATH}/src/github.com/tendermint/iavl
+make get_vendor_deps
+make install
+```
+
+However, as this PR lives in another repo on another branch, it is a bit trickier to build.
```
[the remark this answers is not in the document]
ethanfrey: Nice.

cmd/iaviewer/main.go
```diff
@@ -0,0 +1,164 @@
...
+
+// TODO: make this configurable?
+const (
+	DefaultCacheSize int = 10000
+)
+
+func main() {
+	args := os.Args[1:]
+	if len(args) < 2 || (args[0] != "data" && args[0] != "shape" && args[0] != "versions") {
+		fmt.Println("Usage: iaviewer <data|shape|versions> <leveldb dir> [version number]")
```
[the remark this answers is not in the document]
ethanfrey: Sure

cmd/iaviewer/main.go
ethanfrey: Okay, too verbose I guess.
I will normalize to os.Exit(1) for everything

immutable_tree.go
ethanfrey: Will do.

mutable_tree.go
ethanfrey: There is only one sort.
`sort.IntSlice` is just a [type to enable sorting []int](https://godoc.org/sort#IntSlice)
`sort.Sort` actually does the work.

mutable_tree.go
ethanfrey: I will look to do it a well

cmd/iaviewer/main.go
ethanfrey: Do you have a specific ascii prefix format for cosmos-sdk keys?
I am happy to add that, and maybe a cli flag to select decoder.
I just know weave, where there is readable string followed by (mostly) binary data

ethanfrey:
Thanks for the feedback. I will likely get to this Wednesday

mutable_tree.go
Liamsi: Thanks for the clarification!

cmd/iaviewer/main.go
ethanfrey: I will make a flag here

ethanfrey:
Made most requested cleanups.
Missing feature is a flag to set how we display the nodes.

I can use `nil` for only hex-display.
And add a flag `--weave` to show the weave-specific prefixes.
I'd be happy to add an `--sdk` flag to show cosmos-sdk specific readable text.

Can you point me to how the data stored in substores looks like? Just a few example keys (from root), not the whole code.

ethanfrey:
@Liamsi can you review again to let me know if this addresses all your comments.

Also please let me know if current cli format is good enough or if you want some application specific flags to control parsing/display. And ideally a suggestion as to what you want to see.

Thanks.

Liamsi:
I think hex-display is just fine. The more we can avoid adding some project specific code here the better (iavl shouldn't now about how weave encodes things, or it even shouldn't know how the SDK does it).

cmd/iaviewer/main.go
melekes: Looks like this still needs to be done. Otherwise it assumes weave key format, right?

melekes:
```
nodedb.go:333:12: ndb.batch.Close undefined (type db.Batch has no field or method Close) (typecheck)
	ndb.batch.Close()
```

circle is failing with ^ but seems not related.

ethanfrey:
Looks related to https://github.com/tendermint/iavl/commit/cd1feb8afff203773e848d256184b9e97af18b90

I think the dependency versions changed since I pushed this. I can try to rebase and clean it up... I kinda gave up on anything moving here

melekes:
> I kinda gave up on anything moving here

Sorry it took us so long to merge.

>  can try to rebase and clean it up

That would be great 👍

ethanfrey:
Oof, merged already. Did this break develop?

Ping me here if I should clean it up

melekes:
> Did this break develop?

I guess

> Ping me here if I should clean it up

only if you want to & have time

Liamsi:
Fixed here: https://github.com/tendermint/iavl/pull/146

ethanfrey:
Thanks for that @Liamsi

PR #130, opened 2019-03-12
close batch after write in nodedb

Stumble:
This one is related to https://github.com/tendermint/tendermint/pull/3397, that writebatch should be closed if no longer needed. This one could be the major part of memory leaking when use CLevelDB, since it's in the hot loop.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -138,7 +138,7 @@
   version = "v0.14.1"
 
 [[projects]]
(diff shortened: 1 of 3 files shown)


cwgoes:
Looks like we should patch this pretty quickly downstream.

Liamsi:
Boy, I accidentally merged this directly into master (instead of develop). Did notice the base branch.. Will revert, reopen, change base branch, and create a proper release containing that change (today).

Stumble:
no problem

PR #131, opened 2019-03-13
Revert "close batch after write in nodedb"

Liamsi:
Reverts tendermint/iavl#130
See: https://github.com/tendermint/iavl/pull/130#issuecomment-472456716


Liamsi:
I'll actually create the release directly on master and merge master back to develop instead of creating another commit (the revert one) on master.

PR #134, opened 2019-03-13
Fix commit history on master

Liamsi:
master and develop diverged here. Merge develop into master (and then master back into develop)
image


Liamsi:
resolving the merge conflicts here created a commit on develop. Now we are here:
image

Liamsi:
If we merge this into master, this will create an even history on both branches.

PR #145, opened 2019-06-24
LoadVersionForOverwriting should transit orphans to non-orphans

ridenaio:
LoadVersionForOverwriting should transit orphans to non-orphans for overwriting version and remove nodes, which become useless

Fixes https://github.com/tendermint/iavl/issues/141

### DIFF
--- a/mutable_tree.go
+++ b/mutable_tree.go
@@ -574,6 +574,7 @@ func (tree *MutableTree) deleteVersionsFrom(version int64) error {
 	tree.ndb.resetLatestVersion(newLatestVersion)
 	return nil
 }
+
 // deleteNodes deletes all nodes which have greater version than current, because they are not useful anymore
 func (tree *MutableTree) deleteNodes(version int64, hash []byte) {
 	if len(hash) == 0 {
@@ -597,7 +598,6 @@ func (tree *MutableTree) deleteNodes(version int64, hash []byte) {
 		}
 	}
 }
-
 
 // Rotate right and return the new node and orphan.
 func (tree *MutableTree) rotateRight(node *Node) (*Node, *Node) {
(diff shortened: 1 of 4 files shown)


ridenaio:
@Liamsi @jaekwon what about this PR?

sidenaio:
We've rebased our branch to your latest master. @erikgrinaker @AdityaSripal, please take a look.

tree_test.go
```diff
@@ -1373,3 +1373,34 @@ func BenchmarkTreeLoadAndDelete(b *testing.B) {
 		}
 	})
 }
+func TestLoadVersionForOverwritingCase2(t *testing.T) {
```
erikgrinaker, line 1376: We should check that all keys have the expected values for the various versions.

mutable_tree.go
```diff
@@ -567,6 +575,30 @@ func (tree *MutableTree) deleteVersionsFrom(version int64) error {
 	return nil
 }
 
+// deleteNodes deletes all nodes which have greater version than current, because they are not useful anymore
+func (tree *MutableTree) deleteNodes(version int64, hash []byte) {
+	if len(hash) == 0 {
+		return
+	}
```
erikgrinaker, line 582: I don't think this is correct. We should skip `nil`, but empty trees will have a `[]byte{}` root node, and these should be removed as well. Would be great with a test-case for this.

erikgrinaker:
Hi @sidenaio! Will you be able to follow up on my comments above? Otherwise I'd be happy to fix the remaining minor issues myself and get this across the line.

mutable_tree.go
sidenaio: Empty root doesn't present as a node in DB so it can be skipped.

sidenaio: Moreover tree.ndb.GetNode will panic with zero-length hash

erikgrinaker: That seems to be the case. There _is_ an `[]byte{}` entry for the empty root in the database, but it appears to be correctly removed when calling `LoadVersionForOverwriting`.

PR #147, opened 2019-06-27
Hack1 merge with latest master

Sriep:


marbar3778:
@Sriep, Was this mistakenly opened? If not, could you provide a description of the aim of this PR?

PR #148, opened 2019-06-27
Implement LazyLoadVersion

alexanderbez:
Implement lazy loading of a given version.

ref: https://github.com/cosmos/cosmos-sdk/issues/4326

### DIFF
--- a/mutable_tree.go
+++ b/mutable_tree.go
@@ -264,8 +264,6 @@ func (tree *MutableTree) LoadVersion(targetVersion int64) (int64, error) {
 		return 0, err
 	}
 
-	fmt.Println("NUM ROOTS:", len(roots))
-
 	if len(roots) == 0 {
 		return 0, nil
 	}
@@ -274,8 +272,6 @@ func (tree *MutableTree) LoadVersion(targetVersion int64) (int64, error) {
 
 	var latestRoot []byte
 	for version, r := range roots {
-		fmt.Println("Iterating over root:", version)
-
 		tree.versions[version] = true
 		if version > latestVersion && (targetVersion == 0 || version <= targetVersion) {
 			latestVersion = version
--- a/mutable_tree.go
+++ b/mutable_tree.go
@@ -224,21 +224,54 @@ func (tree *MutableTree) Load() (int64, error) {
 	return tree.LoadVersion(int64(0))
 }
 
+// LazyLoadVersion attempts to lazy load only the specified target version
+// without loading previous roots/versions. If the targetVersion is invalid, an
+// error is returned along with the latest version.
+func (tree *MutableTree) LazyLoadVersion(targetVersion int64) (int64, error) {
+	latestVersion := tree.ndb.getLatestVersion()
+	if targetVersion == 0 || latestVersion < targetVersion {
+		return latestVersion, fmt.Errorf("wanted to load target %d but only found up to %d", targetVersion, latestVersion)
+	}
+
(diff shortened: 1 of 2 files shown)


alexanderbez:
Let me know if this looks correct @Liamsi or if there are any cases I need to be aware of. I've successfully tested this in Gaia.

mutable_tree.go
```diff
@@ -224,21 +224,56 @@ func (tree *MutableTree) Load() (int64, error) {
 	return tree.LoadVersion(int64(0))
 }
 
+// LazyLoadVersion attempts to lazy load only the specified target version
+// without loading previous roots/versions. If the targetVersion is invalid, an
+// error is returned along with the latest version.
+func (tree *MutableTree) LazyLoadVersion(targetVersion int64) (int64, error) {
+	if targetVersion == 0 {
```
melekes:
```suggestion
	if targetVersion <= 0 {
```

alexanderbez:
@Liamsi that's a good point. And as you've brought up, I can't think of a reason why one would use `LoadVersion()` over `LazyLoadVersion()`. It's not immediately clear to me why you'd need to load all the roots -- I was hoping you'd have more context or a clue here.

If we are in agreement, how about we replace `LoadVersion` with `LazyLoadVersion`?

Liamsi:
Sorry, I don't have any more context on this. Maybe @jaekwon (or @ethanfrey) have?

> If we are in agreement, how about we replace LoadVersion with LazyLoadVersion?

As far as I understand it does not make sense to load all the roots. I'm OK with replacing `LoadVersion` with `LazyLoadVersion`.

zmanian:
Okay I am not 100% sure we should replace `LoadVersion()`

LoadVersion is not DOSable. LazyLoadVersion is. Ie. Once you load a tree with `LoadVersion()` the tree cannot be DOSed by creating a query pattern that requires loading lots of different tree versions to bring the node to a halt.

`LoadVersion()` can be used safely with a pruning node and queries against snapshots.

`LazyLoadVersion()` opts a node operator into new risks if the tendermint rpc is world accessible.

alexanderbez:
Valid concerns @zmanian. From doing another pass through of the IAVL implementation and it's relation/usage in the SDK's multistore (I've done this now a few times), I don't quite see where a DOS attack vector could occur? I could very well be wrong here. Here's what I can tell:

- Height queries always call `ndb.getRoot()` regardless
- Root node lookups `ndb.getNode()` can be cached
- The only difference between `LoadVersion` and `LazyLoadVersion` is the former reads all the roots AND sets them in `tree.versions[version]`.
- I don't see how `tree.versions` would correlate to any DOSing or the prevention thereof. It's mainly only used when it comes to deleting.

alexanderbez:
Updated @zmanian. Also fixed the case when 0 is provided -- it loads the latest version by default. This was the previous behavior. I also need to update the semantics of `Load`.

alexanderbez:
Ok so actually taking a stab at replacing `LoadVersion` with `LazyLoadVersion`, I see why the need for the original logic exists:

The original `LoadVersion()` loads all the roots and as a result, when you load previous committed/saved versions and then try to overwrite a value and save, you'll get an error (as you should). I'm sure there are other cases similar to this.

That being the case, I'm going to leave `LoadVersion` as-is and keep `LazyLoadVersion` and update the godoc to state it should only be used in cases where only reads are necessary, not writes.

PR #149, opened 2019-07-01
Remove use of colors

marbar3778:
- Colors.go was removed from tendermint

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ GOTOOLS := github.com/golangci/golangci-lint/cmd/golangci-lint
 
 PDFFLAGS := -pdf --nodefraction=0.1
 
-LINKER_FLAG := -ldflags "-X main.COLOR_ON=on"
+LDFLAGS := -ldflags "-X main.COLOR_ON=on"
 
 all: lint test install
 
@@ -12,7 +12,7 @@ install:
 ifeq ($(COLORS_ON),)
 	go install ./cmd/iaviewer
 else
-	go install $(LINKER_FLAG) ./cmd/iaviewer
+	go install $(LDFLAGS) ./cmd/iaviewer
 endif
 
 test:
--- a/Makefile
+++ b/Makefile
@@ -2,10 +2,18 @@ GOTOOLS := github.com/golangci/golangci-lint/cmd/golangci-lint
 
 PDFFLAGS := -pdf --nodefraction=0.1
 
+LINKER_FLAG := -ldflags "-X main.COLOR_ON=on"
+
 all: lint test install
 
+
+
 install:
+ifeq ($(COLORS_ON),)
 	go install ./cmd/iaviewer
+else
+	go install $(LINKER_FLAG) ./cmd/iaviewer
+endif
 
 test:
 	go test -v --race
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ GOTOOLS := github.com/golangci/golangci-lint/cmd/golangci-lint
 
 PDFFLAGS := -pdf --nodefraction=0.1
 
-LDFLAGS := -ldflags "-X main.COLOR_ON=on"
+LDFLAGS := -ldflags "-X TENDERMINT_IAVL_COLORS_ON=on"
 
 all: lint test install
 
--- a/util.go
+++ b/util.go
@@ -145,8 +145,11 @@ func Cyan(args ...interface{}) string {
 	return treatAll(ANSIFgCyan, args...)
 }
 
+// Coloredbytes takes in the byte that you would like to show as a string and byte
+// and display them in different colors, if `TENDERMINT_IAVL_COLORS_ON` is set to on.
 func ColoredBytes(data []byte, textColor, bytesColor func(...interface{}) string) string {
-	colors := os.Getenv("COLORS_ON")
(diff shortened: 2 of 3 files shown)


melekes:
Are you sure we don't need it? Maybe move them from tendermint-core instead of removing?

marbar3778:
Is there a benefit to having colors in the terminal? If so, I can add that section into here.

alessio:
There should be a output formatting function that may conditionally enable colours on screen IMHO

Makefile
```diff
@@ -2,10 +2,18 @@ GOTOOLS := github.com/golangci/golangci-lint/cmd/golangci-lint
 
 PDFFLAGS := -pdf --nodefraction=0.1
 
+LINKER_FLAG := -ldflags "-X main.COLOR_ON=on"
```
alessio:
```suggestion
LDFLAGS := -ldflags "-X main.COLOR_ON=on"
```

Makefile
```diff
@@ -2,10 +2,18 @@ GOTOOLS := github.com/golangci/golangci-lint/cmd/golangci-lint
...
+
 all: lint test install
 
+
+
 install:
+ifeq ($(COLORS_ON),)
 	go install ./cmd/iaviewer
+else
+	go install $(LINKER_FLAG) ./cmd/iaviewer
```
alessio:
```suggestion
	go install $(LDFLAGS) ./cmd/iaviewer
```

util.go
```diff
@@ -102,3 +104,61 @@ func sortByteSlices(src [][]byte) [][]byte {
...
+func Blue(args ...interface{}) string {
+	return treatAll(ANSIFgBlue, args...)
+}
+
+func Cyan(args ...interface{}) string {
+	return treatAll(ANSIFgCyan, args...)
+}
+
+func ColoredBytes(data []byte, textColor, bytesColor func(...interface{}) string) string {
+	colors := os.Getenv("COLORS_ON")
```
alessio: Mmmmh... If we really want go for an environment variable let's at least prefix it with `IAVL_`:

```suggestion
	colors := os.Getenv("IAVL_COLORS_ON")
```

util.go
```diff
@@ -102,3 +104,61 @@ func sortByteSlices(src [][]byte) [][]byte {
...
+
+func Blue(args ...interface{}) string {
+	return treatAll(ANSIFgBlue, args...)
+}
+
+func Cyan(args ...interface{}) string {
+	return treatAll(ANSIFgCyan, args...)
+}
+
+func ColoredBytes(data []byte, textColor, bytesColor func(...interface{}) string) string {
```
alessio: Please add a godoc comment to explain what this function does, plus document the environment variable usage.

util.go
```diff
@@ -102,3 +104,64 @@ func sortByteSlices(src [][]byte) [][]byte {
...
+
+func Blue(args ...interface{}) string {
+	return treatAll(ANSIFgBlue, args...)
+}
+
+func Cyan(args ...interface{}) string {
+	return treatAll(ANSIFgCyan, args...)
+}
+
+// Coloredbytes takes in the byte that you would like to show as a string and byte
```
alessio:
```suggestion
// ColoredBytes takes in the byte that you would like to show as a string and byte
```

util.go
```diff
@@ -102,3 +104,64 @@ func sortByteSlices(src [][]byte) [][]byte {
...
+// and display them in different colors, if `TENDERMINT_IAVL_COLORS_ON` is set to on.
```
alessio:
```suggestion
// and display them in a human readable format.
// If the environment variable `TENDERMINT_IAVL_COLORS_ON` is set to a non empty string,
// strings and bytes will be displayed with different colours. 
```

util.go
```diff
@@ -102,3 +104,64 @@ func sortByteSlices(src [][]byte) [][]byte {
...
+}
+
+func Cyan(args ...interface{}) string {
+	return treatAll(ANSIFgCyan, args...)
+}
+
+// ColoredBytes takes in the byte that you would like to show as a string and byte
+// and will display them in different colors for readability, if `TENDERMINT_IAVL_COLORS_ON` is set to on.
+func ColoredBytes(data []byte, textColor, bytesColor func(...interface{}) string) string {
+	// If TENDERMINT_IAVL_COLORS_ON is set to on then colors will be used for bytes and strings
```
alessio: This is wrong. To enable colours `TENDERMINT_IAVL_COLORS_ON` needs just to be a non-empty string

util.go
```diff
@@ -102,3 +104,64 @@ func sortByteSlices(src [][]byte) [][]byte {
...
+}
+
+func Cyan(args ...interface{}) string {
+	return treatAll(ANSIFgCyan, args...)
+}
+
+// ColoredBytes takes in the byte that you would like to show as a string and byte
+// and will display them in a human readable format.
+func ColoredBytes(data []byte, textColor, bytesColor func(...interface{}) string) string {
+	// If TENDERMINT_IAVL_COLORS_ON is set to a non-empty string then colors will be used for bytes and strings
```
alessio:
```suggestion
	// If the environment variable TENDERMINT_IAVL_COLORS_ON is set to a non-empty string then different colors will be used for bytes and strings.
```

alessio: Plus, please incorporate this in the godoc comment. Then we are done :)

PR #150, opened 2019-07-02
Storing intermediary IAVL versions in memory and not to disk

mattkanwisher:
Storing intermediary IAVL versions in memory and not to disk

Motivation: Both Cosmos and Loom Network save an IAVL version per block, then go back and delete these versions. So you have constant churn on the IAVL and underlying Leveldb database. When realistically what you want is to only store every X Blocks.

At Berlin Tendermint Conference, Zaki and I surmised a plan where new versions are in memory, while still pointing back to nodes on disk to prevent needing to load entire IAVL into main memory. Loom IAVL tree is around 256gb so this is not feasible otherwise.

Usage

OLD Code would be like

```go
hash, version, err := s.tree.SaveVersion()
```

New Caller code would look like

```go
	oldVersion := s.Version()
  	var version int64
 	var hash []byte
 	//Every X versions we should persist to disk
 	if s.flushInterval == 0 || ((oldVersion+1)%s.flushInterval == 0) {
 		if s.flushInterval != 0 {
 			log.Error(fmt.Sprintf("Flushing mem to disk at version %d\n", oldVersion+1))
 			hash, version, err = s.tree.FlushMemVersionDisk()
 		} else {
 			hash, version, err = s.tree.SaveVersion()
 		}
 	} else {
 		hash, version, err = s.tree.SaveVersionMem()
 	}
```

FlushMemVersionDisk:
Flushes the current memory version to disk

SaveVersionMem:
Saves the current tree to memory instead of disk and gives you back an apphash

This is an opt in feature, you have to call new apis to get it. 
We also have a PR that demonstrates its usage https://github.com/loomnetwork/loomchain/pull/1232/files

We are now commiting every 1000 blocks, so we store 1000x less. Also we have signficant improves in IO at least double from not having to Prune old versions of the IAVL Tree

### DIFF
--- a/mutable_tree.go
+++ b/mutable_tree.go
@@ -16,6 +16,7 @@ type MutableTree struct {
 	*ImmutableTree                  // The current, working tree.
 	lastSaved      *ImmutableTree   // The most recently saved tree.
 	orphans        map[string]int64 // Nodes removed by changes to working tree.
+	memVersions    map[int64]bool   // The previous, saved versions of the tree in mem.
 	versions       map[int64]bool   // The previous, saved versions of the tree.
 	ndb            *nodeDB
 }
@@ -29,6 +30,7 @@ func NewMutableTree(db dbm.DB, cacheSize int) *MutableTree {
 		ImmutableTree: head,
 		lastSaved:     head.clone(),
 		orphans:       map[string]int64{},
+		memVersions:   map[int64]bool{},
 		versions:      map[int64]bool{},
 		ndb:           ndb,
 	}
@@ -364,10 +366,36 @@ func (tree *MutableTree) GetVersioned(key []byte, version int64) (
 	return -1, nil
 }
 
+// SaveVersionMem saves a new tree version to disk, based on the current state of
+// the tree. Returns the hash and new version number.
+func (tree *MutableTree) SaveVersionMem() ([]byte, int64, error) {
+	return tree.saveVersion(false)
+}
+
+// FlushMemDisk saves a new tree to disk and removes all the versions in memory
+func (tree *MutableTree) FlushMemVersionDisk() ([]byte, int64, error) {
+	x, y, err := tree.saveVersion(true)
+	tree.ndb.dbMem = dbm.NewMemDB()
+	tree.memVersions = map[int64]bool{}
+	tree.ndb.memNodes = map[string]*Node{}
+	return x, y, err
+}
+
 // SaveVersion saves a new tree version to disk, based on the current state of
 // the tree. Returns the hash and new version number.
 func (tree *MutableTree) SaveVersion() ([]byte, int64, error) {
+	return tree.saveVersion(true)
+}
+
+func (tree *MutableTree) saveVersion(flushToDisk bool) ([]byte, int64, error) {
 	version := tree.version + 1
+
+	if flushToDisk {
+		tree.ndb.batch = tree.ndb.db.NewBatch()
+	} else {
+		tree.ndb.batch = tree.ndb.dbMem.NewBatch()
+		tree.memVersions[version] = true
+	}
 
 	if tree.versions[version] {
 		//version already exists, throw an error if attempting to overwrite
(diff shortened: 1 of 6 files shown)


jackzampolin:
Thank you for the PR @mattkanwisher!! Getting some folks to review this!

nodedb.go
```diff
@@ -165,11 +185,16 @@ func (ndb *nodeDB) DeleteVersion(version int64, checkLatestVersion bool) {
 // Saves orphaned nodes to disk under a special prefix.
 // version: the new version being saved.
 // orphans: the orphan nodes created since version-1
-func (ndb *nodeDB) SaveOrphans(version int64, orphans map[string]int64) {
+func (ndb *nodeDB) SaveOrphans(version int64, orphans map[string]int64, flushToDisk bool) {
```
HaoyangLiu: What is the purpose of parameter `flushToDisk bool`? It seems useless in `SaveOrphans`

yutianwu:
what if we need to store recent blocks other than every x blocks. should we replay the blocks to get the versions not persisted when restart.

ultraeric:
Concerned with the parametrization of the new functions, in particular `tree.SaveVersion` and `tree.SaveVersionMem` and the complexity that is imposed on the caller. Also could use some better documentation. Two suggestions for changes:

1. Wrap some of the `SaveVersionMem` functionality into SaveVersion and have options be parametrized in the tree constructor. In particular, parametrize the save frequency into the tree constructor. This ideally will reduce migration complexity for anybody that is interested in using these new optimizations. Thus, calling `SaveVersion` will by default use mem-based caching. 
2. Document at the top of `mutable_tree.go` and `nodedb.go` the current orphaning mechanism, as well as this new orphaning mechanism and how to use it.

mutable_tree.go
```diff
@@ -364,11 +366,37 @@ func (tree *MutableTree) GetVersioned(key []byte, version int64) (
...
 
+// SaveVersionMem saves a new tree version to disk, based on the current state of
+// the tree. Returns the hash and new version number.
+func (tree *MutableTree) SaveVersionMem() ([]byte, int64, error) {
+	return tree.saveVersion(false)
+}
+
+// FlushMemDisk saves a new tree to disk and removes all the versions in memory
+func (tree *MutableTree) FlushMemVersionDisk() ([]byte, int64, error) {
+	x, y, err := tree.saveVersion(true)
```
AdityaSripal: could we have more descriptive vars

yutianwu:
on second thought, we can also flush mem version asynchronously. Assume that we have high throughput, though we save state every x blocks, but when it comes to that block, the blocking time may increased. So if we can flush mem version to disk asynchronously, we will have the same performance for each block.

nodedb.go
```diff
@@ -33,9 +33,11 @@ var (
...
 
 type nodeDB struct {
-	mtx   sync.Mutex // Read/write lock.
-	db    dbm.DB     // Persistent node storage.
-	batch dbm.Batch  // Batched writing buffer.
+	mtx      sync.Mutex // Read/write lock.
+	db       dbm.DB     // Persistent node storage.
+	dbMem    dbm.DB     // Memory node storage.
+	batch    dbm.Batch  // Batched writing buffer.
+	memNodes map[string]*Node
```
yutianwu: it seems you do not delete Nodes do not need, the memory will keep growing

jackzampolin:
@yutianwu yeah doing this in a goroutine is def best practice

mutable_tree.go
```diff
@@ -364,11 +366,37 @@ func (tree *MutableTree) GetVersioned(key []byte, version int64) (
...
+	tree.ndb.dbMem = dbm.NewMemDB()
+	tree.memVersions = map[int64]bool{}
```
yutianwu: IMHO, if we do not delete all memVersions and keep recent n version would be better, for it would fail when query recent version of state like querystore, if height is not exactly the latest one, it would fail for the version is just deleted.

marbar3778:
Closing this in favor of #158

AdityaSripal:
@mattkanwisher how often are you flushing versions to disk with loom?

Have you compared the results of various pruning strategies. Are you benchmarking your application with various pruning strategies or are you benchmarking by just testing IAVL in isolation?

Would be very useful to hear your approach to benchmarking this and the results you got. thanks

PR #157, opened 2019-07-12
Remove all cmn from tendermint in IavlStore

marbar3778:
- Remove all instances of cmn in iavl store
- This is the first steps in removing tendermint from a dependancy in this repo & using an adopted error handling practice

ref #46 


mutable_tree.go
```diff
@@ -427,13 +428,13 @@ func (tree *MutableTree) SaveVersion() ([]byte, int64, error) {
 // longer be accessed.
 func (tree *MutableTree) DeleteVersion(version int64) error {
 	if version == 0 {
-		return cmn.NewError("version must be greater than 0")
+		return errors.New("version must be greater than 0")
 	}
 	if version == tree.version {
-		return cmn.NewError("cannot delete latest saved version (%d)", version)
+		return errors.New(fmt.Sprintf("cannot delete latest saved version (%d)", version))
```
melekes: `fmt.Errorf` or `errors.Newf`

proof_test.go
```diff
@@ -17,7 +17,7 @@ func TestTreeGetWithProof(t *testing.T) {
 	require := require.New(t)
 	for _, ikey := range []byte{0x11, 0x32, 0x50, 0x72, 0x99} {
 		key := []byte{ikey}
-		tree.Set(key, []byte(random.Str(8)))
+		tree.Set(key, []byte(RandStr(8)))
```
melekes: what's wrong with using cmn?

proof.go
```diff
@@ -88,9 +88,9 @@ func (pin proofInnerNode) Hash(childHash []byte) []byte {
 //----------------------------------------
 
 type proofLeafNode struct {
-	Key       cmn.HexBytes `json:"key"`
-	ValueHash cmn.HexBytes `json:"value"`
-	Version   int64        `json:"version"`
+	Key       []byte `json:"key"`
```
melekes: are you sure this won't have any effects?

proof_test.go
marbar3778: I don't think we should be importing tendermint into this repo. Its the first step of 1) switching to using an adopted error handling practice in the tendermint org 2) removing tendermint as a dependency in this repo.

proof.go
marbar3778: will investigate, looked into go-amino, but couldn't find a reason for why someone would need to use cmn.hexbytes. Will dig through the SDK next

proof_test.go
marbar3778: back to using cmn, but from github.com/tendermint/tm-cmn

melekes: > I don't think we should be importing tendermint into this repo

Completely agree 👍

proof.go
melekes:
```
// The main purpose of HexBytes is to enable HEX-encoding for json/encoding.
type HexBytes []byte
```

marbar3778: its back, just from tm-cmn

node.go
```diff
@@ -8,9 +8,9 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/pkg/errors"
```
melekes: Would be nice to have a newline between external and internal deps

proof_iavl_value.go
```diff
@@ -3,8 +3,8 @@ package iavl
 import (
 	"fmt"
 
+	"github.com/pkg/errors"
```
melekes: here too

proof_range.go
```diff
@@ -6,8 +6,8 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/pkg/errors"
```
melekes: here too

tree_test.go
```diff
@@ -4,27 +4,25 @@ import (
...
 )
 
 var testLevelDB bool
 var testFuzzIterations int
-var random *cmn.Rand
 
 func init() {
-	random = cmn.NewRand()
-	random.Seed(0) // for determinism
+	mathrand.Seed(0) // for determinism
```
melekes: revert?

tree_test.go
```diff
@@ -1319,7 +1317,7 @@ func BenchmarkTreeLoadAndDelete(b *testing.B) {
 			// If we can load quickly into a data-structure that allows for
 			// efficient deletes, we are golden.
 			for v := 0; v < numVersions/10; v++ {
-				version := (random.Int() % numVersions) + 1
+				version := (rand.Int() % numVersions) + 1
```
melekes: use cmn.RandInt() ?

PR #160, opened 2019-07-30
Move used common files to repo

marbar3778:
- Reason: tm-cmn has been decided to become tm-db, to solely hold the db functionality.
- also did some minor linting cleanup

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -7,15 +7,12 @@ linters:
     - gocyclo
     - golint
     - gosimple
-    - goimports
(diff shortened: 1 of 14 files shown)


.golangci.yml
```diff
@@ -7,15 +7,12 @@ linters:
     - gocyclo
     - golint
     - gosimple
-    - goimports
```
melekes: reason for removing?

marbar3778: its now enabled

PR #163, opened 2019-08-06
Iavl rollback: not ready for merge

ethanfrey:
This is an experimental script to undo the last few blocks of an sdk store.
Useful when you have the wrong version of the software and get
`ERROR: Error during handshake: Error on replay: Wrong Block.Header.AppHash.` on startup.

Usage:

1. Compile script
2. Look for the abci app database (it should be `data/application.db`, you need the full path)
3. Stop your blockchain daemon
4. Check we can read the application db
5. Roll it back
6. Check it was updated
7. Restart application daemon.

This was inspired by issues with regen network's upgrade... so some info my be regen specific.

## Details

Compile Script:

```console
git clone https://github.com/confio/iavl
cd iavl
git checkout iavl-rollback
GO111MODULES=on go install ./cmd/iaviewer

# this should show a just built binary
ls -l `which iaviewer`
```

Check state: `iaviewer stores <PATH TO application.db>`
You should see the last height and all stores

Roll back two blocks: `iaviewer rollback <PATH TO application.db>`

Check state again: `iaviewer stores <PATH TO application.db>`
You should see the latest height is 2 less, but all stores still listed

Once this is valid, restart `xrnd` *v0.4.0* and wait for it to hit the upgrade block (953626).
Then you can stop it, upgrade to *v0.4.1* and continue with no more consensus issues

### DIFF
--- a/cmd/iaviewer/main.go
+++ b/cmd/iaviewer/main.go
@@ -56,19 +56,6 @@ func main() {
 	}
 	fmt.Printf("Latest: %d\n", latest)
 
-	if args[0] == "rollback" {
-		// TODO: rollback all substores
-		err = SetLatestVersion(db, latest-2)
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "Cannot show stores: %s\n", err)
-			os.Exit(1)
-		}
-	}
-
-	if version == 0 {
-		version = int(latest)
-	}
-
 	if args[0] == "stores" {
 		stores, err := GetSubStores(db, latest)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/iaviewer/main.go
```diff
@@ -89,16 +148,80 @@ func PrintDbStats(db dbm.DB) {
...
+
+	stateKey := []byte(fmt.Sprintf("s/%d", ver))
+	stateBz := db.Get(stateKey)
+	err := cdc.UnmarshalBinaryLengthPrefixed(stateBz, &cInfo)
 	if err != nil {
 		return nil, err
 	}
-	tree := iavl.NewMutableTree(db, DefaultCacheSize)
+
+	var res []string
```
golangcibot: Consider preallocating `res` (from `prealloc`)

AdityaSripal:
Very cool, I opened an issue for this on the tendermint repo a while back: https://github.com/tendermint/tendermint/issues/3845

Glad to see other people have been thinking about adding the same functionality. I have two questions:

1. My understanding of this issue is that the IAVL already has the `Rollback` function. We just need to use it in a tm command and also roll back the rest of tendermint state that isn't stored in IAVL. Is the `Rollback` function in `mutable_tree.go` not sufficient to rollback AppState by one block?
2. Why do we need to roll back more than one block in your example above. If the AppHash is unexpected at height `H`, why would we have to rollback beyond height `H-1`?

erikgrinaker:
I agree with @AdityaSripal here - we may already have the basic functionality in IAVL (I'd have to check), and this should be handled higher up in the stack (e.g. Cosmos SDK or Tendermint) in order to handle application-specific rollback logic. Since I believe this script is coupled to the Cosmos SDK (by reading the `StoreInfo` stuff) it's not really appropriate to merge this as is, so I'll close this for now.

ethanfrey:
Yeah, I never wanted to merge, but was here to share some scripts I worked on, that were necessary for some other postmortems.

The fact we have no cli tooling for this pops up many times when there are errors on updates, consensus somehow lost when syncing, etc. It would be great if someone priortized a real solution to this and linked the issue/PR. The PR was more a demo of a one-off tool I did with the hope it would move forward to better tooling.

erikgrinaker:
Totally, the issue to follow would be https://github.com/tendermint/tendermint/issues/3845 I suppose.

PR #164, opened 2019-08-07
Add function to find latest version available.

atchapcyp:
ref : https://github.com/loomnetwork/loomchain/pull/1362


atchapcyp:
Oh, I accidentally push to the wrong place.

PR #158, opened 2019-07-23
Aditya/pruning

AdityaSripal:

PR #109, opened 2018-09-05
WIP - New Benchmarks

jlandrews:

PR #165, opened 2019-08-08
Versioned Benchmarks

AdityaSripal:
- Creates a more detailed version struct
- Prefaces all benchmark results with version information.

Should be merged before pruning PR: #158 

Will be useful to monitor performance regressions/improvements clearly, keep older benchmark results even in new version codebase for comparisons.


AdityaSripal:
TODO:

- [ ] add benchmark results from other machines
- [ ] organize benchmark results in subfolders labelled by version-commit

marbar3778:
can #109 be closed in favor of this one?

AdityaSripal:
#109 adds new benchmark test, while this one is to just add versioning info so they're a bit orthogonal. Though I think any future benchmark results that get merged should have version info included

README.md
```diff
@@ -1,6 +1,6 @@
 ## IAVL+ Tree
 
-**Note: Requires Go 1.8+**
+**Note: Requires Go 1.11+**
```
marbar3778: 👍

PR #166, opened 2019-08-09
Add a couple linters

### LINKED ISSUE #57: Some errors are not checked (errcheck linter)
Some errors are not handled. If we expect that no errors can occur at these places, we can panic to clarify that something is seriously wrong (if an error occurs). If not, at least add a comment. Better handle all errors:
```
node.go:197:14:warning: error return value not checked (hasher.Write(buf.Bytes())) (errcheck)
node.go:216:14:warning: error return value not checked (hasher.Write(buf.Bytes())) (errcheck)
orphaning_tree.go:49:25:warning: error return value not checked (tree.ndb.SaveEmptyRoot(version)) (errcheck)
orphaning_tree.go:55:20:warning: error return value not checked (tree.ndb.SaveRoot(tree.root, version)) (errcheck)
proof.go:70:14:warning: error return value not checked (hasher.Write(buf.Bytes())) (errcheck)
proof.go:101:14:warning: error return value not checked (hasher.Write(buf.Bytes())) (errcheck)
```

marbar3778:
- govet, gosimple, errcheck

- no need to check errors in bench and testing files
- in non-testing files i just added panic, since we are not expecting it to fail.

fixes #57


node.go
```diff
@@ -205,7 +205,10 @@ func (node *Node) _hash() []byte {
 	if err := node.writeHashBytes(buf); err != nil {
 		panic(err)
 	}
-	h.Write(buf.Bytes())
+	_, err := h.Write(buf.Bytes())
+	if err != nil {
```
marbar3778: we weren't checking before so there is a uber low chance of erroring, so panic is fine here

mutable_tree.go
```diff
@@ -331,7 +331,10 @@ func (tree *MutableTree) LoadVersionForOverwriting(targetVersion int64) (int64,
 	if err != nil {
 		return latestVersion, err
 	}
-	tree.deleteVersionsFrom(targetVersion + 1)
+	err = tree.deleteVersionsFrom(targetVersion + 1)
+	if err != nil {
+		panic(err)
```
AdityaSripal: we should not be panicing on this error

marbar3778: I went for panics because this was never checked before, so the assumption was always it wouldn't error. what would you recommend?

AdityaSripal: I think ignoring the error is the correct behavior in this case. It's possible that a previous version got manually deleted before this function got called. In fact, I don't think it is correct for `deleteVersionsFrom` to return an error in the first place if there happens to be a version in its range that already got deleted.

marbar3778: Maybe not panic, but log the error would be useful or we just remove the returning error if we see its not used elsewhere.

alexanderbez: Just my 2 cents, it's generally advisable that libraries should _not_ panic is almost any circumstance, but rather turn (typed) errors. Let the upstream caller decide what to do with an error and don't assume.

PR #168, opened 2019-08-28
pre-allocate memory for collecting orphan nodes

rickyyangz:
estimate and pre-allocate the space for the orphaned nodes. It really matters when the tree become higher and higher and we have frequent set/remove operations.

benchmark
```
before:
BenchmarkMutableTree_Set-12    	  200000	      8845 ns/op	    6359 B/op	      65 allocs/op

after:
BenchmarkMutableTree_Set-12    	  300000	      5339 ns/op	    4604 B/op	      25 allocs/op
```

### DIFF
--- a/mutable_tree.go
+++ b/mutable_tree.go
@@ -78,6 +78,12 @@ func (tree *MutableTree) String() string {
 	return tree.ndb.String()
 }
 
+// Set/Remove will orphan at most tree.Height nodes,
+// balancing the tree after a Set/Remove will orphan at most 3 nodes.
+func (tree *MutableTree) prepareOrphansSlice() []*Node {
+	return make([]*Node, 0, tree.Height()+3)
+}
+
 // Set sets a key in the working tree. Nil values are not supported.
 func (tree *MutableTree) Set(key, value []byte) bool {
 	orphaned, updated := tree.set(key, value)
@@ -85,23 +91,22 @@ func (tree *MutableTree) Set(key, value []byte) bool {
 	return updated
 }
 
-func (tree *MutableTree) set(key []byte, value []byte) ([]*Node, bool) {
+func (tree *MutableTree) set(key []byte, value []byte) (orphans []*Node, updated bool) {
 	if value == nil {
 		panic(fmt.Sprintf("Attempt to store nil value at key '%s'", key))
 	}
 
-	updated := false
 	if tree.ImmutableTree.root == nil {
 		tree.ImmutableTree.root = NewNode(key, value, tree.version+1)
 		return nil, updated
 	}
 
-	orphaned := make([]*Node, 0, tree.Height()*2)
-	tree.ImmutableTree.root, updated = tree.recursiveSet(tree.ImmutableTree.root, key, value, &orphaned)
-	return orphaned, updated
+	orphans = tree.prepareOrphansSlice()
+	tree.ImmutableTree.root, updated = tree.recursiveSet(tree.ImmutableTree.root, key, value, &orphans)
+	return orphans, updated
 }
 
-func (tree *MutableTree) recursiveSet(node *Node, key []byte, value []byte, orphaned *[]*Node) (
+func (tree *MutableTree) recursiveSet(node *Node, key []byte, value []byte, orphans *[]*Node) (
 	newSelf *Node, updated bool,
 ) {
 	version := tree.version + 1
@@ -127,18 +132,18 @@ func (tree *MutableTree) recursiveSet(node *Node, key []byte, value []byte, orph
 				version:   version,
 			}, false
 		default:
-			*orphaned = append(*orphaned, node)
+			*orphans = append(*orphans, node)
 			return NewNode(key, value, version), true
 		}
 	} else {
-		*orphaned = append(*orphaned, node)
+		*orphans = append(*orphans, node)
 		node = node.clone(version)
 
 		if bytes.Compare(key, node.key) < 0 {
-			node.leftNode, updated = tree.recursiveSet(node.getLeftNode(tree.ImmutableTree), key, value, orphaned)
+			node.leftNode, updated = tree.recursiveSet(node.getLeftNode(tree.ImmutableTree), key, value, orphans)
 			node.leftHash = nil // leftHash is yet unknown
 		} else {
-			node.rightNode, updated = tree.recursiveSet(node.getRightNode(tree.ImmutableTree), key, value, orphaned)
+			node.rightNode, updated = tree.recursiveSet(node.getRightNode(tree.ImmutableTree), key, value, orphans)
 			node.rightHash = nil // rightHash is yet unknown
 		}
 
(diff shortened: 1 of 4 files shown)


mutable_tree.go
```diff
@@ -85,21 +85,24 @@ func (tree *MutableTree) Set(key, value []byte) bool {
...
+
+	updated := false
 	if tree.ImmutableTree.root == nil {
 		tree.ImmutableTree.root = NewNode(key, value, tree.version+1)
-		return nil, false
+		return nil, updated
 	}
-	tree.ImmutableTree.root, updated, orphaned = tree.recursiveSet(tree.ImmutableTree.root, key, value)
 
+	orphaned := make([]*Node, 0, tree.Height() * 2)
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	orphaned := make([]*Node, 0, tree.Height()*2)
```

mutable_tree.go
```diff
@@ -85,21 +85,24 @@ func (tree *MutableTree) Set(key, value []byte) bool {
...
+
+	updated := false
 	if tree.ImmutableTree.root == nil {
 		tree.ImmutableTree.root = NewNode(key, value, tree.version+1)
-		return nil, false
+		return nil, updated
 	}
-	tree.ImmutableTree.root, updated, orphaned = tree.recursiveSet(tree.ImmutableTree.root, key, value)
 
+	orphaned := make([]*Node, 0, tree.Height()*2)
```
melekes: Why `tree.Height() * 2`? Shouldn't it be `tree.Size()` or something?

mutable_tree.go
```diff
@@ -160,11 +161,12 @@ func (tree *MutableTree) Remove(key []byte) ([]byte, bool) {
 
 // remove tries to remove a key from the tree and if removed, returns its
 // value, nodes orphaned and 'true'.
-func (tree *MutableTree) remove(key []byte) (value []byte, orphans []*Node, removed bool) {
+func (tree *MutableTree) remove(key []byte) ([]byte, []*Node, bool) {
```
AdityaSripal: Why were named return vals removed here?

mutable_tree.go
AdityaSripal: `recursiveSet` will orphan every node that it recurses upon. Thus, this will orphan `tree.Height` nodes.

I'm guessing the `* 2` is from the balance call at the end? However, if you look at balance implementation; it orphans at most 3 nodes.

Thus, I believe this line should be `tree.Height + 3`. Someone should definitely check my math on this.

Regardless, definitely needs a docstring here explaining how we get this number.

mutable_tree.go
```diff
@@ -229,8 +232,8 @@ func (tree *MutableTree) recursiveRemove(node *Node, key []byte) ([]byte, *Node,
 	}
 	newNode.calcHeightAndSize(tree.ImmutableTree)
 	newNode, balanceOrphaned := tree.balance(newNode)
```
AdityaSripal: Why can't we update `balance` function to take in an orphan list as an argument and append there?

mutable_tree_test.go
```diff
@@ -0,0 +1,25 @@
+package iavl
+
+import (
+	db "github.com/tendermint/tm-db"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

mutable_tree.go
```diff
@@ -184,43 +186,44 @@ func (tree *MutableTree) remove(key []byte) (value []byte, orphans []*Node, remo
 // - new leftmost leaf key for tree after successfully removing 'key' if changed.
 // - the removed value
 // - the orphaned nodes.
-func (tree *MutableTree) recursiveRemove(node *Node, key []byte) ([]byte, *Node, []byte, []byte, []*Node) {
+func (tree *MutableTree) recursiveRemove(node *Node, key []byte, orphaned *[]*Node) ([]byte, *Node, []byte, []byte) {
```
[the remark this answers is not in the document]
rickyyangz: ok, added

rickyyangz:
> Same here. Would love to see some benchmarks

added :)

mutable_tree.go
rickyyangz: good point, changed.

mutable_tree.go
rickyyangz: right, 
`recursiveSet` will orphan at most `tree.Height()` nodes, `balance` will produce at most 3 nodes. so `tree.Height() + 3` is enough.

mutable_tree.go
```diff
@@ -85,21 +85,23 @@ func (tree *MutableTree) Set(key, value []byte) bool {
...
+
 	if tree.ImmutableTree.root == nil {
 		tree.ImmutableTree.root = NewNode(key, value, tree.version+1)
-		return nil, false
+		return nil, updated
 	}
-	tree.ImmutableTree.root, updated, orphaned = tree.recursiveSet(tree.ImmutableTree.root, key, value)
 
-	return orphaned, updated
+	orphans = make([]*Node, 0, tree.Height()+3)
```
AdityaSripal: Again please add docstring explaining how we get this number.

`// Set will orphan at most tree.Height nodes, balancing the tree after a Set will orphan at most 3 nodes.`

mutable_tree.go
```diff
@@ -160,11 +159,12 @@ func (tree *MutableTree) Remove(key []byte) ([]byte, bool) {
 
 // remove tries to remove a key from the tree and if removed, returns its
 // value, nodes orphaned and 'true'.
-func (tree *MutableTree) remove(key []byte) (value []byte, orphans []*Node, removed bool) {
+func (tree *MutableTree) remove(key []byte) (value []byte, orphaned []*Node, removed bool) {
 	if tree.root == nil {
 		return nil, nil, false
 	}
-	newRootHash, newRoot, _, value, orphaned := tree.recursiveRemove(tree.root, key)
+	orphaned = make([]*Node, 0, tree.Height()+3)
```
AdityaSripal: docstring here as well

mutable_tree.go
AdityaSripal: I think it may make sense to make this a helper function that returns the orphan slice for you. Makes the code slightly more DRY making it easier to update if the calculation ever changes, but also can have documentation on that helper function.

mutable_tree.go
AdityaSripal: ^As mentioned elsewhere, this should be abstracted into tiny helper function with documentation imo.

mutable_tree_test.go
```diff
@@ -0,0 +1,26 @@
+package iavl
+
+import (
+	"runtime"
+	"testing"
+
+	db "github.com/tendermint/tm-db"
+)
+
+func BenchmarkMutableTree_Set(b *testing.B) {
```
AdityaSripal: Pipe results to file and save under `benchmarks/result/168/`

mutable_tree.go
rickyyangz: sure

mutable_tree.go
```diff
@@ -78,28 +78,36 @@ func (tree *MutableTree) String() string {
 	return tree.ndb.String()
 }
 
+// Set/Remove will orphan at most tree.Height nodes,
+// balancing the tree after a Set/Remove will orphan at most 3 nodes.
+func (tree *MutableTree) prepareOrphansSlice() []*Node {
+	return make([]*Node, 0, tree.Height() + 3)
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	return make([]*Node, 0, tree.Height()+3)
```

mutable_tree_test.go
rickyyangz: ok. I added. but what's the meaning of 168?

Issue #183, opened 2019-11-06
Feature/ gRPC Server


charlescrain:
Hey everyone!

I wanted to start a discussion/make a feature request for adding a gRPC interface to IAVL. 

How difficult would this be to implement?

I'm currently working on a Haskell implementation of the [Cosmos SDK](https://github.com/f-o-a-m/hs-abci) and it would be great to be able to have the SDK use iavl as its key store via the gRPC.

Cheers!
Charles

alexanderbez:
Hey @charlescrain. What does that look like? Is that just an API bridge/layer on top of the existing IAVL API?

blinky3713:
I can volunteer what I know about it. The idea came from working with [google trillian](https://github.com/google/trillian) and writing a [client library](https://github.com/f-o-a-m/hs-trillian) in haskell.

Basically you would write the protobuf descriptions for the api request/response types and define the api itself as a `service` like [this](https://github.com/google/trillian/blob/master/trillian_log_api.proto#L57-L65).  Then you would have to implement the server itself on your end, either in this repo or it could exist as a standalone package.

I think it has advantages in that it allows abci applications in other languages targeting to use the same db as the official cosmos-sdk, in our case allowing a haskell app to talk to a `go` app. Also from a devops perspective it could make things pretty nice to have a separate DB service.

alexanderbez:
I see. Pretty neat. And this would, of course, be an alternative to writing the IAVL library in Haskell? I think we can do this but it would preferable to see this come as a community contribution and have the server as a standalone package.

blinky3713:
> And this would, of course, be an alternative to writing the IAVL library in Haskell
Yes, and really any other language, even if only a stopgap that does emerge in that language

blinky3713:
I would say that we would be able to commit to attempting a server implementation in go, but that when we get to a point we would like help from ya'll in writing the protobuf files to describe the service. What do you think?

alexanderbez:
Sounds great! Let's get started on it. I can start a PR with a single proto type and you can see if this is the correct approach or not before proceeding with the remaining types?

blinky3713:
Ok we're going to take a stab at it. We are not the best `go` developers, but we probably be fine to get something working

charlescrain:
@alexanderbez We've started a repo for the grp server and made an [issue](https://github.com/f-o-a-m/iavl-grpc/issues/1) for getting the protobuf files for all the types and services in one place. 

[...]

PR #184, opened 2019-11-10
gRPC API

alexanderbez:
gRPC API w/ JSON HTTP proxy (via grpc-gateway) implementation for the IAVL library. 

**TODO**:

- [x] Define gRPC service
- [x] Define all message types
- [x] Implement service implementation
- [x] Implement JSON HTTP proxy server
- [x] Tests
- [x] Fix build/CI

closes: #183

/cc @blinky3713 @charlescrain


blinky3713:
Just a comment after checking on on the status here, will there also be a `Rollback` endpoint? Is there a different API for the mutable tree vs immutable ?

server/server.go
```diff
@@ -0,0 +1,198 @@
...
+// iavlServer implements the gRPC IAVLServiceServer interface. It provides a gRPC
+// API over an IAVL tree.
+type IAVLServer struct {
+	tree *iavl.MutableTree
+}
+
+func New(db dbm.DB, cacheSize, version int64) (*IAVLServer, error) {
+	tree := iavl.NewMutableTree(db, int(cacheSize))
+
+	// TODO: Consider storing and load latest height (per version save) when the
```
golangcibot: server/server.go:25: Line contains TODO/BUG/FIXME: "TODO: Consider storing and load latest h..." (from `godox`)

util.go
```diff
@@ -105,6 +107,52 @@ func sortByteSlices(src [][]byte) [][]byte {
...
+			Height:  int8(n.Height),
+			Size:    n.Size,
+			Version: n.Version,
+			Left:    n.Left,
+			Right:   n.Right,
+		}
+	}
+
+	innerNodes := make([]PathToLeaf, len(proofPb.InnerNodes))
+	for i, pl := range proofPb.InnerNodes {
```
golangcibot: ranges should only be cuddled with assignments used in the iteration (from `wsl`)

server/server.go
```diff
@@ -0,0 +1,198 @@
...
+// iavlServer implements the gRPC IAVLServiceServer interface. It provides a gRPC
+// API over an IAVL tree.
+type IAVLServer struct {
+	tree *iavl.MutableTree
+}
+
+func New(db dbm.DB, cacheSize, version int64) (*IAVLServer, error) {
+	tree := iavl.NewMutableTree(db, int(cacheSize))
+
+	// TODO: Consider storing (per version save) and load latest height when the
```
golangcibot: server/server.go:25: Line contains TODO/BUG/FIXME: "TODO: Consider storing (per version save..." (from `godox`)

alexanderbez:
@blinky3713 I've implemented what I believe to be a majority of the implementation. To answer, your question, the API currently works on a mutable tree. If we'd like to differentiate and support both tree types, that can be done later down the line after this is reviewed and merged.

However, there is still some work to be done which I'd like your help on as I can no longer dedicate too much time to this atm. The work needed to be done:

* Update makefile to install proto dependencies (if needed)
* Implement `Server#GetWithProof` with unit tests
* Implement unit tests for `Server#{Rollback|VerifyAbsence|VerifyItem|Verify}`
* Docs (can be simple usage instructions in the main README).

This can be done as a separate PR (forked off of this one) with the PR merging directly into this PR (which I'll keep open).

Also, another outstanding item. Do we want to support the automatic loading of the latest version (e.g. if 0 is provided)?

server/server.go
charlescrain: @alexanderbez Quick question on your `TODO` here. I like the idea of defaulting to the latest height when `version` is 0. I don't quite follow the `Consider storing (per version save)` part, could you elaborate some more?

charlescrain: Actually it looks like that is the default behavior anyway:
https://github.com/tendermint/iavl/blob/master/mutable_tree.go#L243-L246

alexanderbez: > I took this to mean somehow looking up the latest version from the DB.

Yes, exactly. This is correct. This is precisely what the SDK does: https://github.com/cosmos/cosmos-sdk/blob/3c2ace95100d47f7fc4f68241fa91c2158ee368a/store/rootmulti/store.go#L596. You'll notice we save the version (on commit) under a unique key. The TODO suggests we do the same here.

The reason it is a TODO is because idk if we _should_ do this or if we should keep the API and "clean" as possible. The latter would mean the client would need to know the latest height.

charlescrain: @alexanderbez Ahh okay, I get you now. Since the sdk takes responsibility for saving the version I'd vote that we leave it out of the gRPC server as it then better reflects the IAVL native API.

Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM golang:1.12 as build
```
marbar3778:
```suggestion
FROM golang as build
```

cmd/iavlserver/main.go
```diff
@@ -0,0 +1,184 @@
...
+// server.
+func startRPCGateway() error {
+	jsonPb := &gateway.JSONPb{
+		EmitDefaults: true,
+		Indent:       "  ",
+		OrigName:     true,
+	}
+	gatewayMux := runtime.NewServeMux(
+		runtime.WithMarshalerOption(runtime.MIMEWildcard, jsonPb),
+		// This is necessary to get error details properly marshalled in unary requests.
```
golangcibot: `marshalled` is a misspelling of `marshaled` (from `misspell`)
```suggestion

```

server/server_test.go
```diff
@@ -0,0 +1,662 @@
...
+	versionRes, err := suite.server.Version(context.TODO(), nil)
+	suite.NoError(err)
+
+	for i := 0; i < n; i++ {
+		req := &pb.SetRequest{
+			Key:   []byte(fmt.Sprintf("key-%d", i)),
+			Value: []byte(fmt.Sprintf("value-%d", i)),
+		}
+
+		_, err := suite.server.Set(context.TODO(), req)
```
golangcibot: shadow: declaration of "err" shadows declaration at line 33 (from `govet`)

Makefile
```diff
@@ -71,4 +68,19 @@ exploremem:
...
-.PHONY: lint test tools install delve exploremem explorecpu profile fullbench bench
+protogen:
+	protoc -I/usr/local/include -I. \
+	-I$(GOPATH)/src \
+	-I$(GOPATH)/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis \
+	--go_out=plugins=grpc:. \
+	--grpc-gateway_out=logtostderr=true:. \
+	proto/iavl_api.proto
+
+protolint:
```
marbar3778: we should use buf or prototool here and also add https://buf.build/docs/tour-5 in ci

marbar3778:
I believe this will need to extended after the pruning PR is merged #158

blinky3713:
thanks for merging from `f-o-a-m/ival` @marbar3778 , do I understand that yall's team will now take care of the remaining issues?

blinky3713:
One issue that has already come up for us:

When querying via `Get` for a value associated to a key `k`, when the key is not present in the db the value returned is `""` because protobuf lacks support for an optional type. This makes it impossible to tell whether `k` is actually missing or if the associated value is actually `""`. 

We thought of one way around this which would be that if the value `""` is returned we could try to use the `Has` method to double check,  but it turns out that requires a `Version` to query against? If that's the case, how can we query against an unsaved version?

Ya'll might have an answer, let me know if this doesn't make sense.

marbar3778:
this PR needs to be extended for pruning.

alexanderbez:
> One issue that has already come up for us:
> 
> When querying via `Get` for a value associated to a key `k`, when the key is not present in the db the value returned is `""` because protobuf lacks support for an optional type. This makes it impossible to tell whether `k` is actually missing or if the associated value is actually `""`.
> 
> We thought of one way around this which would be that if the value `""` is returned we could try to use the `Has` method to double check, but it turns out that requires a `Version` to query against? If that's the case, how can we query against an unsaved version?
> 
> Ya'll might have an answer, let me know if this doesn't make sense.

Interesting. Sorry, I haven't looked at this in a while, but `GetResponse.Value` is a `bytes` type. So if the key doesn't exist, shouldn't the result be `null` instead of `""`? Or does gRPC convert this to an encoded string?

tnachen:
@marbar3778 @alexanderbez I'll look into this

blinky3713:
@alexanderbez we think that the problem is coming from the protobuf (de)serialization, but not 100% sure.

Is the index of a missing key well defined? This would be another problem because I'm assuming that's also `null`, but here again because of protobuf codecs it comes out as `0`. 

Am I missing something, because it seems like a pretty annoying feature of protobufs :/

alexanderbez:
Mhhh, actually, is this in regards to JSON over gRPC? If so, I believe proto's behavior will automatically covert the byte slice to a base64 encoded string. If the value was `null`, then this will yield `""`. I think we may need to update `IAVLServer#Get` to instead return an error when the value doesn't exist so the client handle tell the difference.

blinky3713:
according to https://developers.google.com/maps-booking/reference/grpc-api/status_codes this would be appropriate
```
5 | NOT_FOUND | Some requested entity was not found.
```

Makefile
tnachen: Seems like tendermint is using protoc too, how about switching both when we learn how to use it?

marbar3778: I can switch tendermint to buf. I already have a branch with it but it contains proto breakage

tnachen:
Nice, can you share the branch?

marbar3778:
https://github.com/tendermint/tendermint/tree/proto-breakage

tnachen:
@blinky3713 I've updated the PR to return the grpc status code, do you mind verifying it on your end?

server/server.go
```diff
@@ -0,0 +1,230 @@
+package server
+
+import (
+	"context"
+	"errors"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	"errors"

```

marbar3778:
proto-breakage will fail until this is merged into master

tnachen:
@marbar3778 I think this is ready to merge, any more reviews?

marbar3778:
Havent read much into it since the updates but does it support the new pruning logic?

can you also make sure that all the check boxes in the body of the PR are checked if they are completed?

tnachen:
@marbar3778 Just checked the boxes and updated the flags to include pruning options

erikgrinaker:
I understand the use-case for accessing IAVL in other languages without a writing a wrapper. However, going from an embedded database library to a networked database server is a pretty fundamental change - are we sure this is something we want to do?

In particular, I am thinking of concurrency control. IAVL is not concurrency safe, but gRPC is fundamentally concurrent. There is also an expectation that a network server can handle multiple clients, but IAVL is not built for this and will likely not perform well.

If we want do to this, we should wrap the database in a `RWMutex` as a first step, limit simultaneous connections to e.g. 2, and also document thoroughly that the gRPC service is meant for use by a single client. To achieve good multi-client performance we will need to use a database backend with proper concurrency control built-in (e.g. BadgerDB), and should also support something approximating ACID transactions to avoid conflicts between clients. This is related to #241.

Thoughts?

alexanderbez:
I have to agree with @erikgrinaker here. At the very least, we need to add smart mutex locking and think carefully around multi-client use and ensure we're ACID-compliant -- at least multi-client reads.

blinky3713:
I'm not sure I understand why the full enterprise solution of a distributed database is required *before* the grpc layer is implemented. The semantics can remain undefined in the event that you have multiple concurrent writes -- just leave a big warning or mark the feature as experimental.

Unless I'm severely mistaking, the original intended use case of the grpc api is unaffected by concurrency, which was to be able to use the cosmos iavl storage from a non-golang abci application. When we have one tendermint client + abci application, the transactions are executed sequentially during block execution, and the transactions themselves are sequential. There are no race conditions for writes. As for reads, the applications is responsible for keeping the appropriate indices for writes -- e.g. the index used for `query` or `checkTx` commands points to committed storage, hence there are no race conditions there either.

Would it not be reasonable to just commit the work that the tendermint team and @charlescrain already did here and then open up separate issues to discuss the possible dirrections to allow concurrency?

erikgrinaker:
> I'm not sure I understand why the full enterprise solution of a distributed database is required _before_ the grpc layer is implemented. The semantics can remain undefined in the event that you have multiple concurrent writes -- just leave a big warning or mark the feature as experimental.

At the very least, we should wrap the tree in an RWMutex.

> Unless I'm severely mistaking, the original intended use case of the grpc api is unaffected by concurrency, which was to be able to use the cosmos iavl storage from a non-golang abci application.

Yes, this is _your_  use-case (and I totally get that use-case). However, others are quite likely to start using this for other use-cases that it's not well-suited for, and then get sad when things don't work out very well.

> Would it not be reasonable to just commit the work that the tendermint team and @charlescrain already did here and then open up separate issues to discuss the possible dirrections to allow concurrency?

Personally I'm a bit worried about the additional maintenance overhead, since IAVL is already a but under-maintained. Maybe we could split this out as a separate package that could be maintained on a best-effort basis?

@aaronc Since you will be taking over IAVL maintainership shortly, perhaps you'd like to weigh in here.

aaronc:
I think the limitations that @erikgrinaker proposes are reasonable.

Regarding maintenance overhead, this is a relatively small layer on top of the go API so it would likely be easier to maintain within this source tree because bugs would ideally be caught by tests. I do think we should mark this as an alpha API so that we don’t make any strong compatibility guarantees.

https://github.com/google/trillian, a similar project, also has a gRPC interface so this is not unprecedented. 

I would like to see some tests which I think should be possible in process by just spinning up a server at a random port in a separate thread and sending requests to it.

We don’t currently have the internal bandwidth to finish this work quite yet. But I know the Kepler team needs this and if they have the bandwidth to wrap up this work with adequate test coverage I would support merging it.

I do want to add the disclaimer that big changes may be coming to IAVL, hopefully big improvements. But hopefully this doesn't affect the user facing API too much.

blinky3713:
@aaronc I have started the work here https://github.com/f-o-a-m/iavl/tree/martin/grpc . I was able to bring it up to speed with master and all tests are passing. I will implement the dumbest concurrency possible:

1. a global write mutex
2. single sequential writes
3. unlimited concurrent reads

There are already tests (`server/server_test.go`). When I submit the PR you can explain what changes to make to them or what needs to be added.

erikgrinaker:
Sounds good, thanks!

blinky3713:
I would propose closing this in favor of #296

erikgrinaker:
Closing this in favor of #296, discussion is continued there.

PR #194, opened 2020-01-08
dep: update tm-db to latest release

marbar3778:


nodedb.go
```diff
@@ -116,17 +116,22 @@ func (ndb *nodeDB) SaveNode(node *Node) {
...
 
 // Has checks if a hash exists in the database.
-func (ndb *nodeDB) Has(hash []byte) bool {
+func (ndb *nodeDB) Has(hash []byte) (bool, error) {
 	key := ndb.nodeKey(hash)
 
 	if ldb, ok := ndb.db.(*dbm.GoLevelDB); ok {
 		exists, err := ldb.DB().Has(key, nil)
 		if err != nil {
-			panic("Got error from leveldb: " + err.Error())
```
marbar3778: do we want to keep the panic here? or let the end-user handle it? do you have an opinion @alexanderbez

alexanderbez: It is of my opinion a DB library should never panic -- allow the upstream user to decide on how to handle the error.

nodedb.go
```diff
@@ -73,8 +73,8 @@ func (ndb *nodeDB) GetNode(hash []byte) *Node {
 	}
 
 	// Doesn't exist, load.
-	buf := ndb.db.Get(ndb.nodeKey(hash))
-	if buf == nil {
+	buf, err := ndb.db.Get(ndb.nodeKey(hash))
+	if err != nil {
```
melekes:
```suggestion
	if buf == nil {
```

nodedb.go
```diff
@@ -73,8 +73,8 @@ func (ndb *nodeDB) GetNode(hash []byte) *Node {
...
 		panic(fmt.Sprintf("Value missing for hash %x corresponding to nodeKey %s", hash, ndb.nodeKey(hash)))
 	}
```
melekes: I believe a new condition is needed

```
	if err != nil {
		panic(fmt.Sprintf("can't get node %X: %v", hash, err))
	}
```

nodedb.go
melekes: Strongly endorse ^

nodedb.go
```diff
@@ -116,17 +119,22 @@ func (ndb *nodeDB) SaveNode(node *Node) {
...
 	key := ndb.nodeKey(hash)
 
 	if ldb, ok := ndb.db.(*dbm.GoLevelDB); ok {
 		exists, err := ldb.DB().Has(key, nil)
 		if err != nil {
-			panic("Got error from leveldb: " + err.Error())
+			return false, err
 		}
-		return exists
+		return exists, err
```
melekes:
```suggestion
		return exists, nil
```

?

mutable_tree.go
```diff
@@ -268,7 +268,10 @@ func (tree *MutableTree) LazyLoadVersion(targetVersion int64) (int64, error) {
 		targetVersion = latestVersion
 	}
 
-	rootHash := tree.ndb.getRoot(targetVersion)
+	rootHash, err := tree.ndb.getRoot(targetVersion)
+	if err != nil {
+		return 0, err
```
melekes: so here is the question: at which point we should retry DB operation (say 5 times) before bubbling up an error to the caller?

marbar3778: in my view, I would prefer the application to handle this and here we present the error each time. @alexanderbez do you have a preference. The sdk is the main user of this repo so your preference matters most

alexanderbez: This change looks fine and behaves as I'd expect. If the version provided cannot be fetched, then the upstream caller should handle that error (either retry another version or fail completely).

melekes: Agree. My question of course goes beyond this case.

PR #199, opened 2020-01-15
errors: add some error handling

marbar3778:
tm-db 0.4.0 introduced more returns with errors, bubbling up more errors because of this.

### DIFF
--- a/nodedb.go
+++ b/nodedb.go
@@ -7,6 +7,8 @@ import (
 	"sort"
 	"sync"
 
+	"github.com/pkg/errors"
+
 	"github.com/tendermint/tendermint/crypto/tmhash"
 	dbm "github.com/tendermint/tm-db"
 )
@@ -159,22 +161,22 @@ func (ndb *nodeDB) Has(hash []byte) (bool, error) {
 
 	val, err := ndb.recentDB.Get(key)
 	if err != nil {
-		return false, err
+		return false, errors.Wrap(err, "recentDB")
 	}
 	if val != nil {
 		return true, nil
 	}
 
 	if ldb, ok := ndb.snapshotDB.(*dbm.GoLevelDB); ok {
-		exists, err2 := ldb.DB().Has(key, nil)
-		if err2 != nil {
-			return false, err2
+		exists, err := ldb.DB().Has(key, nil)
+		if err != nil {
+			return false, errors.Wrap(err, "snapshotDB")
 		}
 		return exists, nil
 	}
 	value, err := ndb.snapshotDB.Get(key)
 	if err != nil {
-		return false, err
+		return false, errors.Wrap(err, "snapshotDB")
 	}
 	return value != nil, nil
 }
@@ -290,7 +292,7 @@ func (ndb *nodeDB) deleteOrphans(version int64, memOnly bool) error {
 		})
 	}
 	if err != nil {
-		return err
+		return errors.Wrap(err, "snapshotDB err in delete")
 	}
 	return nil
 }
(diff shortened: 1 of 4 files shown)


nodedb.go
```diff
@@ -166,9 +166,9 @@ func (ndb *nodeDB) Has(hash []byte) (bool, error) {
 	}
 
 	if ldb, ok := ndb.snapshotDB.(*dbm.GoLevelDB); ok {
-		exists, err := ldb.DB().Has(key, nil)
-		if err != nil {
-			return false, err
+		exists, err2 := ldb.DB().Has(key, nil)
```
AdityaSripal: Why was renaming to `err2` necessary here?

nodedb.go
```diff
@@ -535,28 +548,42 @@ func (ndb *nodeDB) cacheNode(node *Node) {
...
 	ndb.mtx.Lock()
 	defer ndb.mtx.Unlock()
 
+	var err error
 	if ndb.opts.KeepEvery != 0 {
 		if ndb.opts.Sync {
-			ndb.snapshotBatch.WriteSync()
+			err = ndb.snapshotBatch.WriteSync()
+			if err != nil {
+				return err
```
AdityaSripal: Can these errors be wrapped with info on which DB failed?

nodedb.go
```diff
@@ -535,28 +548,42 @@ func (ndb *nodeDB) cacheNode(node *Node) {
...
+			}
 		}
 		ndb.snapshotBatch.Close()
 	}
 	if ndb.opts.KeepRecent != 0 {
 		if ndb.opts.Sync {
-			ndb.recentBatch.WriteSync()
+			err = ndb.recentBatch.WriteSync()
+			if err != nil {
+				return err
```
AdityaSripal: If there's an error here should we delete the write in snapshotDB? Could be ok with either behavior. Just want to see if there's been thought on this

Pro of deleting:
Maintains invariant that pruning logic is always correct

Pros of not deleting:
Simpler
Doesn't seem like it would create obvious issues if we don't delete

nodedb.go
marbar3778: `shadow: declaration of "err" shadows declaration at line 162 (govet)` I can `nolint` it

nodedb.go
marbar3778: yes will wrap them

nodedb.go
marbar3778: I haven't thought about it. i think it would be your call as you are the main person for this repo @AdityaSripal

nodedb.go
marbar3778: ah my local setup is more strict then the bot it doesnt seem to be throwing the error

nodedb.go
tnachen: I think attempting to delete to maintain state would be ideal, since down the road it will be harder to tell in what situations can we run into states that we don't anticipate in.

marbar3778: The previous logic was that if there is an error it would panic. If there is an error here then there is a larger problem at hand and the application would most likely panic. I do believe this is how the sdk will be handling it. I can add documentation for this in the godoc.

AdityaSripal: Perhaps it may be worth not deleting here, since the Delete function may return another error. It's unclear what to do in this case. And it still makes it possible for the invariant to be broken

I think we just return wrapped errors so the user at least knows where the execution failed and can determine DB state from this information

AdityaSripal: So im for leaving it as it is

PR #201, opened 2020-01-16
optimize

klim0v:
make slice with cap


tnachen:
Thanks the changes make sense, just curious did you run any before / after benchmark or found this to be a hot path, or just by reading the code?

klim0v:
> Thanks the changes make sense, just curious did you run any before / after benchmark or found this to be a hot path, or just by reading the code?

I wrote a test after the changes, it turned out to be slow

PR #205, opened 2020-01-31
export ProofLeafNode and ProofInnerNode

### LINKED ISSUE #73: Proposal: export `proofLeafNode`
Currently, `proofLeafNode` is not exported. While it is good practice to export only what is really neccessary (and hide fields and functionality that are not essential for the API), I think this type should be public/exported: If someone is about to write a verifier in another language, he needs to know enough details what a `RangeProof` consists of (and `proofLeafNode` is part of that):
https://github.com/tendermint/iavl/blob/f2c70966873724a199ea37d0cbfc8b50baa0f7bf/proof_range.go#L12-L21

Exporting the type would also create a proper godoc item. Similarly, `proofInnerNode` (as  `PathToLeaf` is just a type alias to `[]proofInnerNode`).

cc @jlandrews @jaekwon

tessr:
Closes #73.

### DIFF
--- a/proof.go
+++ b/proof.go
@@ -24,7 +24,7 @@ var (
 
 //----------------------------------------
 
-type proofInnerNode struct {
+type ProofInnerNode struct {
 	Height  int8   `json:"height"`
 	Size    int64  `json:"size"`
 	Version int64  `json:"version"`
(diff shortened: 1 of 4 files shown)


proof.go
```diff
@@ -92,18 +92,18 @@ func (pin proofInnerNode) Hash(childHash []byte) []byte {
 
 //----------------------------------------
 
-type proofLeafNode struct {
+type ProofLeafNode struct {
```
erikgrinaker: Now that we're exporting this, it should be documented as well.

tessr: I agree, and I started writing godoc for it, before I realized I didn't know enough about how these proofs work to write more than a cursory godoc. ("ProofLeafNode provides a merkle proof for a leaf node" is not terribly helpful.)

None of our proof types are documented anywhere at the moment, and this lack of clarity on my part was the motivation for opening https://github.com/tendermint/iavl/issues/204. Godoc for proofs can either happen along with that issue or right after it.

erikgrinaker: Ah, right. Sure, let's document it then.

PR #213, opened 2020-02-23
Add export/import API

### LINKED ISSUE #210: Export/import API
For [state sync](https://github.com/tendermint/tendermint/issues/828), we need an API to export a single `ImmutableTree` (i.e. a single version), and import it into an empty `MutableTree` (i.e. an empty database). Both export and import should be iterator-based, and contain the minimum amount of data necessary to reconstruct the tree.

Exporting and importing all versions from a `MutableTree` is considered out-of-scope for this initial API.

erikgrinaker:
Fixes #210. Exports the minimum amount of data for a single tree version to be able to reconstruct an identical tree:

```go
type ExportNode struct {
	Key     []byte
	Value   []byte
	Version int64
	Height  int8
}
```

It may be possible to drop `Height`, but I couldn't come up with an algorithm, and an additional byte per node seems acceptable. Relies on tree traversal, thus performance is likely suboptimal.

The iterator pattern here is slightly different from that used by `tm-db`, and is based on [Google's Go iterators](https://github.com/googleapis/google-cloud-go/wiki/Iterator-Guidelines). I'd be happy to change this if necessary.

I will be running a few export/import cycles of complete Cosmos Hub databases for benchmarking and testing, will post results before merging.

Spec: https://github.com/tendermint/iavl/blob/erik/export-import/docs/tree/export_import.md


erikgrinaker:
Below are benchmarks of an export/import of cosmoshub-3 at height 870068, obtained from here:

http://quicksync.chainlayer.io/cosmos/cosmoshub-3.20200221.0105.tar.lz4

Benchmarks were taken on a 2017 iMac with SSD, having flushed disk caches via `purge`. Data is exported from LevelDB-backed IAVL trees into memory, and then reimported from memory into new LevelDB IAVL trees on disk. There is no (de)serialization to disk of exported data, but this overhead is considered negligible.

```
Exporting cosmoshub-3 database at version 870068

acc          : 67131 nodes (33566 leaves) in 2.236s with size 3 MB
distribution : 66509 nodes (33255 leaves) in 4.537s with size 3 MB
evidence     : 0 nodes 
god          : 0 nodes 
main         : 1 nodes (1 leaves) in 0s with size 0 MB
mint         : 1 nodes (1 leaves) in 0s with size 0 MB
params       : 59 nodes (30 leaves) in 0s with size 0 MB
slashing     : 1128139 nodes (564070 leaves) in 29.299s with size 41 MB
staking      : 44573 nodes (22287 leaves) in 1.842s with size 3 MB
supply       : 1 nodes (1 leaves) in 0s with size 0 MB
upgrade      : 0 nodes 

Exported 11 stores with 1306414 nodes (653211 leaves) in 37.914s with total size 52 MB

Importing into new LevelDB stores

acc         : 67131 nodes (33566 leaves) in 352ms
distribution: 66509 nodes (33255 leaves) in 328ms
main        : 1 nodes (1 leaves) in 25ms
mint        : 1 nodes (1 leaves) in 25ms
params      : 59 nodes (30 leaves) in 25ms
slashing    : 1128139 nodes (564070 leaves) in 11.79s
staking     : 44573 nodes (22287 leaves) in 290ms
supply      : 1 nodes (1 leaves) in 28ms

Imported 11 stores with 1306414 nodes (653211 leaves) in 12.863s
```

The amount of data appears miniscule, bearing in mind that this is from a 45 GB data dump. Can someone please confirm these numbers?

If this is the order of magnitude data we are dealing with, then this naïve approach is more than sufficient for an initial state sync implementation.

With warm disk caches, the export duration is roughly halved:

```
Exporting cosmoshub-3 database at version 870068

acc          : 67131 nodes (33566 leaves) in 671ms with size 3 MB
distribution : 66509 nodes (33255 leaves) in 753ms with size 3 MB
evidence     : 0 nodes 
god          : 0 nodes 
main         : 1 nodes (1 leaves) in 0s with size 0 MB
mint         : 1 nodes (1 leaves) in 0s with size 0 MB
params       : 59 nodes (30 leaves) in 0s with size 0 MB
slashing     : 1128139 nodes (564070 leaves) in 16.447s with size 41 MB
staking      : 44573 nodes (22287 leaves) in 446ms with size 3 MB
supply       : 1 nodes (1 leaves) in 0s with size 0 MB
upgrade      : 0 nodes 

Exported 11 stores with 1306414 nodes (653211 leaves) in 18.317s with total size 52 MB
```

CC @milosevic

erikgrinaker:
@AdityaSripal We need to make sure a version that is being exported will not be pruned until after the export completes. What do you think would be the best way to do this?

AdityaSripal:
Hmm if this is being used for state-sync then it should be only using snapshot versions which never get pruned by application. However IAVL has no idea which versions will never get pruned and which may get pruned at later time.

Probably the best way to handle this is to have some state (maybe a map int64=>bool) on the mutable tree that lets you know which version(s) is currently being exported. If anyone tries to delete during this time, we check this state and return error. Once exporting is complete, we reset this state so that future delete calls on that version will succeed

erikgrinaker:
> Probably the best way to handle this is to have some state (maybe a map int64=>bool) on the mutable tree that lets you know which version(s) is currently being exported.

Yeah, I think so too. Someday™️ it would be nice to have proper concurrency control in IAVL, but that should work for now. Thanks!

export_test.go
```diff
@@ -0,0 +1,327 @@
...
+
+	require.Equal(t, []int{2, 3, 4}, tree.AvailableVersions())
+
+	exporter.Close()
+	tree.Set([]byte("e"), []byte{5})
+	_, _, err = tree.SaveVersion()
+	require.NoError(t, err)
+
+	// FIXME This seems like a bug in the version pruner; version 2 should have been pruned here
+	require.Equal(t, []int{2, 4, 5}, tree.AvailableVersions())
```
erikgrinaker: @AdityaSripal Seems to be a bug in the tree pruner here - it only deletes the most recent version that is eligible for pruning, and does not check for any previous versions.

marbar3778:
I like to go iterator pattern and believe it is something we should try to adopt for the db repo. should close in export be renamed to `Stop` since it is not returning an error?

Is there a way a quick markdown file could be added to the specs?

erikgrinaker:
>I like to google iterator pattern and believe it is something we should try to adopt for the db repo. should close in export be renamed to Stop since it is not returning an error?

Hm, interesting. I feel like `Close()` is the idiomatic way to do this in Go - see e.g. `io`, `http`, `sql`, and `net`, although those all return `error`. Don't have a strong opinion, but `Stop()` doesn't really imply "free up resources" to me.

>Is there a way a quick markdown file could be added to the specs?

Sure, I'll add a quick writeup tomorrow.

marbar3778:
>Hm, interesting. I feel like Close() is the idiomatic way to do this in Go - see e.g. io, http, sql, and net, although those all return error. Don't have a strong opinion, but Stop() doesn't really imply "free up resources" to me.

I wasn't sure how close you were following the google iteractors, ignore my comment 😄

erikgrinaker:
@marbar3778 Added export/import docs here: https://github.com/tendermint/iavl/blob/erik/export-import/docs/tree/export_import.md

benchmarks/cosmos-exim/main.go
```diff
@@ -0,0 +1,196 @@
...
+	"os"
+	"time"
+
+	"github.com/tendermint/iavl"
+	tmdb "github.com/tendermint/tm-db"
+)
+
+// stores is the list of stores in the CosmosHub database
+// FIXME would be nice to autodetect this
+var stores = []string{
```
ValarDragon: In lieu of auto-detecting this, perhaps it can be pushed to a config in a future PR.

ValarDragon:
Looks good to me!

One suggestion is to serialize the final hash along with export, so that when you re-import, its easy to convince yourself that you get the same hash (for debugging purports)

erikgrinaker:
> One suggestion is to serialize the final hash along with export, so that when you re-import, its easy to convince yourself that you get the same hash (for debugging purports)

I was thinking that would be the responsibility of the caller - it's easy enough by grabbing `ImmutableTree.Hash()` on export and comparing with `MutableTree.Hash()` after importing.

`Export`  doesn't actually do any serialization (in the marshalling sense) at all, it just gives an iterator over `ExportItem`, so that the caller is free to store/send the nodes in whatever format it wishes. The caller would probably want to generate an envelope containing the root hash of the tree for verification, among other data.

This is how state sync is planned to work, where it is the SDKs responsibility to marshal, compress, chunk, and verify exported items as part of multi-tree snapshots.

ValarDragon:
That makes sense, LGTM

benchmarks/cosmos-exim/main.go
```diff
@@ -0,0 +1,196 @@
...
+	if err != nil {
+		return 0, nil, err
+	}
+	version, err := tree.LoadVersion(0)
+	if err != nil {
+		return 0, nil, err
+	}
+	fmt.Printf("Exporting cosmoshub database at version %v\n\n", version)
+
+	exports := make(map[string][]*iavl.ExportNode, 16)
```
melekes: why 16?

erikgrinaker: Changed it to the actual number of stores to export instead.

erikgrinaker:
Ran another benchmark on Terra's `colombus-3` chain:

```
Exporting colombus-3 database at version 1225912

acc          : 2488933 nodes (1244467 leaves) in 2m2.264s with size 157 MB
distribution : 4833 nodes (2417 leaves) in 167ms with size 0 MB
evidence     : 0 nodes (0 leaves) in 0s with size 0 MB
god          : 0 nodes (0 leaves) in 0s with size 0 MB
gov          : 9 nodes (5 leaves) in 0s with size 0 MB
main         : 1 nodes (1 leaves) in 0s with size 0 MB
market       : 1 nodes (1 leaves) in 0s with size 0 MB
mint         : 0 nodes (0 leaves) in 0s with size 0 MB
oracle       : 1015 nodes (508 leaves) in 21ms with size 0 MB
params       : 87 nodes (44 leaves) in 1ms with size 0 MB
slashing     : 444911 nodes (222456 leaves) in 6.822s with size 16 MB
staking      : 3077 nodes (1539 leaves) in 44ms with size 0 MB
supply       : 1 nodes (1 leaves) in 0s with size 0 MB
treasury     : 87 nodes (44 leaves) in 1ms with size 0 MB
upgrade      : 0 nodes (0 leaves) in 0s with size 0 MB

Exported 15 stores with 2942955 nodes (1471483 leaves) in 2m9.32s with size 174 MB

Importing into new LevelDB stores

acc         : 2488933 nodes (1244467 leaves) in 24.32s with size 157 MB
distribution: 4833 nodes (2417 leaves) in 51ms with size 0 MB
evidence    : 0 nodes (0 leaves) in 23ms with size 0 MB
god         : 0 nodes (0 leaves) in 29ms with size 0 MB
gov         : 9 nodes (5 leaves) in 37ms with size 0 MB
main        : 1 nodes (1 leaves) in 31ms with size 0 MB
market      : 1 nodes (1 leaves) in 41ms with size 0 MB
mint        : 0 nodes (0 leaves) in 47ms with size 0 MB
oracle      : 1015 nodes (508 leaves) in 50ms with size 0 MB
params      : 87 nodes (44 leaves) in 44ms with size 0 MB
slashing    : 444911 nodes (222456 leaves) in 1.397s with size 16 MB
staking     : 3077 nodes (1539 leaves) in 45ms with size 0 MB
supply      : 1 nodes (1 leaves) in 34ms with size 0 MB
treasury    : 87 nodes (44 leaves) in 37ms with size 0 MB
upgrade     : 0 nodes (0 leaves) in 51ms with size 0 MB

Imported 15 stores with 2942955 nodes (1471483 leaves) in 26.235s with size 174 MB
```

This seems "good enough" for a first iteration. Any other chains we want to test before I merge this?

PR #227, opened 2020-03-24
fmt: add make format

marbar3778:
- add cmd for gofmt and goimports

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,19 +19,28 @@ ifeq ($(COLORS_ON),)
 else
 	go install $(CMDFLAGS) ./cmd/iaviewer
 endif
+.PHONY: install
 
 test:
 	@echo "--> Running go test"
 	@go test ./... $(LDFLAGS) -v --race
+.PHONY: test
 
 tools:
 	go get -v $(GOTOOLS)
+.PHONY: tools
+
+format:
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -19,19 +19,28 @@ ifeq ($(COLORS_ON),)
...
 	@go test ./... $(LDFLAGS) -v --race
+.PHONY: test
 
 tools:
 	go get -v $(GOTOOLS)
+.PHONY: tools
+
+format:
+	find . -name '*.go' -type f -not -path "*.git*" -not -name '*.pb.go' -not -name '*pb_test.go' | xargs gofmt -w -s
+	find . -name '*.go' -type f -not -path "*.git*"  -not -name '*.pb.go' -not -name '*pb_test.go' | xargs goimports -w -local github.com/tendermint/iavl
```
tessr: I have a preference to remove the -local flag here, especially since most of our import lists are very short in this repo.

marbar3778: This came about while reading the style guide since we will most likely want all repos to adhere to the style guide in Tendermint.

tessr: Heh, I need to go catch up on that discussion.

tessr: Ah! Ok. So I do agree with the guideline to separate the libs into std, external and tendermint. However, that's not what this flag does - it groups (non-iavl) tendermint libs with external libs, and puts iavl libs on their own.

marbar3778: ah okay, was viewing everything outside the repo as external, even tendermint. Will make the change to remove local.

PR #230, opened 2020-04-04
import: set correct version on empty import

erikgrinaker:


erikgrinaker:
@marbar3778 Do you mind merging this and cutting a release? I don't have access.

PR #238, opened 2020-04-24
.github: configure stalebot

tessr:
Stalebot had been turned on in the org settings but was missing a config file. 

No more!

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,47 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 60
+
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 9
+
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,47 @@
...
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 9
+
+# Only issues or pull requests with all of these labels are check if stale. Defaults to `[]` (disabled)
+onlyLabels: []
+
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+exemptLabels:
+  - major-release
```
marbar3778, line 15: i don't think this repo has major-release, should we make it `S:wip`

erikgrinaker: Major release can be removed. I'd say we should use drafts over wip, would be nice if dependabot ignored drafts.

erikgrinaker: Doesn't look like Stalebot supports draft handling, there's an open issue for it: https://github.com/probot/stale/issues/265. I'll We can use `s:wip` for now, I'll update and merge.

PR #239, opened 2020-04-29
FlushVersion API

alexanderbez:
Implement `FlushVersion(version int64)` which attempts to flush a previously committed version to disk.

TODO:

- [x] Test with Gaia manually halting and verifying the latest version was persisted (with all types of pruning strategies).
- [x] Remove nodes from `recentDB` when flushing.

ref:

* https://github.com/cosmos/cosmos-sdk/pull/6199
* https://github.com/cosmos/gaia/pull/403


nodedb.go
```diff
@@ -126,6 +126,10 @@ func (ndb *nodeDB) GetNode(hash []byte) *Node {
 
 // SaveNode saves a node to disk.
 func (ndb *nodeDB) SaveNode(node *Node, flushToDisk bool) {
+	ndb.saveNodeBatch(node, flushToDisk, ndb.recentBatch, ndb.snapshotBatch)
+}
+
+func (ndb *nodeDB) saveNodeBatch(node *Node, flushToDisk bool, rb, sb dbm.Batch) {
```
alexanderbez, line 132: Created an aux method so we can manually provide batch objects to not clobber the existing batch objects on the actual `nodeDB` when flushing a previous version.

Note to authors/maintainers: Having the batch objects be part of the actual `nodeDB` seems like a code smell and might warrant an architecture redesign.

nodedb.go
```diff
@@ -195,6 +211,10 @@ func (ndb *nodeDB) SaveTree(root *Node, version int64) []byte {
 // calls _hash() on the given node.
 // TODO refactor, maybe use hashWithCount() but provide a callback.
 func (ndb *nodeDB) SaveBranch(node *Node, flushToDisk bool) []byte {
+	return ndb.savebranchBatch(node, flushToDisk, ndb.recentBatch, ndb.snapshotBatch)
+}
+
+func (ndb *nodeDB) savebranchBatch(node *Node, flushToDisk bool, rb, sb dbm.Batch) []byte {
```
alexanderbez, line 217: ditto; see comment above

nodedb.go
```diff
@@ -671,6 +698,41 @@ func (ndb *nodeDB) decrVersionReaders(version int64) {
...
+	//
+	// NOTE: We ignore any write that happen to recentBatch.
+	rb := ndb.recentDB.NewBatch()
+	sb := ndb.snapshotDB.NewBatch()
+
+	// save branch, the root, and the necessary orphans
+	node := ndb.GetNode(rootHash)
+	ndb.savebranchBatch(node, true, rb, sb)
+
+	traverseOrphansVersionFromDB(ndb.recentDB, version, func(k, v []byte) {
```
alexanderbez, line 714: @AdityaSripal is this the correct logic for flushing orphans? Does the test case cover this case? Are there corner-cases we aren't considering?

mutable_tree.go
```diff
@@ -431,19 +437,45 @@ func (tree *MutableTree) GetVersioned(key []byte, version int64) (
...
+	if rootHash == nil {
+		return ErrVersionDoesNotExist
+	}
+
+	ok, err := tree.ndb.HasSnapshot(rootHash)
+	if err != nil {
+		return err
+	}
+	if ok {
+		return ErrVersionAlreadyFlushed
```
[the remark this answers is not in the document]
alexanderbez: Wouldn't we incur IO overhead of already writing something that was written? We can short-circuit the entire procedure by just checking.

nodedb.go
alexanderbez: Good point, I'll add the comment as a godoc.

mutable_tree.go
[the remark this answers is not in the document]
erikgrinaker: Sure, we can check, but I'd say this should be idempotent and simply return `nil`.

nodedb.go
```diff
@@ -671,6 +702,41 @@ func (ndb *nodeDB) decrVersionReaders(version int64) {
...
+		var fromVersion, toVersion int64
+		orphanKeyFormat.Scan(k, &toVersion, &fromVersion)
+
+		ndb.saveOrphan(v, fromVersion, toVersion, true)
+	})
+
+	sb.Set(ndb.rootKey(version), node.hash)
+
+	if ndb.opts.Sync {
```
erikgrinaker, line 727: This is sort of surprising to me, I would expect a `flush` method to flush the data all the way to disk (without sync, the OS buffers the write in memory). There could perhaps be a parameter `forceSync` or something.

nodedb.go
erikgrinaker: I'm not very familiar with the orphan logic. Is the purpose of `flushVersion` to take a version committed to memory and save it to disk instead? If so, why are we only considering orphans, shouldn't we consider all live objects?

alexanderbez: The purpose is to flush a previously _committed_ version that is in memory, yes. The flushing logic here is identical to what you'll find in `SaveVersion` so that's what I went off of. It's also what @AdityaSripal suggested to do.

mutable_tree.go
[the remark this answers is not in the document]
alexanderbez: Fair. Either way, it's idempotent, but perhaps it is better to simply return `nil` 👍

nodedb.go
alexanderbez: Indeed, good catch. This logic was based on what is found in `nodeDB#Commit`. However, since we want to flush to disk, I see no reason to consider the non-`ndb.opts.Sync` case -- we should always use `WriteSync`. Correct me if I'm wrong.

erikgrinaker: I think that makes sense.

alexanderbez:
Wired everything up in SDK and Gaia to test this.

1. Run a network with default (syncable) pruning settings
2. Kill a node
3. Observe logs

```
panic: nodeDB.GetNode() requires hash

goroutine 162 [running]:
github.com/tendermint/iavl.(*nodeDB).GetNode(0xc000e1f980, 0x2374ab8, 0x0, 0x0, 0x0)
	github.com/tendermint/iavl@v0.13.3/nodedb.go:86 +0x8af
github.com/tendermint/iavl.(*nodeDB).flushVersion(0xc000e1f980, 0x8, 0x2374ab8, 0x0, 0x0, 0x0, 0x0)
	github.com/tendermint/iavl@v0.13.3/nodedb.go:715 +0xc3
github.com/tendermint/iavl.(*MutableTree).FlushVersion(0xc0012a69c0, 0x8, 0xc000e72ee0, 0x18fdf60)
	github.com/tendermint/iavl@v0.13.3/mutable_tree.go:463 +0x13d
github.com/cosmos/cosmos-sdk/store/iavl.(*Store).FlushVersion(...)
	github.com/cosmos/cosmos-sdk@v0.34.4-0.20200430150743-930802e7a13c/store/iavl/store.go:123
github.com/cosmos/cosmos-sdk/store/rootmulti.(*Store).FlushLatestVersion(0xc000c84c60, 0x7f1d29101c48)
	github.com/cosmos/cosmos-sdk@v0.34.4-0.20200430150743-930802e7a13c/store/rootmulti/store.go:317 +0x12a
github.com/cosmos/cosmos-sdk/baseapp.(*BaseApp).FlushLatestVersion(...)
	github.com/cosmos/cosmos-sdk@v0.34.4-0.20200430150743-930802e7a13c/baseapp/baseapp.go:234
github.com/cosmos/gaia/app.(*GaiaApp).FlushLatestVersion(0xc001284000, 0x18f7f40)
	github.com/cosmos/gaia/app/app.go:418 +0x37
github.com/cosmos/cosmos-sdk/server.startInProcess.func2()
	github.com/cosmos/cosmos-sdk@v0.34.4-0.20200430150743-930802e7a13c/server/start.go:214 +0xfd
github.com/cosmos/cosmos-sdk/server.TrapSignal.func1(0xc00057e840, 0xc00020a5d0)
	github.com/cosmos/cosmos-sdk@v0.34.4-0.20200430150743-930802e7a13c/server/util.go:215 +0xc1
created by github.com/cosmos/cosmos-sdk/server.TrapSignal
	github.com/cosmos/cosmos-sdk@v0.34.4-0.20200430150743-930802e7a13c/server/util.go:211 +0xcb
```

erikgrinaker:
> It could very well be an error in my wiring up. I'm committing `rs.lastCommitInfo.Version`, whereas I may need to commit `rs.lastCommitInfo.Version-1`?

I'd think it should be `Version`.

alexanderbez:
More updates:

I couldn't seem to figure out for the life of me why some root node hashes didn't exist in the recentDB when flushing. I noticed, to what seemed to be consistent, that the root node hash didn't exist for `x/ibc` and `x/evidence` stores.

I then thought well these modules do not inherently have any state by default -- so maybe their root node hash should be empty. I decided to _not_ error in this case. As expected, the flush worked without error.

However, after restarting a node that was flushed, the restart resulted in a panic:

```shell
I[2020-05-06|18:39:23.702] starting ABCI with Tendermint                module=main
?��{��/]�6ސF��g��*���/���hash 7489a50d3fb8c67b1be62f5d9936de904687a3670e95e82abab180122f9d9f95 corresponding to nodeKey nt��

goroutine 66 [running]:
github.com/tendermint/iavl.(*nodeDB).GetNode(0xc000317680, 0xc000486a80, 0x20, 0x20, 0x0)
	github.com/tendermint/iavl@v0.13.3/nodedb.go:109 +0x7bf
github.com/tendermint/iavl.(*Node).getLeftNode(...)
	github.com/tendermint/iavl@v0.13.3/node.go:410
github.com/tendermint/iavl.(*Node).traverseInRange(0xc0004fd040, 0xc000c0aec0, 0xc0000b6c50, 0x10, 0x10, 0xc0000b6c60, 0x10, 0x10, 0x1, 0xc00008ff48, ...)
	github.com/tendermint/iavl@v0.13.3/node.go:473 +0x33c
github.com/tendermint/iavl.(*ImmutableTree).IterateRange(0xc000c0aec0, 0xc0000b6c50, 0x10, 0x10, 0xc0000b6c60, 0x10, 0x10, 0xc000de1e01, 0xc00004ffb8, 0x1)
	github.com/tendermint/iavl@v0.13.3/immutable_tree.go:187 +0xd2
github.com/cosmos/cosmos-sdk/store/iavl.(*iavlIterator).iterateRoutine(0xc00011fef0)
	github.com/cosmos/cosmos-sdk@v0.34.4-0.20200430150743-930802e7a13c/store/iavl/store.go:379 +0x98
created by github.com/cosmos/cosmos-sdk/store/iavl.newIAVLIterator
	github.com/cosmos/cosmos-sdk@v0.34.4-0.20200430150743-930802e7a13c/store/iavl/store.go:372 +0x18c
```

nodedb.go
```diff
@@ -195,6 +211,14 @@ func (ndb *nodeDB) SaveTree(root *Node, version int64) []byte {
...
 // TODO refactor, maybe use hashWithCount() but provide a callback.
 func (ndb *nodeDB) SaveBranch(node *Node, flushToDisk bool) []byte {
+	return ndb.savebranchBatch(node, flushToDisk, ndb.recentBatch, ndb.snapshotBatch)
+}
+
+// TODO: Reconsider design to not have batch objects be fields of a nodeDB type.
+// Instead, batch objects should be created when needed as passed as arguments
+// where needed. This allows the IO flow to be easier to reason about and impproves
+// testability.
+func (ndb *nodeDB) savebranchBatch(node *Node, flushToDisk bool, rb, sb dbm.Batch) []byte {
```
melekes, line 221:
```suggestion
func (ndb *nodeDB) saveBranchBatch(node *Node, flushToDisk bool, rb, sb dbm.Batch) []byte {
```

nodedb.go
```diff
@@ -671,6 +702,35 @@ func (ndb *nodeDB) decrVersionReaders(version int64) {
...
 }
 
+func (ndb *nodeDB) flushVersion(version int64, rootHash []byte) error {
+	// Create new recentDB and snapshotDB batch objects. We don't use the current
+	// batch objects of the nodeDB as we don't want to write state prematurely or
+	// have conflicting state.
+	//
+	// NOTE: We ignore any write that happen to recentBatch.
+	rb := ndb.recentDB.NewBatch()
+	sb := ndb.snapshotDB.NewBatch()
```
melekes, line 712: Do you have to `.Close` batch or something?

erikgrinaker:
> These trees are empty and there is special logic for empty trees in SaveVersion. I think I need to follow the same logic in FlushVersion.

Ah yes, this has tripped me up before as well. We should probably try to use common logic for `SaveVersion` and `FlushVersion`.

nodedb.go
```diff
@@ -606,7 +638,7 @@ func (ndb *nodeDB) getRoot(version int64) ([]byte, error) {
 			return nil, err
 		}
 		// TODO: maybe I shouldn't check in snapshot if it isn't here
-		if memroot != nil {
+		if len(memroot) != 0 {
```
erikgrinaker, line 641: This change broke tests, reverting it.

```
    mutable_tree_test.go:145: 
        	Error Trace:	mutable_tree_test.go:145
        	Error:      	Expected nil, but got: version does not exist
        	Test:       	TestEmptyRecents
```

alexanderbez: Cool, but note the tests should be fixed, no? It's better to check length opposed to nil.

erikgrinaker: But `nil` and `[]byte{}` are not the same thing. `nil` means the node does not exist, `[]byte{}` means the node exists but is empty. For root nodes, this is semantically different: a missing version vs an empty version.

erikgrinaker: And `len(nil) == len([]byte{}) == 0`

alexanderbez: Right, I was under the impression we'd never store `nil`, hence we should check the length.

erikgrinaker: Well, no, but this check is for whether the version exists in the recentDB, i.e. that it's not `nil`. The difference between `nil` and `[]byte{}` matters here.

alexanderbez:
Not sure how you got this to work locally @erikgrinaker -- I still see failed invariants/invalid state when restarting a node after it was flushed manually (via SIGINT).

erikgrinaker:
> Not sure how you got this to work locally @erikgrinaker -- I still see failed invariants/invalid state when restarting a node after it was flushed manually (via SIGINT).

Still seems to work fine here:

```sh
cd iavl
git checkout bez/flush-committed-version
git pull

cd ../cosmos-sdk
git checkout bez/dnm-iavl-flush-debugging
git pull
sed -i '' -e 's|/Users/aleksbez/code/tendermint/iavl|../iavl|g' go.mod

cd ../gaia
git checkout bez/dnm-iavl-flush-debugging
git pull
sed -i '' -e 's|/Users/aleksbez/go/src/github.com/cosmos/cosmos-sdk|../cosmos-sdk|g' go.mod
sed -i '' -e 's|/Users/aleksbez/code/tendermint/iavl|../iavl|g' go.mod
sed -i '' -e 's|docker-compose up -d|docker-compose up|g' Makefile
go mod tidy
rm -rf build/node* && make localnet-start

gaiadnode2    | I[2020-05-18|09:34:43.768] Executed block                               module=state height=1 validTxs=0 invalidTxs=0
gaiadnode3    | I[2020-05-18|09:34:43.768] Executed block                               module=state height=1 validTxs=0 invalidTxs=0
gaiadnode2    | I[2020-05-18|09:34:43.771] Committed state                              module=state height=1 txs=0 
[...]

docker exec gaiadnode3 sh -c 'ps aux | grep "gaiad start" | grep -v grep | awk "{print \$2}" | xargs kill -s INT'

gaiadnode3    | I[2020-05-18|09:36:31.564] flushed latest version                       module=main version=21
gaiadnode3    | I[2020-05-18|09:36:31.569] exiting...                                   module=main 

docker start gaiadnode3

gaiadnode3    | I[2020-05-18|09:36:45.887] Executed block                               module=state height=22 validTxs=0 invalidTxs=0
gaiadnode3    | I[2020-05-18|09:36:45.890] Committed state                              module=state height=22 txs=0 appHash=63F048E39F4015220FBD0FF45155AA2BBAA3D4C324F1F7A853E685610CE8BB06```

**Comment by alexanderbez** [comment]
Sorry, I forgot to pull the latest changes from the SDK which contained a critical fix:

```go
flushCommitInfo(rs.db, rs.lastCommitInfo.Version, rs.lastCommitInfo)
```

alexanderbez:
So we can successfully restart the nodes but doesn't really help much at the moment because, on a restart, the manually flushed version will be pruned via `PruneRecentVersions` during `SaveVersion`. Need to think around this somehow...

erikgrinaker:
I think we should rethink our approach to pruning in general. Storing on-disk metadata is probably the way to go - we should have a set of metadata about each version, where persistence is one of those flags. This shouldn't just be used for manually flushed versions, but also for the other pruning behavior, which would simplify a lot of the pruning logic and avoid a bunch of edge cases.

alexanderbez:
Yes, I do agree and that was my intent -- all flushed versions, manual or not, are saved to disk with metadata. Alternatively, before pruning we can call and check `HasSnapshot`, this is a simpler approach but introduces more overhead.

Overall, yes, I agree. The pruning design needs serious reconsideration. The recent changes to pruning really are causing headaches...

erikgrinaker:
> But we need to pick a direction for now at least -- metadata or `HasSnapshot` calls?

I think we should move towards metadata, but stored separately per version. We may want to store each metadata item for each version as a separate key, or store all metadata for a version as a single key - the optimal choice will depend on access patterns.

I am a bit wary of adding this without a clear plan of where we want to go. Since this will be persisted it can be a bit harder to change later, and without a clear plan it may end up causing further headaches.

alexanderbez:
I propose we persist flushed version metadata per-version (i.e. indexed by key). The metadata can be the following:

```go
type CommitMetadata struct {
  Version uint64
  Timestamp int64
  Root []byte
}
```

alexanderbez:
@AdityaSripal thoughts?

erikgrinaker:
> ```go
> type CommitMetadata struct {
>   Version   uint64
>   Timestamp int64
>   Root      []byte
> }
> ```

I was thinking we'd store something about how this version should be handled/pruned. As a totally naïve example, something like `Prune: never` or `Prune: auto`.

alexanderbez:
Ahhh right, good catch -- flushed versions can be pruned at a later point in time 👍

cwgoes:
> > ```go
> > type CommitMetadata struct {
> >   Version   uint64
> >   Timestamp int64
> >   Root      []byte
> > }
> > ```
> 
> I was thinking we'd store something about how this version should be handled/pruned. As a totally naïve example, something like `Prune: never` or `Prune: auto`.

I can imagine this coming in useful if we want to give users the specific ability to commit a version of the tree to disk which they do *not* want pruned - maybe a use-case would be committing the state when a particular event occurs, but otherwise pruning. Potentially we might also want a user-specifiable "name" in the metadata for later inspection or retrievability.

alexanderbez:
Ok great, looks we decent enough buy-in. I'll proceed with the metadata approach. Thanks for the input everyone :D

nodedb.go
```diff
@@ -195,6 +211,14 @@ func (ndb *nodeDB) SaveTree(root *Node, version int64) []byte {
 // calls _hash() on the given node.
 // TODO refactor, maybe use hashWithCount() but provide a callback.
 func (ndb *nodeDB) SaveBranch(node *Node, flushToDisk bool) []byte {
+	return ndb.saveBranchBatch(node, flushToDisk, ndb.recentBatch, ndb.snapshotBatch)
+}
+
+// TODO: Reconsider design to not have batch objects be fields of a nodeDB type.
```
melekes, line 217: do we have an issue for this?

nodedb.go
```diff
@@ -671,6 +712,55 @@ func (ndb *nodeDB) decrVersionReaders(version int64) {
...
+		orphanKeyFormat.Scan(k, &toVersion, &fromVersion)
+
+		ndb.saveOrphan(v, fromVersion, toVersion, true)
+	})
+
+	if err := sb.WriteSync(); err != nil {
+		return fmt.Errorf("failed to write (sync) the snapshot batch: %w", err)
+	}
+
+	if err := rb.WriteSync(); err != nil {
```
melekes, line 757: what if this write fails, but `sb.WriteSync()` succeeds? will the IAVL end up in inconsistent state?

nodedb.go
alexanderbez: No we do not -- I can create one :0

nodedb.go
alexanderbez: Technically, we don't even need the `rb.WriteSync` call, because everything we should be writing to is the snapshotDB batch object. I'll remove this line for clarity.

alexanderbez:
Update: Gaia restart works and picks up after manually flushed version (via SIGINT). However, I'm not able to query against the manually flushed version. I need to dig into it and see if the version is not pruned by accident or if something else in IAVL needs to change.

erikgrinaker:
> Update: Gaia restart works and picks up after manually flushed version (via SIGINT). However, I'm not able to query against the manually flushed version. I need to dig into it and see if the version is not pruned by accident or if something else in IAVL needs to change.

Could it be that it's pruned by the additional pruning logic in the SDK?

alexanderbez:
It could very well be the case @erikgrinaker. I have a good understanding of where to look and debug (i.e. add print statements). I'll post updates shortly.

alexanderbez:
Derp, it's probably because I'm not updating the version metadata during `FlushVersion` 😆

PR #243, opened 2020-05-08
ci: change to official linter

marbar3778:
- change linter in ci
- enable nolintlint


marbar3778:
yes, they were unused.

PR #250, opened 2020-05-22
ADR 001 - Flush Version

alexanderbez:
references:

- https://github.com/tendermint/iavl/pull/239

### DIFF
--- a/docs/architecture/adr-001-flush-version.md
+++ b/docs/architecture/adr-001-flush-version.md
@@ -1,0 +1,42 @@
+# ADR 001: Flush Version
+
+## Changelog
+
+- May 22, 2020: Initial Draft
+
+## Status
+
+Proposed
+
+## Context
+
+> This section describes the forces at play, including technological, political, social, and project
+> local. These forces are probably in tension, and should be called out as such. The language in this
+> section is value-neutral. It is simply describing facts.
+> {context body}
+
+## Decision
+
+> This section describes our response to these forces. It is stated in full sentences, with active
+> voice. "We will ...".
+> {decision body}
+
+## Consequences
+
+> This section describes the resulting context, after applying the decision. All consequences should
+> be listed here, not just the "positive" ones. A particular decision may have positive, negative,
+> and neutral consequences, but all of them affect the team and project in the future.
+
+### Positive
+
+{positive consequences}
+
+### Negative
+
+{negative consequences}
+
+### Neutral
+
+{neutral consequences}
+
+## References
--- a/docs/architecture/adr-001-flush-version.md
+++ b/docs/architecture/adr-001-flush-version.md
@@ -63,7 +63,7 @@ type CommitMetadata struct {
 Currently, during `SaveVersion` we evaluate which versions to prune via `PruneRecentVersions` using
 the `Options` provided to the tree's `nodeDB`. The change we propose is now to use `CommitMetadata.Prune`
 to determine if we should actually prune a previously committed version. Under strategy `"auto"`, the
-current logic will be executed. However, if a given version is to be pruned but it's strategy is `"never"`,
+current logic will be executed. However, if a given version is to be pruned but its strategy is `"never"`,
 the version will not be pruned. During `FlushVersion`, we automatically set the strategy to `"never"`,
 otherwise, during `SaveVersion` it's set to `"auto"`.
 
--- a/docs/architecture/adr-001-flush-version.md
+++ b/docs/architecture/adr-001-flush-version.md
@@ -46,22 +46,11 @@ flushed to disk for every version saved, regardless if the version is flushed to
 type will also provide us with added benefits around pruning model and how clients treat snapshots.
 
 ```go
-type VersionStatus string
-
-const (
-  Saving   VersionStatus = "saving"   // set when a version is about to be saved
-  Buffered VersionStatus = "buffered" // set when a version has been saved to volatile storage
-  Saved    VersionStatus = "saved"    // set when a version has been saved to disk
-  Pruned   VersionStatus = "pruned"   // set when a version has been pruned
-  Deleted  VersionStatus = "deleted"  // set when a version has been explicitly deleted
-)
-
 type VersionMetadata struct {
   Version   int64         // tree version for the corresponding metadata
   Committed int64         // the UNIX timestamp of when the metadata was committed to disk
-  Updated   int64         // the UNIX timestamp of when the metadata was updated (see VersionStatus)
+  Updated   int64         // the UNIX timestamp of when the metadata was updated
   RootHash  []byte        // the root hash of the tree for the corresponding metadata
-  Status    VersionStatus // the status of the version
   Snapshot  bool          // if this version corresponds to a version that is flushed to disk
 }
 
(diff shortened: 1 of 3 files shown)


docs/architecture/adr-001-flush-version.md
```diff
@@ -0,0 +1,111 @@
...
+
+The IAVL library recently underwent changes to the pruning and version commitment model. Specifically,
+the invariant that a version is flushed to disk when it is committed via `SaveVersion` no longer holds
+true. Instead, versions are kept in memory and periodically flushed to disk and then pruned based on
+the client supplied pruning strategy parameters. For more detailed information, see the
+[PRUNING](../tree/PRUNING.md) document.
+
+These changes, while drastically improving performance under certain circumstances, introduces certain
+tradeoffs. Specifically, an application with a deterministic state-machine that commits and merkle-izes
+state via IAVL tree(s), can no longer guarantee that when a state is committed, that it exists on disk
```
cwgoes, line 21:
```suggestion
state via IAVL tree(s), can no longer guarantee that a state is written to disk when it is committed
```

docs/architecture/adr-001-flush-version.md
```diff
@@ -0,0 +1,111 @@
...
+and must rely on some sort of state replay mechanism (e.g. Tendermint p2p gossip). While this inherently
```
cwgoes, line 22:
```suggestion
and must rely on some sort of state replay mechanism (e.g. Tendermint peer-to-peer gossip & state machine execution). While this inherently
```

docs/architecture/adr-001-flush-version.md
```diff
@@ -0,0 +1,111 @@
...
+is not necessarily a problem, it becomes a problem under certain contexts that depend on a specific
+version existing on disk.
+
+One such example is live upgrades. Specifically, when a live upgrade occurs in some application, typically
+a new binary will be started in place of the old binary and some set of business logic will need to
+be executed in order to migrate and handle old and/or invalid state to make it compatible with the
+new version. Being that these operations must occur on the latest canonical state, we must ensure the
+latest version is committed and flushed to disk so all operators and clients see the same view when
+doing this upgrade. Hence, we need a means to manually signal to the IAVL tree, that a specific
```
cwgoes, line 31:
```suggestion
doing this upgrade. Hence, we need a means to manually signal to the IAVL tree that a specific
```

docs/architecture/adr-001-flush-version.md
```diff
@@ -0,0 +1,111 @@
...
+  Timestamp int64
+  RootHash  []byte
+  Prune     PruneStrategy
+}
+```
+
+Currently, during `SaveVersion` we evaluate which versions to prune via `PruneRecentVersions` using
+the `Options` provided to the tree's `nodeDB`. The change we propose is now to use `CommitMetadata.Prune`
+to determine if we should actually prune a previously committed version. Under strategy `"auto"`, the
+current logic will be executed. However, if a given version is to be pruned but it's strategy is `"never"`,
```
cwgoes, line 66:
```suggestion
current logic will be executed. However, if a given version is to be pruned but its strategy is `"never"`,
```

docs/architecture/adr-001-flush-version.md
```diff
@@ -0,0 +1,111 @@
...
+
+We propose to introduce a new API, `FlushVersion(version int64)`. This new API will attempt to manually
+flush a previously committed version to disk. It will return an error if the version does not exist
+or if the underlying flush fails. If the version is already flushed, the method performs a no-op
+and no error is returned. After the version is flushed to disk and the version is not the latest
+version, we will also remove the version from the recent (in-memory) database via `DeleteVersionFromRecent`.
+
+In order to not have a manually flushed version pruned at a later point in time
+(e.g. after an application restarts), we also introduce a new type, `CommitMetadata`, that will be
+flushed to disk for every version saved, regardless if it's flushed to disk.
```
erikgrinaker, line 45: What is the benefit of always flushing this to disk? Will metadata be kept even for deleted versions? What about future versions that were discarded via `LoadVersionForOverwriting`? Should there be an API to fetch and delete metadata?

docs/architecture/adr-001-flush-version.md
```diff
@@ -0,0 +1,111 @@
...
+  PruneStrategyNever PruneStrategy = "never"
+  PruneStrategyAuto  PruneStrategy = "auto"
+)
+
+type CommitMetadata struct {
+  Version   uint64
+  Timestamp int64
+  RootHash  []byte
+  Prune     PruneStrategy
+}
```
erikgrinaker, line 60: Perhaps the metadata should include whether the version is committed to disk or in memory? That way applications can query this state.

docs/architecture/adr-001-flush-version.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+Currently, during `SaveVersion` we evaluate which versions to prune via `PruneRecentVersions` using
+the `Options` provided to the tree's `nodeDB`. The change we propose is now to use `CommitMetadata.Prune`
+to determine if we should actually prune a previously committed version. Under strategy `"auto"`, the
+current logic will be executed. However, if a given version is to be pruned but its strategy is `"never"`,
+the version will not be pruned. During `FlushVersion`, we automatically set the strategy to `"never"`,
+otherwise, during `SaveVersion` it's set to `"auto"`.
```
erikgrinaker, line 68: I'm wondering if we can replace much of the current pruning logic with this. For example, we could set `Prune: yes` if the version is to be pruned on the next commit, and `Prune: no` if it is to be kept (i.e. a snapshot version). This is sort of a half-baked idea from my side, I'd like to think about it until tomorrow if that's ok.

docs/architecture/adr-001-flush-version.md
```diff
@@ -0,0 +1,111 @@
...
+type PruneStrategy string
+
+const (
+  PruneStrategyNever PruneStrategy = "never"
+  PruneStrategyAuto  PruneStrategy = "auto"
+)
+
+type CommitMetadata struct {
+  Version   uint64
+  Timestamp int64
```
erikgrinaker, line 57: Why not `time.Time`?

docs/architecture/adr-001-flush-version.md
alexanderbez: > What is the benefit of always flushing this to disk?

It makes the model easier to reason about. Alternatively, we can flush to disk periodically. Most importantly, we need to think about any edge cases where a client may halt and a subset of metadata was not flushed to disk. What happens in this case when the metadata needs to be fetched and isn't there?

> Will metadata be kept even for deleted versions?

In this ADR, that is the implicit assumption, yes. Do you propose we delete metadata along with the version when the version is deleted?

> What about future versions that were discarded via `LoadVersionForOverwriting`?

This would overwrite any metadata that exists for an already committed version.

> Should there be an API to fetch and delete metadata?

There will be an API to retrieve metadata, but not delete it. Metadata should only be deleted as the internal API sees fit.

docs/architecture/adr-001-flush-version.md
alexanderbez: They're just much easier to use, especially if this becomes a protobuf message.

docs/architecture/adr-001-flush-version.md
alexanderbez: > Perhaps the metadata should include whether the version is committed to disk or in memory? 

Yes, good idea!

> This may also replace the current isSnapshotVersion pruning logic, which will change if the strategy changes - which could cause IAVL to look in the wrong database, since the assumption no longer holds.

Not sure I follow this?

docs/architecture/adr-001-flush-version.md
erikgrinaker: > we need to think about any edge cases where a client may halt and a subset of metadata was not flushed to disk.

We could do two updates: store the metadata with a flag saying that the version is about to be saved, and then flip the flag once it has been saved. That would allow us to detect incomplete saves as well.

> Do you propose we delete metadata along with the version when the version is deleted?

No, I think it probably makes sense to keep metadata around anyway. But we would have to handle backwards compatibility with existing applications, and handle the missing metadata.

docs/architecture/adr-001-flush-version.md
alexanderbez: Yeah, my hunch when writing this was that we can simultaneously cleanup the pruning logic.  Specifically, `CommitMetadata.Prune` should be a `bool` instead of a `string`. Let me think on this some more as well, but I think we can simplify the current model.

docs/architecture/adr-001-flush-version.md
alexanderbez: > We could do two updates: store the metadata with a flag saying that the version is about to be saved, and then flip the flag once it has been saved. That would allow us to detect incomplete saves as well.

Could you elaborate a bit? Store the metadata where? Do you think it'll be easier to just persist to disk every save? It'll only be a few bytes.

erikgrinaker: > Could you elaborate a bit?

I mean, when we call `SaveVersion()` we would do the following:

* Store `CommitMetadata{Version: 10, Status: "saving"}` to disk
* Store the version to memory or disk as appropriate
* Store `CommitMetadata{Version: 10, Status: "saved"}` to disk

> Store the metadata where? Do you think it'll be easier to just persist to disk every save? It'll only be a few bytes.

Yeah, we should always keep the metadata on disk.

docs/architecture/adr-001-flush-version.md
erikgrinaker: > Not sure I follow this?

There's a lot of logic sprinkled around that calls `isSnapshotVersion()` to figure out whether it should look in memory or on disk - which is this:

```go
func (ndb *nodeDB) isSnapshotVersion(version int64) bool {
	return ndb.opts.KeepEvery != 0 && version%ndb.opts.KeepEvery == 0
}
```

What if I stored versions 10, 20, and 30 to disk (KeepEvery=10), but then changed my settings to KeepEvery=20. `isSnapshotVersion()` will now report `false` for versions 10 and 30, even though they are clearly stored on disk. If we instead rely on metadata, it would recognize that versions 10 and 30 are on disk regardless of what my pruning settings say.

alexanderbez: I see. Yes, I agree and that makes sense. Although I don't think it's possible to change the pruning strategy on the fly, this is a case we should consider 👍

docs/architecture/adr-001-flush-version.md
alexanderbez: I see. What does this buy us as opposed to just saving metadata to disk once? It can even be part of the snapshot batch to make it atomic.

docs/architecture/adr-001-flush-version.md
erikgrinaker: Yes, it might be as simple as that. There are a couple of things I'm thinking about here:

* We rely on `KeepRecent` to figure out how many versions to keep in memory. It would be nice if we could use a common mechanism to prune from memory and prune from disk.

* The SDK currently has hacks to only keep the latest `sdk.KeepEvery` version around, but all `sdk.SnapshotEvery`. We should eventually push this logic down into IAVL.

* The pruning settings may change, we should handle this in a reasonable way (see my [other comment](https://github.com/tendermint/iavl/pull/250/files#r430008815)).

* The current approach of `KeepEvery` was originally meant for state sync snapshots, but those don't need IAVL versions to be kept around, so I'm not sure if this option makes a whole lot of sense anymore.

Generally, it's always a good idea to separate policy and mechanism - I'd like to have the metadata be the mechanism and the pruning options be the policy, but I'm not entirely sure what that would look like yet.

docs/architecture/adr-001-flush-version.md
erikgrinaker: This was in response to your original comment

> we need to think about any edge cases where a client may halt and a subset of metadata was not flushed to disk

I don't think tm-db guarantee that batches are always transactional or atomic, as it depends on the backend. I believe some of them simply implement batches as a foreach.

docs/architecture/adr-001-flush-version.md
erikgrinaker: The above would be triggered by restarting Gaia with a different custom pruning strategy.

alexanderbez:
I moved `VersionStatus` to `Future Improvements`. Thanks for the review everyone :D

PR #251, opened 2020-05-26
Version Metadata

alexanderbez:
ref: https://github.com/tendermint/iavl/pull/250


nodedb.go
```diff
@@ -347,26 +347,28 @@ func (ndb *nodeDB) deleteOrphansHelper(db dbm.DB, batch dbm.Batch, flushToDisk b
 	}
 }
 
-func (ndb *nodeDB) PruneRecentVersions() (prunedVersions []int64, err error) {
+func (ndb *nodeDB) PruneRecentVersion() (int64, error) {
```
alexanderbez, line 350: Note we only return a single value. We never returned more than one so I'm not sure why we originally returned an array...

erikgrinaker: I'm just guessing, but he idea was perhaps to return which versions were pruned, but it seems like it was changed to only prune one version at a time.

go.mod
```diff
@@ -3,10 +3,13 @@ module github.com/tendermint/iavl
 go 1.13
 
 require (
+	github.com/golang/protobuf v1.4.0
```
erikgrinaker, line 6: We're also importing the `google.golang.org/protobuf` package with the v2 API, it might be better to standardize on that.

metadata.go
```diff
@@ -0,0 +1,23 @@
...
+}
+
+// Unmarshal attempts to decode a VersionMetadata object from Protobuf marshaled
+// bytes.
+func (vm *VersionMetadata) Unmarshal(bz []byte) error {
+	return proto.Unmarshal(bz, vm)
+}
+
+// VersionMetadataKey returns the indexing key for a VersionMetadata object.
+func VersionMetadataKey(version int64) []byte {
```
erikgrinaker, line 21: This shouldn't be exported.

metadata.go
```diff
@@ -0,0 +1,23 @@
...
+	return []byte(fmt.Sprintf("metadata/%d", version))
```
erikgrinaker, line 22: We should use a binary encoding which retains the natural ordering of versions and makes it possible to scan across them. `key_format.go` has tooling for this, I'd suggest prefix `m` or `v`.

mutable_tree.go
```diff
@@ -378,14 +379,29 @@ func (tree *MutableTree) LoadVersion(targetVersion int64) (int64, error) {
...
+	if err := tree.deleteVersionsFrom(targetVersion + 1); err != nil { // nolint
+		return latestVersion, err
+	}
+
+	vm, err := tree.ndb.GetVersionMetadata(targetVersion)
+	if err != nil {
+		return latestVersion, err
+	}
+
+	vm.Updated = time.Now().UTC().Unix()
```
erikgrinaker, line 400: I don't think we need to update this, since this version isn't changing in any way?

mutable_tree.go
```diff
@@ -442,80 +458,123 @@ func (tree *MutableTree) GetVersioned(key []byte, version int64) (
...
+
+		if err := tree.ndb.SaveOrphans(version, tree.orphans); err != nil {
+			panic(err)
+		}
+
+		if err := tree.ndb.SaveRoot(tree.root, version); err != nil {
 			panic(err)
 		}
-	}
-	err := tree.ndb.Commit()
```
erikgrinaker, line 485: Let's keep this commit, removing it may cause higher memory usage since the in-memory batches now have to hold both the queued changes and the pruning changes.

mutable_tree.go
```diff
@@ -442,80 +458,123 @@ func (tree *MutableTree) GetVersioned(key []byte, version int64) (
...
 	return tree.Hash(), version, nil
 }
 
+// PruneRecentVersion looks for a recent version to remove from the recentDB. If
+// such a version exists, the corresponding metadata will be updated as well.
+func (tree *MutableTree) PruneRecentVersion() error {
+	prunedVersion, err := tree.ndb.PruneRecentVersion()
+	if err != nil {
+		return err
+	} else if prunedVersion > 0 {
```
erikgrinaker, line 550: This should just be an `if`, since the above branch returns.

mutable_tree.go
```diff
@@ -442,80 +458,123 @@ func (tree *MutableTree) GetVersioned(key []byte, version int64) (
...
+	if err := tree.ndb.SetVersionMetadata(vm); err != nil {
+		return nil, version, err
+	}
+
 	return tree.Hash(), version, nil
 }
 
+// PruneRecentVersion looks for a recent version to remove from the recentDB. If
+// such a version exists, the corresponding metadata will be updated as well.
+func (tree *MutableTree) PruneRecentVersion() error {
```
erikgrinaker, line 546: I don't think we want this to be exported.

mutable_tree.go
```diff
@@ -442,80 +458,123 @@ func (tree *MutableTree) GetVersioned(key []byte, version int64) (
...
+
+		delete(tree.versions, prunedVersion)
+		return tree.ndb.Commit()
+	}
+
+	return nil
+}
+
 // DeleteVersion deletes a tree version from disk. The version can then no
 // longer be accessed.
```
erikgrinaker, line 569: Maybe add a comment that version metadata is retained.

mutable_tree.go
```diff
@@ -526,77 +585,105 @@ func (tree *MutableTree) DeleteVersion(version int64) error {
...
 		}
+
 		if version == lastestVersion {
 			root, err := tree.ndb.getRoot(version)
 			if err != nil {
 				return err
 			}
-			tree.deleteNodes(newLatestVersion, root)
+
+			if err := tree.deleteNodes(newLatestVersion, root); err != nil {
```
erikgrinaker, line 632: Thanks for adding these error checks!

nodedb.go
```diff
@@ -64,12 +72,82 @@ func newNodeDB(snapshotDB dbm.DB, recentDB dbm.DB, cacheSize int, opts *Options)
...
+	}
+
+	bz, err := ndb.snapshotDB.Get(key)
+	if err != nil {
+		return nil, err
+	}
+
+	// In cases where version metadata does not exist, say for a non-empty tree,
+	// we construct metadata on the fly. This is for backwards compatibility WRT
+	// to snapshotting.
```
erikgrinaker, line 99: Clever!

metadata.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+// Marshal returns Protobuf marshaled bytes of a VersionMetadata object.
+func (vm *VersionMetadata) Marshal() ([]byte, error) {
+	return proto.Marshal(vm)
+}
+
+// Unmarshal attempts to decode a VersionMetadata object from Protobuf marshaled
+// bytes.
+func (vm *VersionMetadata) Unmarshal(bz []byte) error {
```
erikgrinaker, line 16: Maybe this should be a constructor instead? It's a bit unnecessary to create an empty struct to unmarshal into.

go.mod
alexanderbez: We're using APIv2 -- `github.com/golang/protobuf` comes from the compiled proto code.

alexanderbez: I bumped my `protoc-gen-go` to `v1.4.2`.

metadata.go
alexanderbez: I didn't imagine we'd have a need to range of these -- I'm still not sure why we would, we don't even expose an API for this. But we can leverage the existing key formatting logic.

metadata.go
alexanderbez: It's now removed.

mutable_tree.go
alexanderbez: Indeed. I'll remove it.

mutable_tree.go
alexanderbez: The commit is kept. It happens in `MutableTree#PruneRecentVersion`.

go.mod
erikgrinaker: Interesting, wonder why they haven't migrated. Anyway, not a big deal I think, as long as we're using v2 ourselves.

mutable_tree.go
erikgrinaker: Got it, I misread the diff.

go.mod
alexanderbez: They have, I think it's used for different purposes. See https://github.com/golang/protobuf/releases/tag/v1.4.2

metadata.go
```diff
@@ -0,0 +1,18 @@
...
+	return proto.Marshal(vm)
+}
+
+// VersionMetadata attempts to decode a VersionMetadata object from Protobuf
+// marshaled bytes.
+func UnmarshalVersionMetadata(bz []byte) (*VersionMetadata, error) {
+	vm := &VersionMetadata{}
+	err := proto.Unmarshal(bz, vm)
+	return vm, err
+}
```
erikgrinaker, line 18: We may not want to expose these as part of our public API. Can we make them private?

alexanderbez: Sure, but there's no harm in keeping them public. A client has no way of interacting with the persistence of metadata.

erikgrinaker: Public APIs are subject to backwards compatibility guarantees, keeping them internal gives us more flexibility.

alexanderbez: This API would never change and is that a guarantee that we hold in the IAVL lib? We do this all the time in the SDK and Tendermint, where Tendermint is a worse offender because even point releases break APIs.

alexanderbez: Furthermore, what if an upstream client wants to serialize and store metadata for whatever reason?

erikgrinaker: Yeah, that's sort of the thing - if we make the marshaling public then we're committing to a certain encoding, making it harder to change if we ever decide to. If we feel like that's necessary then I'm not totally opposed to it, I just prefer keeping implementation details internal to have more freedom.

alexanderbez: Sure, I can make them private then. But note, if you decide to change the encoding, then you further complicate the Get/Set APIs on `nodeDB` because now you have to handle these cases.

erikgrinaker: Yup, but at least we have the option. I've been burnt way too many times on exposing too large of an API surface and not being able to put the genie back in the bottle - but maybe it's just PTSD, used to work on SaaS APIs which were set in stone since things would immediately break out in the world if we changed anything at all.

alexanderbez: Great points!

mutable_tree.go
```diff
@@ -526,77 +578,105 @@ func (tree *MutableTree) DeleteVersion(version int64) error {
...
+	}
 
+	delete(tree.versions, version)
 	return nil
 }
 
-// deleteVersionsFrom deletes tree version from disk specified version to latest version. The version can then no
-// longer be accessed.
+// deleteVersionsFrom deletes tree version from disk specified version to latest
+// version. The version can then no longer be accessed.
```
AdityaSripal, line 601: `VersionMetadata` is also deleted

types.proto
```diff
@@ -0,0 +1,22 @@
...
+// VersionMetadata defines the metadata associated with a committed IAVL version.
+message VersionMetadata {
+  // tree version for the corresponding metadata
+  int64 version = 1;
+
+  // the UNIX timestamp of when the metadata was committed to disk
+  int64 committed = 2;
+
+  // the UNIX timestamp of when the metadata was updated
+  int64 updated = 3;
```
AdityaSripal, line 15: Why call this `updated` when it is only different from `committed` IF the version gets deleted? This will always be true since VersionedTree is immutable.

Think it makes more sense to just call this `deleted` and only set it on deletion

alexanderbez: No, actually the `Updated` timestamp will also be updated during manual flushing as well. See the #239 PR.

PR #265, opened 2020-06-16
remove Amino

### LINKED ISSUE #242: Use a standard data encoding
We currently use Amino, we should use a standard serialization format instead, e.g. Protobuf. ~However, Protobuf does not support random access and always decodes e.g. whole nodes, which may be too expensive when we're simply traversing the tree. Other encodings may be more appropriate.~

erikgrinaker:
Removes Amino and adds varint utility-functions instead, keeping the same serialized format. Partial fix for #242.

The last vestige of Amino is in the proof handling, namely `AbsenceOp` and `ValueOp`, which are structs encoded/decoded with Amino.

### DIFF
--- a/docs/node/node.md
+++ b/docs/node/node.md
@@ -34,15 +34,15 @@ Every node is persisted by encoding the key, version, height, size and hash. If 
 ```golang
 // Writes the node as a serialized byte slice to the supplied io.Writer.
 func (node *Node) writeBytes(w io.Writer) error {
-	cause := amino.EncodeInt8(w, node.height)
+	cause := encodeVarint(w, node.height)
 	if cause != nil {
 		return errors.Wrap(cause, "writing height")
 	}
-	cause = amino.EncodeVarint(w, node.size)
+	cause = encodeVarint(w, node.size)
 	if cause != nil {
 		return errors.Wrap(cause, "writing size")
 	}
-	cause = amino.EncodeVarint(w, node.version)
+	cause = encodeVarint(w, node.version)
 	if cause != nil {
 		return errors.Wrap(cause, "writing version")
 	}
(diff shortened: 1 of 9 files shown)


AdityaSripal:
Amino can be removed from the proofs. This will affect the `IavlSpec` in ICS-23 since the prefix would change. Think it's worth it to purge iAVL of amino though.

Think #276 should be merged after this PR, and `IavlSpec` needs to be changed in https://github.com/confio/ics23

Please confirm @ethanfrey

ethanfrey:
As to ics23 proofs, we allow a spec to length-prefix data many ways. Ideally the output of [`binary.Varint`](https://github.com/cosmos/iavl/pull/265/files#diff-c1e0b73f3e74e3389ee4dab001672f51R33) is compatible with [one of the supported methods](https://github.com/confio/ics23/blob/master/proofs.proto#L15-L40). If so, we just update the spec along with the new proofs. If this is different, please make a PR to add this to ics23 (we need it in proto and 3 languages) or use a supported format, or wait til I have bandwidth for this.

ethanfrey:
It looks like `binary.Varint` [is protobuf compatible](https://golang.org/pkg/encoding/binary):

```
The varint functions encode and decode single integer values using a variable-length encoding; smaller values require fewer bytes. For a specification, see https://developers.google.com/protocol-buffers/docs/encoding. 
```

So this shouldn't even be an issue for IavlSpec (but please test)

ethanfrey:
I would also say that once #276 is merged and deeply integrated (in a separate PR?) to access the store rather than translating exisiting proofs, you may wish to remove the custom amino-style proof format. How are RangeProofs serialized? Maybe not in this module, but in the sdk they are only serialized using amino before hitting the wire on sdk queries.

erikgrinaker:
Following discussions in call last week with @AdityaSripal and @alexanderbez, we'll keep the current proofs along with the new proofs, but try to migrate them to Protobuf if we are able to keep byte-compatibility.

I'll do that in a separate PR, so this PR is good to review and merge.

PR #271, opened 2020-06-20
DeleteVersions

alexanderbez:
Implement `MutableTree#DeleteVersions`. This method will be used in the revised pruning strategy in the Cosmos SDK, where pruned versions will be deleted in "batches" as opposed to on a block-by-block basis.

@erikgrinaker feel free to keep this PR open while you tackle removing pruning from the IAVL as I know you'll also further optimize this method. Or you can also merge this and address optimizations later.

/cc @zmanian


alexanderbez:
That was just the result of a `go mod tidy` :)

PR #276, opened 2020-06-23
Port over ics23-iavl

AdityaSripal:
Currently a simple copy of https://github.com/confio/ics23-iavl into this repo. This will remove the outside dependency sitting between iavl and SDK.

A future PR will optimize this further by directly creating the ICS proof rather than converting

cc: @ethanfrey


ics23.go/test_helpers.go
```diff
@@ -0,0 +1,108 @@
...
+// GetKey this returns a key, on Left/Right/Middle
+func GetKey(allkeys [][]byte, loc Where) []byte {
+	if loc == Left {
+		return allkeys[0]
+	}
+	if loc == Right {
+		return allkeys[len(allkeys)-1]
+	}
+	// select a random index between 1 and allkeys-2
+	idx := rand.Int()%(len(allkeys)-2) + 1
```
fedekunze, line 70:
```
G404: Use of weak random number generator (math/rand instead of crypto/rand) (gosec)
```

ics23.go/proof.go
```diff
@@ -0,0 +1,167 @@
+package iavl
```
erikgrinaker, line 1: This probably shouldn't be named `iavl` when the root package is also called `iavl`. Let's either move this into the root, or rename it.

marbar3778:
since we are doing multiple releases of IAVL we could mark the current proofs as deprecated and add these then in the next major release remove the old ones. The sdk does not use them and other users would be using iavl the same way the sdk is so it shouldn't be a worry.

erikgrinaker:
> since we are doing multiple releases of IAVL we could mark the current proofs as deprecated and add these then in the next major release remove the old ones. The sdk does not use them and other users would be using iavl the same way the sdk is so it shouldn't be a worry.
> 
> Could you add deprecated flag to the old proofs godoc

That might be a good idea. If so, we should make sure the API is final, and make sure there are no bugs that require breaking changes to proofs. We should also update the proof documentation.

Could someone give me a quick summary of where IAVL proofs are used today?

marbar3778:
> Could someone give me a quick summary of where IAVL proofs are used today?

I couldn't find them being used in the sdk. I looked over a few projects that are forks of the sdk and they dont use them as well.. deadcode?

ethanfrey:
They are turned into ProofOps when you query `/store` with `proof: true` (query raw value by raw key). It can be validated by the ProofRuntime (which seems to be in 0.33 tendermint but gone in 0.34).

On the client side, you are supposed to parse and validate all this. Theoretically, there is a light-client mode in the cli (not sure how to turn it on) and this will validate all such queries. I was using them back in the original sdk :wink: 0.8 or so.... hope it didn't all get refactored away

ics23.go/proof.go
ethanfrey: Why not just place these files in the top level package?

Rename `ics23.go/proof.go` => `proof_ics23.go` or such
And `test_helpers.go` can be merged into `proof_ics23_test.go` (as it doesn't seem to be used otherwise.

2 files. Simple.

ics23.go/test_helpers.go
ethanfrey: Yeah, this is only for test code. It should probably be turned into a `_test.go` file.

ethanfrey:
> I think at some point this should be made methods on the tree itself. Are we planning on replacing the current proofs with these, or will these be in addition?

I agree. I kept it separate to avoid super slow PR review cycles last summer. Now the core team will use this code it definitely should use private methods and returns the proof directly (just like current range proofs). In fact, we could replace all proofs with this, and returns `CompressedBatchProofs` instead of range proofs (they allow you to return a whole lot of proofs, possibly continuous).

I would merge this first, then add them as first class citizens to the tree, then see if we can start deprecating what is there

ics23.go/proof.go
erikgrinaker: Yeah, let's move them to the root, then we can merge this and start integrating with the main tree.

erikgrinaker:
@AdityaSripal Could you move this into the root package and fix the linter issues please?

I'd also like to start discussing the plan for the final 0.15 proof API so that we can cut it soon.

proof_ics23.go
```diff
@@ -0,0 +1,166 @@
...
+			Hash:   ics23.HashOp_SHA256,
+			Prefix: prefix,
+			Suffix: suffix,
+		}
+		steps = append(steps, op)
+	}
+	return steps
+}
+
+func aminoVarInt(orig int64) []byte {
```
erikgrinaker, line 149: Is this really necessary? As far as I know, Amino uses standard Go variants, and we have utility functions for generating these in `encoding.go` - can you check if we can use any of those instead?

proof_ics23_test.go
```diff
@@ -0,0 +1,194 @@
...
+// returns a list of all keys in sorted order
+func BuildTree(size int) (itree *ImmutableTree, keys [][]byte, err error) {
+	tree, _ := NewMutableTree(db.NewMemDB(), 0)
+
+	// insert lots of info and store the bytes
+	keys = make([][]byte, size)
+	for i := 0; i < size; i++ {
+		// create random 4 byte key
+		// nolint:gosec
+		key := []byte{byte(rand.Uint64()), byte(rand.Uint64()), byte(rand.Uint64()), byte(rand.Uint64())}
```
erikgrinaker, line 184: You can use `rand.Read()` as well to generate random byte slices.

proof_ics23.go
```diff
@@ -0,0 +1,166 @@
...
+	"fmt"
+
+	ics23 "github.com/confio/ics23/go"
+)
+
+/*
+GetMembershipProof will produce a CommitmentProof that the given key (and queries value) exists in the iavl tree.
+If the key doesn't exist in the tree, this will return an error.
+*/
+func (tree *ImmutableTree) GetMembershipProof(key []byte) (*ics23.CommitmentProof, error) {
```
erikgrinaker, line 13: We'll have to use `t *ImmutableTree` to appease the linter.

erikgrinaker:
Thanks for updating! Added a couple of minor comments - can you update the changelog as well, and then merge this?

PR #287, opened 2020-07-01
proof: migrate ProofOp encoding to Protobuf

### LINKED ISSUE #242: Use a standard data encoding

erikgrinaker:
Fixes #242. Switches ProofOp encoding from Amino to Protobuf, but retains the same byte layout for backwards compatibility. This is not strictly Protobuf since the proofs are length-prefixed, but that's what Amino did so that's what we'll keep doing.

The Protobuf types need to be moved to a separate package and cleaned up: they conflict with the native types, and we're making the same mistake as before by leaking Protobuf types through public APIs. I'll submit a separate PR for this.


erikgrinaker:
> LGTM
> 
> would we want to break the encoding in the future?

I think we want to migrate to the ICS proofs, which presumably have their own binary encoding. Let's keep the encoding for the current proofs stable until they're removed entirely.

PR #292, opened 2020-07-03
proof: move Protobuf schemas to internal package

erikgrinaker:
Hides the Protobuf types from the public API and uses the same name for native and Protobuf types.


marbar3778:
should this be `.proto` files in `/proto` and then the script that generates the proto stubs just moves the generated code into `/internal`

erikgrinaker:
> should this be `.proto` files in `/proto` and then the script that generates the proto stubs just moves the generated code into `/internal`

Sure

PR #293, opened 2020-07-08
dep: update to latest master of TM

marbar3778:


erikgrinaker:
This doesn't seem right to me: why are these APIs now taking Protobuf messages?

marbar3778:
ProofOp/Proof was always a proto type, but now the location of the file has moved.

erikgrinaker:
Ok, that seems like a bit of a code smell to me, but sure. If we merge this we can't cut 0.15 until Tendermint 0.34 is out; we cool with that?

marbar3778:
that's how it has always been with this repo. The only way to avoid this is to remove tm dep on proofs but this has other consequences.

PR #296, opened 2020-07-22
gRPC API Redux

blinky3713:
This PR builds on the work of @alexanderbez and @charlescrain to implement the gRPC server interface via protobuf service descriptions. Currently the only real difference between this branch and #184 is the `RWMutex` used to control reads and writes. I am not a `golang` developer, but this seems pretty straight forward.

In addition to the tests in `server_test.go`, I have tested [hs-iavl-client](https://github.com/f-o-a-m/kepler/blob/update-iavl-client/hs-iavl-client/test/Database/IAVL/RPCSpec.hs) against a docker image build from this branch and (after updating the library with the correct proto files) it passes. If you would like more tests, please specify here.


blinky3713:
note that I was forced to copy files over from the other branch rather than try to rebase because the repositories had diverged so much that it not worth the effort

cmd/iavlserver/main.go
```diff
@@ -0,0 +1,191 @@
+package main
```
erikgrinaker, line 1: I keep getting these log messages on all HTTP requests:

```
Failed to extract ServerMetadata from context
```

cmd/iavlserver/main.go
```diff
@@ -0,0 +1,191 @@
...
+var (
+	dbDataDir       = flag.String("datadir", "", "The database data directory")
+	dbName          = flag.String("db-name", "", "The database name")
+	dbBackend       = flag.String("db-backend", string(dbm.GoLevelDBBackend), "The database backend")
+	version         = flag.Int64("version", 0, "The IAVL version to load")
+	cacheSize       = flag.Int64("cache-size", 10000, "Tree cache size")
+	gRPCEndpoint    = flag.String("grpc-endpoint", "localhost:8090", "The gRPC server endpoint (host:port)")
+	gatewayEndpoint = flag.String("gateway-endpoint", "localhost:8091", "The gRPC-Gateway server endpoint (host:port)")
+	cpuProfile      = flag.String("cpuprofile", "", "If set, write CPU profile to this file")
+	memProfile      = flag.String("memprofile", "", "If set, write memory profile to this file")
```
erikgrinaker, line 35: Maybe we could simply expose the pprof server via HTTP instead of using command-line arguments for profiling? This allows the caller to better control the profiling by submitting options.

cmd/iavlserver/README.md
```diff
@@ -0,0 +1,72 @@
...
+        The gRPC server endpoint (host:port) (default "localhost:8090")
+  -memprofile string
+        If set, write memory profile to this file
+  -no-gateway
+        Disables the gRPC-Gateway server
+  -version int
+        The IAVL version to load
+```
+
+### Example
```
erikgrinaker, line 48: It would be great to have a couple of get/set/save example calls here as well, and maybe an example of how to generate and use a gRPC client in a different language.

server/server_test.go
```diff
@@ -0,0 +1,677 @@
...
+
+	server, err := server.New(db, 1000, 0)
+	suite.NoError(err)
+
+	suite.server = server
+	suite.populateItems(100)
+}
+
+func (suite *ServerTestSuite) populateItems(n int) {
+	versionRes, err := suite.server.Version(context.TODO(), nil)
```
erikgrinaker, line 35: Use `context.Background()`, not `context.TODO()` - the latter is only meant to be used as a placeholder/marker while porting non-context-aware code to newer context APIs, and should not be used in finished code.

server/server.go
```diff
@@ -0,0 +1,301 @@
...
+
+var _ pb.IAVLServiceServer = (*IAVLServer)(nil)
+
+// IAVLServer implements the gRPC IAVLServiceServer interface. It provides a gRPC
+// API over an IAVL tree.
+// rwLock is used to ensure:
+// 1. No reading while writing, no writing while reading.
+// 2. Unlimited concurrent reads.
+// 3. Sequential writes.
+type IAVLServer struct {
```
erikgrinaker, line 25: The server does not appear to have any panic handlers installed, such that a panic causes the entire server to crash. This is particularly problematic since IAVL uses panics as a primary error mechanism. We should use some sort of middleware which catches the panics and converts them into error responses.

Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM golang:1.12 as build
```
erikgrinaker, line 1: We can bump this to 1.14.

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# Now add the local iavl repo, which typically isn't cacheable.
+COPY . .
+
+# Build the server.
+RUN go get ./cmd/iavlserver
+
+# Make a minimal image.
+FROM gcr.io/distroless/base
+
+COPY --from=build /go/bin/iavlserver /
```
erikgrinaker, line 22: This should have a default entrypoint or command. The typical way to do this is to use an entrypoint with the base binary, and a default command giving some basic parameters for a transient database. We should also expose the port numbers.

Makefile
```diff
@@ -1,4 +1,3 @@
-GOTOOLS := github.com/golangci/golangci-lint/cmd/golangci-lint
 VERSION := $(shell echo $(shell git describe --tags) | sed 's/^v//')
```
erikgrinaker, line 1: `GOTOOLS` shouldn't be removed, I think?

proof_iavl_absence.go
```diff
@@ -53,16 +53,16 @@ func AbsenceOpDecoder(pop tmmerkle.ProofOp) (merkle.ProofOperator, error) {
 	if err != nil {
 		return nil, err
 	}
-	proof, err := rangeProofFromProto(pbProofOp.Proof)
+	proof, err := RangeProofFromProto(pbProofOp.Proof)
```
erikgrinaker, line 56: Maybe these should be moved into an internal package or the `proto` package or something, don't think we want them to become part of the public API.

cmd/iavlserver/main.go
```diff
@@ -0,0 +1,191 @@
...
+
+	case *dbBackend == "":
+		return nil, errors.New("database backend cannot be empty")
+
+	case *dbDataDir == "":
+		return nil, errors.New("database datadir cannot be empty")
+	}
+
+	defer func() {
+		if r := recover(); r != nil {
```
erikgrinaker, line 155: This is no longer necessary, since `NewDB()` now returns an error instead.

proto/iavl_api.proto
```diff
@@ -0,0 +1,252 @@
...
+// gRPC service
+// ----------------------------------------------------------------------------
+
+// IAVLService defines the gRPC service API contract for the IAVL tree.
+service IAVLService {
+  // Has returns a result containing a boolean on whether or not the IAVL tree
+  // has a given key at a specific tree version.
+  rpc Has(HasRequest) returns (HasResponse) {
+    option (google.api.http) = {
+      get: "/v1/tree/{version}/has"
```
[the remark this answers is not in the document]
blinky3713: No idea, I think @alexanderbez named the methods. I can remove it or replace it with anything you want

proto/iavl_api.proto
```diff
@@ -0,0 +1,252 @@
...
+import "third_party/google/api/annotations.proto";
+import "google/protobuf/empty.proto";
+import "proto/proof.proto";
+
+// ----------------------------------------------------------------------------
+// gRPC service
+// ----------------------------------------------------------------------------
+
+// IAVLService defines the gRPC service API contract for the IAVL tree.
+service IAVLService {
```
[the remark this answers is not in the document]
blinky3713: The api was written in like November of last year and could be out of date / was never complete. Can you either enumerate the missing methods here or tell me which module I can use to compare this service definition for incompleteness ?

Dockerfile
blinky3713: I've been running with something like

```
> sudo docker run --name iavl -p 8090:8090 -p 8091:8091 -d martin/iavl:latest /iavlserver  -db-name "test" -datadir "."  -grpc-endpoint "0.0.0.0:8090"  -gateway-endpoint "0.0.0.0:8091"

```

I'm not like a Dockerfile power user, but are you saying you want to use environment variables instead of this kind of invocation to pass the cli args and flags?

proto/iavl_api.proto
```diff
@@ -0,0 +1,252 @@
+syntax = "proto3";
+package iavl;
+
+option go_package = "proto/iavl";
```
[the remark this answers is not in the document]
blinky3713: I'm not super familiar with the combination of `go` plugins for `protoc` that glue everything, together with the bash script. I was kind of just stumbling around until I got something to work -- it was only non-trivial because this package imports something from `proof.proto`. If I can still give `proof.proto` and `iavl_api.proto` the same `go_package`, I don't care what it's called. Otherwise I would need someone to explicitly tell me the file structure, package names that you want.

On a side note, there are too many `*proto*` packages/libraries/plugins in go :exploding_head:

proof_iavl_absence.go
blinky3713: what about the `utils`? module?

erikgrinaker: That will wind up in the base package (all files in a directory make up a package). I'd rather not have this litter the base package public API, see e.g. the `Blue` type in the API reference for why I think this is bad:

https://pkg.go.dev/github.com/tendermint/iavl?tab=doc#pkg-index

proto/iavl_api.proto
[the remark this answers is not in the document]
erikgrinaker: I think we should move all of the Protobuf schemas into the same file, and use `proto` as the Go package name.

>On a side note, there are too many *proto* packages/libraries/plugins in go 🤯

Tell me about it, we've wasted a significant amount of time dealing with the fallout of this lately.

proto/iavl_api.proto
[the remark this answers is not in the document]
erikgrinaker: I'd like to remove it, but will let @alexanderbez weigh in.

proto/iavl_api.proto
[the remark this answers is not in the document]
erikgrinaker: API reference here: https://pkg.go.dev/github.com/tendermint/iavl?tab=doc#pkg-index

We don't need to implement _everything_, but should at least implement stuff that users would reasonably expect to be included. Personally, I would expect at least these:

* `Iterate` and `IterateRange` (which should probably just be a streaming `List` method)
* `AvailableVersions`
* `Load`, `LoadVersion`, and `LoadVersionForOverwriting` (can probably be combined)
* `Version`
* `Size`
* `GetMembershipProof`
* `GetNonMembershipProof`

In addition, we can consider `Export` and `Import`.

Dockerfile
erikgrinaker: No, I'm saying we should start `/iavlserver` with a transient database by default:

```Dockerfile
EXPOSE 8090 8091
ENTRYPOINT ["/iavlserver"]
COMMAND ["-db-name", "iavl", "-datadir", "."]
```

Users can override the parameters like so:

```sh
$ docker run -d iavl:latest -db-name "foo"
```

proto/iavl_api.proto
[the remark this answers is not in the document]
clevinson: Not sure how helpful this is, but for reference of how we organize proto packages in Cosmos SDK, you can see [this ADR](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-023-protobuf-naming.md) which documents our guidelines for proto package & naming conventions.

and here's the PR that tackled this for the SDK - https://github.com/cosmos/cosmos-sdk/pull/6413 

Multiple files can indeed have the same `go_package`'s. I'm not too familiar w/ how IAVL is already structuring its proto stuff though, and I'd default to @erikgrinaker's suggestions so that we have internal consistency with how IAVL is already doing proto stuff for now.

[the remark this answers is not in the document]
erikgrinaker: Following the SDK's guidelines makes sense; we only have a handful of internal Protobuf messages currently.

proto/iavl_api.proto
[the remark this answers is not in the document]
blinky3713: `GetMembershipProof` `GetNonMembershipProof` require types from the `ics23` package that are not represented in `proof.proto` and there are no conversion functions to write them in terms of anything that appears in `proof.proto` (although you seem to have things for the other direction). 

Maybe someone who works on `iavl` take care of this before making it a requirement for this PR?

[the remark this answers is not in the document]
blinky3713: `Iterate` and `IterateRange` require the caller to provide a function for deciding rather to continue or terminate the traversal. The options are:

1. Leave these methods out.
2. Use a constant function evaluating to `false`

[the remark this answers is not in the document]
erikgrinaker: Ok, let's drop the ICS proofs for now. For Iterate, just use `List` instead with options from/to (i.e. use a nook function). Listing keys/values is a very common operation.

proof_iavl_absence.go
erikgrinaker: Good point, I guess we can make these part of the public API then - i.e. leave them as is.

blinky3713:
Ok so I think I have done everything we've discussed up to now. In the end the methods added were:

- `List`
- `AvailableVersions`
- `Load`, `LoadVersion`, and `LoadVersionForOverwriting`
- `Version`
- `Size`


I haven't written tests yet for any of the new methods, I was thinking I would write them for each one. However if there any additional tests requested, now would be the time to make sure.

erikgrinaker:
Thanks @blinky3713, that's awesome! Would you like me to do another pass over this now, or wait until all unresolved comments have been addressed?

As for tests, it's always desirable with more tests, but since this is just a thin wrapper it should be sufficient as a start to just check that all gRPC and HTTP methods work and that all parameters are properly passed through.

blinky3713:
@erikgrinaker I still have some of the smaller initial comments to address, I will ping you when it's ready for final review.

server/server.go
blinky3713: copy-pasted the solution from http://www.inanzzz.com/index.php/post/oicw/panic-recovery-middleware-for-golang-http-server

cmd/iavlserver/README.md
blinky3713: I will just link to Kepler's `hs-iavl-package` for an example of how to do this in haskell.

cmd/iavlserver/main.go
blinky3713: I think that's the equivalent of a 404, I don't know how to make it more precise but I only notice that when sending a message to an endpoint that doesn't exist.

blinky3713:
@erikgrinaker I think it's ready for a final review, I don't understand this conflicting `go.sum` error but it's causing problems with CI.

I didn't include tests for `Load`, `LoadVersion` and `LoadVersionForOverwrite` because they were just direct wrappers.

My test for `List` could probably be improved but I'm open to advice.

I built a docker image and tested against `hs-iavl-client` ([this branch](https://github.com/f-o-a-m/kepler/tree/update-iavl-client)), everything seems to work still. However, when I try to test a simple set/get using curl for the `README`, i'm failing to save values. for example:

```
> curl --location --request GET 'http://localhost:8091/v1/version' --header 'Content-Type: application/json' \
--data-raw '{
    "key": "aGVsbG8=",
    "value": "Ynll"
}'

{
  "result": false
}

> curl --location --request GET 'http://localhost:8091/v1/get' \
--header 'Content-Type: application/json' \
--data-raw '{
    "key": "aGVsbG8=",
}'

{
  "code": 5,
  "message": "the key requested does not exist",
  "details": []
}
```

Is there something I'm missing about how keys have to be formatted?

blinky3713:
also as a side note, the linters are all broken:
- for the `proto-lint` and `proto-breakage` jobs it's unclear to me how to include the `third_party` directory.
- for the `golangci-lint` job, I don't see what file it's complaining about. if I install version `1.30.0` on my machine and run the command I get a completely different output:

```bash
(base)  ✘ ✝  code/golang/iavl   martin/grpc  golangci-lint run --out-format=github-actions --timeout 10m
::error file=benchmarks/bench_test.go,line=70,col=13::G404: Use of weak random number generator (math/rand instead of crypto/rand) (gosec)
::error file=benchmarks/bench_test.go,line=89,col=15::G404: Use of weak random number generator (math/rand instead of crypto/rand) (gosec)
::error file=benchmarks/bench_test.go,line=128,col=16::G404: Use of weak random number generator (math/rand instead of crypto/rand) (gosec)

```

erikgrinaker:
Sorry for not getting back to you here @blinky3713 - this weekend is pretty busy, but I'll have a look on Monday for sure.

proto/iavl_api.proto
```diff
@@ -0,0 +1,336 @@
...
+
+message LoadVersionRequest {
+  int64 version = 1;
+}
+
+message LoadVersionForOverwritingRequest {
+  int64 version = 1;
+}
+
+message ListElementsRequest {
```
erikgrinaker, line 268: Shouldn't this be called `ListRequest`, for consistency with the method name?

proto/iavl_api.proto
```diff
@@ -0,0 +1,336 @@
...
+  bytes from_key = 1;
+  bytes to_key = 2;
+  bool ascending = 3;
```
erikgrinaker, line 271: Maybe we want to use `descending` instead, so that the default order when not specified is ascending. Or use an enum with the default set to ascending.

proto/iavl_api.proto
```diff
@@ -0,0 +1,336 @@
...
+  bool result = 1;
+}
+
+message GetResponse {
+  int64 index = 1;
+  bytes value = 2;
+}
+
+message SetResponse {
+  bool result = 1;
```
erikgrinaker, line 289: This should probably be called `updated`, it's somewhat unintuitive to set a new value and get `result:false` back.

go.mod
```diff
@@ -4,14 +4,20 @@ go 1.13
 
 require (
 	github.com/confio/ics23/go v0.0.0-20200610201322-18c7bd6b2dd3
+	github.com/gogo/gateway v1.1.0
 	github.com/gogo/protobuf v1.3.1
+	github.com/golang/protobuf v1.4.2
+	github.com/golangci/golangci-lint v1.29.0 // indirect
```
erikgrinaker, line 10: This should't be here, as it's not a code dependency of IAVL.

go.mod
```diff
@@ -4,14 +4,20 @@ go 1.13
...
+	github.com/grpc-ecosystem/grpc-gateway v1.14.6
 	github.com/niemeyer/pretty v0.0.0-20200227124842-a10e7caefd8e // indirect
 	github.com/pkg/errors v0.9.1
 	github.com/stretchr/testify v1.6.1
 	github.com/tendermint/tendermint v0.34.0-rc2
 	github.com/tendermint/tm-db v0.6.0
 	golang.org/x/crypto v0.0.0-20200622213623-75b288015ac9
-	google.golang.org/protobuf v1.24.0 // indirect
+	google.golang.org/genproto v0.0.0-20200526211855-cb27e3aa2013
```
erikgrinaker, line 18: This probably shouldn't be here either.

erikgrinaker:
> However, when I try to test a simple set/get using curl for the README, i'm failing to save values.

`GET` requests can't contain a request body, parameters must be specified as URL parameters - this works:

```
$ curl -XPOST http://localhost:8091/v1/set -d '{"key": "'$(echo -n foo | base64)'", "value": "'$(echo -n bar | base64)'"}'
{
  "result": false
}

$ curl "http://localhost:8091/v1/get?key=$(echo -n foo | base64)"
{
  "index": "0",
  "value": "YmFy"
}
```

erikgrinaker:
>for the proto-lint and proto-breakage jobs it's unclear to me how to include the third_party directory.

Add them to `build.roots` in `buf.yaml`.

> for the golangci-lint job, I don't see what file it's complaining about

You have to expand the command output in the Github lint job output (not very intuitive, I agree). It's complaining about this:

```
server/server_test.go:39:2: directive `//nolint` is unused (nolintlint)
	//nolint
```

Just remove the `nolint` comment.

erikgrinaker:
>I don't understand this conflicting go.sum error but it's causing problems with CI

Remove some of the `go.mod` packages that shouldn't be there (see review comments), then run `go mod tidy` to clean up the `go.mod` and `go.sum` files. That should do it.

blinky3713:
I have made all of the requested changes and updated the example in the README. I'm still getting a conflicting error in `go.sum` after running `go mod tidy`. I've never worked with a language where checking in these kinds of files is advised, so I'm not really sure what to do about it.

This should be the last thing so just tell me what else I can do and hopefully that will be it.

blinky3713:
@erikgrinaker any advice on this last little step ?

erikgrinaker:
Sorry again for the late reply, I was on vacation last week.

> I'm still getting a conflicting error in go.sum after running go mod tidy.

Ah. Yes, this is a Git conflict. You'll need to merge `master` into your branch, or rebase your branch. In order to fix the `go.sum` conflicts, just run `go mod tidy` and regenerate the correct file instead of trying to resolve manually.

blinky3713:
:tada:

erikgrinaker (approved):
Cool cool, let's merge this thing! Thanks for the PR, you really went above and beyond here! 🚀

PR #304, opened 2020-08-07
return hash for empty trees in Hash() and SaveVersion()

erikgrinaker:
Related to https://github.com/tendermint/tendermint/issues/5192, for RFC-6962 conformance.

### DIFF
--- a/node.go
+++ b/node.go
@@ -218,7 +218,12 @@ func (node *Node) _hash() []byte {
 
 // Hash the node and its descendants recursively. This usually mutates all
 // descendant nodes. Returns the node hash and number of nodes hashed.
+// If the tree is empty (i.e. the node is nil), returns the hash of an empty input,
+// to conform with RFC-6962.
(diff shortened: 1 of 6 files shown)


node.go
```diff
@@ -218,7 +218,12 @@ func (node *Node) _hash() []byte {
 
 // Hash the node and its descendants recursively. This usually mutates all
 // descendant nodes. Returns the node hash and number of nodes hashed.
+// If the tree is empty (i.e. the node is nil), returns the hash of an empty input,
+// to conform with RFC-6962.
 func (node *Node) hashWithCount() ([]byte, int64) {
+	if node == nil {
+		return sha256.New().Sum(nil), 0
```
alexanderbez, line 225: Should this be `nil` or an empty string?

erikgrinaker: You mean the hash input? `nil` and `[]byte{}` are equivalent.

PR #310, opened 2020-08-25
changelog: add entry for gRPC server

erikgrinaker:


marbar3778 (approved):
wow this was a long time coming

PR #312, opened 2020-08-26
add MutableTree.SetInitialVersion()

erikgrinaker:
Requested in https://github.com/cosmos/cosmos-sdk/pull/7089#issuecomment-680877202.

I didn't add a `GetInitialVersion()`, since it's not clear whether this should return only the option or the _actual_ initial version that exists in a tree.

It might be cleaner to add `GetOptions()`, `SetOptions()`, and `UpdateOptions(func (o Options) Options)`, but we may not want all options to be mutable after construction.

CC @amaurymartiny

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 - [\#296](https://github.com/cosmos/iavl/pull/296) Add `iavlserver`, a gRPC/REST API server.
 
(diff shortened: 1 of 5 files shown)


amaurymartiny (approved):
lgtm, I believe this will unblock my PR on the cosmos-sdk 👍

erikgrinaker:
I'll need an approval from a grown-up, I'm afraid - the repo is set up like that. @alexanderbez, want to do the honors?

PR #318, opened 2020-09-30
fix: constant overflow when compiling for 32bit machines

troian:
on Go 1.x(x) all constants are ints by default which causing
build for 32bit machines to fail due to overflow
https://github.com/golang/go/issues/23086


erikgrinaker:
Thanks! Can you add an entry to `CHANGELOG.md` as well please, under "Unreleased"?

PR #324, opened 2020-11-04
Add an alternative to DeleteVersions that leaves no garbage

klim0v:
The beginning of this issue is that the DeleteVersions method leaves behind a lot of dirty information, which leads to an increase in storage used by the storage.

I first added a method to remove versions up to a certain height, and now I have expanded it to allow removal within a given range.


nodedb.go
```diff
@@ -237,6 +237,53 @@ func (ndb *nodeDB) DeleteVersionsFrom(version int64) error {
...
+	}
+
+	ndb.mtx.Lock()
+	defer ndb.mtx.Unlock()
+
+	// Delete orphans
+	ndb.traverseOrphans(func(key, hash []byte) {
+		var fromVersion, toVersion int64
+		orphanKeyFormat.Scan(key, &toVersion, &fromVersion)
+		if toVersion < version {
```
klim0v, line 272:
```suggestion
		if fromVersion < version {
```
Is it possible to do this?
It seems to me that this is not logically correct, but the tests pass. Apparently, this is a bottleneck for me, could you explain to me?

tree_test.go
```diff
@@ -1324,14 +1324,107 @@ func TestLoadVersionForOverwriting(t *testing.T) {
...
-	//Now any key value can be written into the tree
+	// The tree version now is 52 which is equal to latest version.
+	// Now any key value can be written into the tree
 	tree.Set([]byte("key any value"), []byte("value any value"))
 	_, _, err = tree.SaveVersion()
 	require.NoError(err, "SaveVersion should not fail.")
 }
 
-//////////////////////////// BENCHMARKS ///////////////////////////////////////
+func TestDeleteVersionsCompare(t *testing.T) {
```
klim0v, line 1334: For comparison, you can see my TestDeleteVersionsCompare, which I wrote for debugging, and maybe it should be deleted already. Here is its result:
```
=== RUN   TestDeleteVersionsCompare
    tree_test.go:1365: DeleteVersionsFromInterval database.size 1614
    tree_test.go:1395: DeleteVersion database.size 1614
    tree_test.go:1423: DeleteVersions database.size 1618
--- PASS: TestDeleteVersionsCompare (0.04s)
```

erikgrinaker:
Thanks for contributing! I see you're still making changes to this, and that there a linter issues. Let me know when it's ready for review!

klim0v:
> Thanks for contributing! I see you're still making changes to this, and that there are a few linter issues. Let me know when it's ready for review! Also, maybe convert this to a draft until it's ready.

I see linter crashing in CI / CD config, updated to 1.28 and added `nosec` tag to ignore
_Error: Failed to run: Error: requested golangci-lint version 'v1.27' isn't supported: we support only v1.28.3 and later versions, Error: requested golangci-lint version 'v1.27' isn't supported: we support only v1.28.3 and later versions_

I finished making changes, you can proceed to check! @erikgrinaker

klim0v:
Finished!

klim0v:
@erikgrinaker, I added an alternative, can you check both of them? https://github.com/klim0v/iavl/compare/6933e534d28136c59f00896079d5ea4c952e5097...69e21305a688313e99ff0b2f51379186c497645c

klim0v:
I also suggest to mark the DeleteVersions method as deprecated and try to pass sorted versions from it to the DeleteVersionsInterval.

klim0v:
@alexanderbez, how do you like it?

erikgrinaker:
I've backported a bunch of linter fixes from `master`, so you can rebase to avoid all of the linter stuff.

I'd be interested to find out exactly what this garbage you see left behind by `DeleteVersions` actually is. We have a test `TestRandomOperations` that runs randomized operations against a tree and check invariants, including calling `DeleteVersion` on all versions in random order, and in #326 I've added a test to make sure there are no stray database entries left behind at the end. It doesn't leave any garbage.

Since `DeleteVersions` basically just iterates over all versions and calls `DeleteVersion`, this should be equivalent to what this test does, and it doesn't produce any garbage in the database. Can you elaborate?

klim0v:
> Since `DeleteVersions` basically just iterates over all versions and calls `DeleteVersion`, this should be equivalent to what this test does, and it doesn't produce any garbage in the database. Can you elaborate?

Operation DeleteVersion and DeleteVersions differ in that the commit does not occur on every delete. This is the problem.
Pay attention to the method [deleteOrphans](https://github.com/klim0v/iavl/blob/82251d865e4733c7337f2862d459dae8087fb3cc/nodedb.go#L341) `ndb.getPreviousVersion(version)` will return the values of the previous node that may have been removed in the previous iteration, since it takes this data directly from the persistent node store. As a result, the logic of using this field below is broken in the function passed to the traverseOrphansVersion method

erikgrinaker (changes requested):
Ah, good catch! This should really be implemented using ACID database transactions, to avoid deviations between batched changes and the database state, but that's a larger issue.

This looks great, from what I can tell. Some minor requests:

* Let's rename `DeleteVersionsInterval` to `DeleteVersionsRange`, since we use range as a term for this elsewhere in the code.

* It would be great if you could update `DeleteVersions` to use `DeleteVersionsInterval`.

* It would also be great to add random `DeleteVersionsInterval` calls to `TestRandomOperations`, e.g. by first checking whether a version should be deleted, then randomly choosing to delete either a single or multiple versions, and if multiple delete a random range of e.g. 2-5 versions (which may or may not have gaps).

* The `#nosec` comments are unnecessary now and can be removed.

* `RangeProof.ToProto` and `RangeProofFromProto` are erroneously removed.

klim0v:
> RangeProof.ToProto and RangeProofFromProto are erroneously removed.

the latest version that you uploaded does not find import for `iavlproto`

erikgrinaker:
>>RangeProof.ToProto and RangeProofFromProto are erroneously removed.

>the latest version that you uploaded does not find import for iavlproto

Sorry, my bad - cherry-pick gone wrong. I've reverted this change on `0.14.x` now.

klim0v:
@erikgrinaker, could you help me that I found another subtle place. This happens when you delete blocks again. In the case when I previously deleted data in the old way using DeleteVersions. I got some trash in there. And I clean it up with my new _DeleteVersionsRange_ implementation. But when I call _DeleteVersionsRange_ again at the same interval, I see a tenfold increase in execution time, particularly in the _ndb.traverseOrphansVersion_ location. Why do you think this might be happening? Unfortunately, I can't check the _tree.versions [version]_, because it won't allow me to delete the old garbage. What would you suggest to do?

klim0v:
@erikgrinaker 
why don't we use `&util.Range{Start: start, Limit: end}` in [(db *GoLevelDB) Iterator](https://github.com/tendermint/tm-db/blob/bdf7336cc676c0b7bb518db7e9f247d3a87d7759/goleveldb.go#L161)?

erikgrinaker:
>why don't we use &util.Range{Start: start, Limit: end} in (db *GoLevelDB) Iterator?

Right, this is going to give pretty poor performance. I'm not sure why it's done like this - I know we've had to do some weird things with iterators to get the tm-db semantics to map onto the semantics of the underlying database, but I _think_ in this case it should work.

We have fairly good test coverage for iterators, want to just try using `Range` and see if the tests pass?

klim0v:
> We have fairly good test coverage for iterators, want to just try using `Range` and see if the tests pass?

Yes, all tests pass, including I ran tests from iavl, replacing the database in them with leveldb

erikgrinaker:
Ok, great - did it help with the performance though? I suspect using `Range` and calling `Seek()` are basically the same thing, thus it shouldn't have an effect on performance, but would be good to verify.

klim0v:
> Ok, great - did it help with the performance though? I suspect using `Range` and calling `Seek()` are basically the same thing, thus it shouldn't have an effect on performance, but would be good to verify.

Yes, it helped! Now I'm looking for an explanation of why this happened.

klim0v:
I suppose Seek() could be slow due to the need to go through the entire tree to make sure there is no such key.

erikgrinaker:
Great! It may be that `Seek()` actually iterates. Either way, want to submit a PR for tm-db?

klim0v:
> Great! It may be that `Seek()` actually iterates. Either way, want to submit a PR for tm-db?

Yes, but so far I cannot confirm my words with performance tests to describe this problem...

tree_random_test.go
```diff
@@ -148,12 +149,23 @@ func testRandomOperations(t *testing.T, randSeed int64) {
...
 			versions := getMirrorVersions(diskMirrors, memMirrors)
 			if len(versions) > 2 {
-				deleteVersion := int64(versions[r.Intn(len(versions)-1)])
-				t.Logf("Deleting version %v", deleteVersion)
-				err = tree.DeleteVersion(deleteVersion)
-				require.NoError(t, err)
-				delete(diskMirrors, deleteVersion)
-				delete(memMirrors, deleteVersion)
+				i := r.Intn(len(versions) - 1)
+				if i > versions[0] {
```
erikgrinaker, line 153: I don't understand this logic. `i` here is the index of a random element in the array, why are we comparing that with a version number?

What we should do here, is to randomly pick whether to delete a single or multiple versions (by `deleteRangeChange`), and if multiple the randomly pick a small range to delete (e.g. 2-5 versions).

erikgrinaker (commented):
This PR includes a bunch of noisy formatting changes, it would be nice to avoid these since they're unrelated to the main change.

Also, I've released new versions of tm-db and updated the 0.15 and 0.14 branches of IAVL.

klim0v:
> This PR includes a bunch of noisy formatting changes, it would be nice to avoid these since they're unrelated to the main change.
> 
> Also, I've released new versions of tm-db and updated the 0.15 and 0.14 branches of IAVL.

Done ✅

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+### Bug Fixes
+
+- [\#324](https://github.com/cosmos/iavl/pull/324) Fix `DeleteVersions` causing a memory leak. 
+  Add `DeleteVersionsRange` to delete in range (@klim0v)
```
erikgrinaker, line 8:
```suggestion
- [\#324](https://github.com/cosmos/iavl/pull/324) Fix `DeleteVersions` not properly removing
  orphans, and add `DeleteVersionsRange` to delete a range (@klim0v)
```

erikgrinaker (approved):
Great stuff, thank you for this! Suggested a few minor nitpicks, otherwise LGTM!

klim0v:
I have added your recommendations, thanks!

PR #326, opened 2020-11-06
test: check stray database entries in TestRandomOperations

erikgrinaker:
Checks that we don't leave behind any junk in the database, as reported in #324.


marbar3778 (approved):
utACK

PR #332, opened 2020-11-19
docker: ci

### LINKED ISSUE #311: docker: deploy to docker hub
Deploy releases of `iavlserver` to docker hub.

Should we do on pushes to master as well?

marbar3778:
- push docker images

closes #311

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,6 +1,6 @@
 name: Build & Push
-#  Build & Push rebuilds the tendermint docker image on every push to master and creation of tags
-# and pushes the image to https://hub.docker.com/r/interchainio/simapp/tags
+#  Build & Push rebuilds the iavl docker image on every push to master and creation of tags
+# and pushes the image to https://hub.docker.com/r/interchainio/iavl/tags
 on:
   pull_request:
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -0,0 +1,54 @@
...
+#  Build & Push rebuilds the iavl docker image on every push to master and creation of tags
+# and pushes the image to https://hub.docker.com/r/interchainio/iavl/tags
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+" # Push events to matching v*, i.e. v1.0, v20.15.10
+      - "v[0-9]+.[0-9]+.[0-9]+-rc*" # Push events to matching v*, i.e. v1.0-rc1, v20.15.10-rc5
```
alexanderbez, line 11: I think these will only match versions with major.minor.patch, not for example, major.minor. Correct me if I'm wrong. Why not just `v[0-9]+.[0-9]*.?[0-9]*.*` or even more relaxed `v[0-9]+.*`?

marbar3778:
```
v[12].[0-9]+.[0-9]+ | Matches all semantic versioning tags with major version 1 or 2
```

Went with this approach from github actions. We could relax it more

erikgrinaker (approved):
Looks great, thanks for this!

PR #337, opened 2020-11-23
github: increase test timeout to 10 minutes 

erikgrinaker:
See failure in #333.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: test & coverage report creation
         run: |
-          go test ./... -mod=readonly -timeout 8m -race -coverprofile=coverage.txt -covermode=atomic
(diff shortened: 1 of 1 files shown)


marbar3778 (approved):
LGTM. 

I saw the error was a data race is this caused by the timeout?

erikgrinaker:
Looks like it. There's one backtrace to the coverage thing, and one hidden one (possibly some internal Go thing). I'm guessing something went wrong while handling the timeout panic.

marbar3778:
It is super odd that this test is running for so long though, is this due to a new change?

erikgrinaker:
`TestRandomOperations` sets up a bunch of trees and runs random operations against them. When the race detector is enabled, this takes forever.

I'm looking into running them in parallel now, although it may make it harder to read the output since the different tests' output will be jumbled together.

erikgrinaker:
Seems to work well enough, #338. Suggest we just do that instead, and drop this timeout bump.

erikgrinaker:
Superseded by #338.

PR #338, opened 2020-11-23
test: run slow tests without race detector in CI

erikgrinaker:


marbar3778 (approved):

erikgrinaker:
Well, that didn't really help much. Locally I'm seeing 70-90% reductions in test time, but then I also have an 8-core machine. GitHub actions uses 2 cores, I think.

Will try using default parallelism (2) in case the additional jobs just cause more overhead.

tree_random_test.go
```diff
@@ -28,9 +30,12 @@ func TestRandomOperations(t *testing.T) {
 		957382170,
 	}
 
-	for _, seed := range seeds {
-		seed := seed
+	for i, seed := range seeds {
+		i, seed := i, seed
 		t.Run(fmt.Sprintf("Seed %v", seed), func(t *testing.T) {
+			if testing.Short() && i >= 2 {
```
marbar3778, line 36: 😄

erikgrinaker: Hm?

marbar3778: Just a smile for going with shortening the amount of test cases.

marbar3778 (approved):
looks good to me. Thank you for taking the time to make this go faster.

erikgrinaker:
Superseded by #338.

PR #340, opened 2020-12-12
fix decodeBytes() integer overflow on 32-bit systems

### LINKED ISSUE #339: decodeBytes is susceptible to an overflow bug sneak and can incur an OOM or runtime panic due to unproperly checked size on both 32-bit and 64-bit architectures
Reading through the code in encoding.go, we've got this interesting pattern of code https://github.com/cosmos/iavl/blob/f5c104d6591ea3ec0c4933f5a00c530632d471a5/encoding.go#L15-L25

In that code block, you'll notice that:
a) `size, n, err := decodeUvarint(bz []byte) (uint64, int, error)` --> size is a uint64 containing the size aka length of the buffer to be allocated
b) To check for overflows, we use int(size): which checks if uint64->int returns a negative value
c) Unfortunately: `make([]byte, size)` uses the plain uint64 value unfettered and unchecked

Point b) masks a problem with overflows; Point c) is where the problem is, and to cause this disruption, I can Uvarint encode any size larger than 32-bit ranges, the overflow will wrap around to it and given the fallacious check yet naked transmission to make, it'll panic with out of memory

## Implications
Given that int's range is (-2147483648 to 2147483647) on 32-bit systems, and (-9223372036854775808 to 9223372036854775807) on 64-bit systems, on 32-bit systems, any value with a wrap around of greater than (1<<32), e.g. (int32(uint64(1<<32)+2)) == 2; will return that value so to falsely make all those checks pass and then poison the code with an OOM or runtime panic, I can do this
```go
package iavl

import (
        "encoding/binary"
        "testing"
)       
        
func TestDecodeBytesBlow(t *testing.T) {
        // Attack works on a 32-bit computer, and causes an OOM.
        buf := make([]byte, 1<<63-1)
        _ = binary.MaxVarintLen64
        n := binary.PutUvarint(buf, (1<<33)+2)
        // n:    ~> 5
        // size: ~> 2, given that on a 32-bit computer: int(uint64(1<<33)+2) == 2 :-(
        nx := copy(buf[n:], "abcde")
        println("n+nx", n+nx)
        if _, _, err := decodeBytes(buf); err == nil {
                t.Fatal("Failed to decode bytes")
        }
}
```

## Recommendation
For robust checks we can do the following:
a) Ensure that size never exceeds the max value of int on either of 32-bit or 64-bit architecures
b) Ensure that make receives the same truncation used, so size is passed in as int(size) -- in fact just the lone change would have masked the problem
```diff
diff --git a/encoding.go b/encoding.go
index 27c10fe..40c9d73 100644
--- a/encoding.go
+++ b/encoding.go
@@ -19,10 +19,14 @@ func decodeBytes(bz []byte) ([]byte, int, error) {
 	if int(size) < 0 {
 		return nil, n, fmt.Errorf("invalid negative length %v decoding []byte", size)
 	}
+	// ^uint(0) >> 1 will help determine the max int value variably on 32-bit and 64-bit machines.
+	if size >= uint64(^uint(0)>>1) {
+		return nil, n, fmt.Errorf("invalid out of range length %v decoding []byte", size)
+	}
 	if len(bz) < n+int(size) {
 		return nil, n, fmt.Errorf("insufficient bytes decoding []byte of length %v", size)
 	}
-	bz2 := make([]byte, size)
+	bz2 := make([]byte, int(size))
 	copy(bz2, bz[n:n+int(size)])
 	n += int(size)
 	return bz2, n, nil
```

I believe this is a high value DOS vector that can take down various components

erikgrinaker:
Fixes #339.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Bug Fixes
+
+- [\#340](https://github.com/cosmos/iavl/pull/340) Fix integer overflow in `decodeBytes()` that can cause out-of-memory errors on 32-bit machines for certain inputs.
+
 ## 0.15.0 (November 23, 2020)
 
 The IAVL project has moved from https://github.com/tendermint/iavl to
--- a/encoding.go
+++ b/encoding.go
@@ -12,19 +12,21 @@ import (
 // decodeBytes decodes a varint length-prefixed byte slice, returning it along with the number
 // of input bytes read.
 func decodeBytes(bz []byte) ([]byte, int, error) {
-	size, n, err := decodeUvarint(bz)
+	s, n, err := decodeUvarint(bz)
 	if err != nil {
 		return nil, n, err
 	}
-	if int(size) < 0 {
-		return nil, n, fmt.Errorf("invalid negative length %v decoding []byte", size)
+	// ^uint(0) >> 1 will help determine the max int value variably on 32-bit and 64-bit machines.
+	if uint64(n)+s >= uint64(^uint(0)>>1) {
+		return nil, n, fmt.Errorf("invalid out of range length %v decoding []byte", uint64(n)+s)
 	}
-	if len(bz) < n+int(size) {
+	size := int(s)
+	if len(bz) < n+size {
 		return nil, n, fmt.Errorf("insufficient bytes decoding []byte of length %v", size)
 	}
 	bz2 := make([]byte, size)
-	copy(bz2, bz[n:n+int(size)])
-	n += int(size)
+	copy(bz2, bz[n:n+size])
+	n += size
 	return bz2, n, nil
 }
 


marbar3778 (approved):

odeke-em (approved):
LGTM, thank you @erikgrinaker! Minor nit: if we could adapt the test in the repro would be nice. Also when this is approved, could you please cut a new release, and let's mail an update for all dependencies for cosmos/iavl? Thank you.

erikgrinaker:
> Minor nit: if we could adapt the test in the repro would be nice.

I don't have access to a 32-bit system at the moment, we'll look into this next week.

> could you please cut a new release, and let's mail an update for all dependencies for cosmos/iavl?

I'll cut a new release and ping the SDK people, but don't feel like we need to announce this more broadly as 32-bit is fairly niche and supported on a best-effort basis.

odeke-em:
Thank you for the fix @erikgrinaker, and thank you for the review @marbar3778!

>I'll cut a new release and ping the SDK people, but don't feel like we need to announce this more broadly as 32-bit is fairly niche and supported on a best-effort basis.

As I commented in https://github.com/cosmos/iavl/issues/339#issuecomment-744057334, I just re-ran some thoughts and also fuzzed and alas identified a vector that crashes on 64-bit systems, as well with the same runtime out of range error when the length is saved as `9223372036854775806` and it cleverly circumvents the checks given that the code
```go
 if len(bz) < n+int(size) { 
 	return nil, n, fmt.Errorf("insufficient bytes decoding []byte of length %v", size) 
 } 
```
n+int(9223372036854775806) on 64-bit systems --> -X
e.g. if n = 9, we get `size 9223372036854775806 int(size) 9223372036854775806 n 9 len(bz) 9 -9223372036854775801`.

This PR fixes also the 32-bit problems as well as for 64-bit systems, thus I recommend that we should perhaps issue updates to code using this package ASAP.

PR #348, opened 2020-12-14
check version<0 in deleteVersion

### LINKED ISSUE #345: MutableTree.deleteVersion should use version <= 0 instead of  version == 0 to catch values <= 0
The code in https://github.com/cosmos/iavl/blob/c5ad3d8fbc6d2c25329685f328462332016c8e2d/mutable_tree.go#L519-L521 intends to catch values <= 0, and the error message intends it, but the condition doesn't reflect it and the code flows all the way to checking in tree.versions[version], for which ErrVersionDoesNotExist then gets returned. However, it shouldn't have gotten to that spot in the first place.

erikgrinaker:
Fixes #345. This would cause an error in any case, since the version does not exist, but it's better to catch it here.

### DIFF
--- a/mutable_tree.go
+++ b/mutable_tree.go
@@ -516,7 +516,7 @@ func (tree *MutableTree) SaveVersion() ([]byte, int64, error) {
 }
 
 func (tree *MutableTree) deleteVersion(version int64) error {
-	if version == 0 {
+	if version <= 0 {
 		return errors.New("version must be greater than 0")
 	}
 	if version == tree.version {


mutable_tree.go
```diff
@@ -516,7 +516,7 @@ func (tree *MutableTree) SaveVersion() ([]byte, int64, error) {
 }
 
 func (tree *MutableTree) deleteVersion(version int64) error {
-	if version == 0 {
+	if version <= 0 {
```
marbar3778, line 519: Is there a tradeoff between this and migrating to uin64? Out of scope for the PR but curious.

marbar3778 (approved):

mutable_tree.go
erikgrinaker: `uint64` is a breaking change, since we use `int64` in the public APIs.

PR #352, opened 2020-12-15
convertVarIntToBytes: use reusable bytes array not expensive, discarded bytes.Buffer

### LINKED ISSUE #344: convertVarIntToBytes is heavy handed, burns at least 48Bytes, CPU cycles and causes convertLeafOp to run in microseconds instead of nanoseconds, and could take in a byte array
Noticed a bunch of code that invokes convertVarIntToByteswhose internals and seems to be heavily used in the code base, stemming from (*ImmutableTree).GetNonMembership. Looking at some hits
```shell
$ git grep -n 'convertVarIntToBytes('
proof_ics23.go:98:      prefix := convertVarIntToBytes(0)
proof_ics23.go:99:      prefix = append(prefix, convertVarIntToBytes(1)...)
proof_ics23.go:100:     prefix = append(prefix, convertVarIntToBytes(version)...)
proof_ics23.go:121:             prefix := convertVarIntToBytes(int64(path[i].Height))
proof_ics23.go:122:             prefix = append(prefix, convertVarIntToBytes(path[i].Size)...)
proof_ics23.go:123:             prefix = append(prefix, convertVarIntToBytes(path[i].Version)...)
proof_ics23.go:150:func convertVarIntToBytes(orig int64) []byte {
```
https://github.com/cosmos/iavl/blob/903e53d2f5309e685dd54f5ba42ab564cbbc1993/proof_ics23.go#L150-L158

Notice in there that for every invocation, we create a bytes.Buffer, use it and then discard it :-), that's at least 48Bytes gone, as well as garbage collection penalty. It is heavily use by convertLeafOp and others.

## Remedy
Given the successive hits identified above e.g. if proof_ics23.go 
https://github.com/cosmos/iavl/blob/903e53d2f5309e685dd54f5ba42ab564cbbc1993/proof_ics23.go#L121-L123

which is 48 * 3 = 144 bytes at least wasted

and same for https://github.com/cosmos/iavl/blob/903e53d2f5309e685dd54f5ba42ab564cbbc1993/proof_ics23.go#L98-L100

I think we can definitely do better. The functions reuse code but almost pedantically to the point of just a varint encoding needing to be written to a writer then back to a slice. We can fix this by perhaps reusing a byte array and updating the signature for convertVarintToBytes to take a byte array that's allocated once, and then reused in all those call sites. We'll pass in a byte array and not a byte slice to ensure no need for bounds checking. Perhaps this diff
```diff
diff --git a/proof_ics23.go b/proof_ics23.go
index 293270d..43c36df 100644
--- a/proof_ics23.go
+++ b/proof_ics23.go
@@ -1,7 +1,7 @@
 package iavl
 
 import (
-	"bytes"
+	"encoding/binary"
 	"fmt"
 
 	ics23 "github.com/confio/ics23/go"
@@ -95,9 +95,10 @@ func convertExistenceProof(p *RangeProof, key, value []byte) (*ics23.ExistencePr
 
 func convertLeafOp(version int64) *ics23.LeafOp {
 	// this is adapted from iavl/proof.go:proofLeafNode.Hash()
-	prefix := convertVarIntToBytes(0)
-	prefix = append(prefix, convertVarIntToBytes(1)...)
-	prefix = append(prefix, convertVarIntToBytes(version)...)
+	var varintBuf [binary.MaxVarintLen64]byte
+	prefix := convertVarIntToBytes(0, varintBuf)
+	prefix = append(prefix, convertVarIntToBytes(1, varintBuf)...)
+	prefix = append(prefix, convertVarIntToBytes(version, varintBuf)...)
 
 	return &ics23.LeafOp{
 		Hash:         ics23.HashOp_SHA256,
@@ -109,6 +110,7 @@ func convertLeafOp(version int64) *ics23.LeafOp {
 
 // we cannot get the proofInnerNode type, so we need to do the whole path in one function
 func convertInnerOps(path PathToLeaf) []*ics23.InnerOp {
+	var varintBuf [binary.MaxVarintLen64]byte
 	steps := make([]*ics23.InnerOp, 0, len(path))
 
 	// lengthByte is the length prefix prepended to each of the sha256 sub-hashes
@@ -118,9 +120,9 @@ func convertInnerOps(path PathToLeaf) []*ics23.InnerOp {
 	// we want to go up from the leaf to the root
 	for i := len(path) - 1; i >= 0; i-- {
 		// this is adapted from iavl/proof.go:proofInnerNode.Hash()
-		prefix := convertVarIntToBytes(int64(path[i].Height))
-		prefix = append(prefix, convertVarIntToBytes(path[i].Size)...)
-		prefix = append(prefix, convertVarIntToBytes(path[i].Version)...)
+		prefix := convertVarIntToBytes(int64(path[i].Height), varintBuf)
+		prefix = append(prefix, convertVarIntToBytes(path[i].Size, varintBuf)...)
+		prefix = append(prefix, convertVarIntToBytes(path[i].Version, varintBuf)...)
 
 		var suffix []byte
 		if len(path[i].Left) > 0 {
@@ -147,12 +149,7 @@ func convertInnerOps(path PathToLeaf) []*ics23.InnerOp {
 	return steps
 }
 
-func convertVarIntToBytes(orig int64) []byte {
-	buf := new(bytes.Buffer)
-	err := encodeVarint(buf, orig)
-	// write should not fail
-	if err != nil {
-		panic(err)
-	}
-	return buf.Bytes()
+func convertVarIntToBytes(orig int64, buf [binary.MaxVarintLen64]byte) []byte {
+	n := binary.PutVarint(buf[:], orig)
+	return buf[:n]
 }
```

## Results
Given this benchmark
```go
package iavl

import (
        "testing"
)       

var values = []struct {
        n    int64
        want int
}{      
        {0, 1},
        {1, 1},
        {100, 2},
        {127, 2},
        {128, 2},
        {1 << 29, 5},
        {-0, 1},
        {-1, 1},
        {-100, 2},
        {-127, 2},
        {-128, 2},
        {-1 << 29, 5},  
}
        
var sink interface{}
                
func BenchmarkConvertLeafOp(b *testing.B) { 
        b.ReportAllocs()
        for i := 0; i < b.N; i++ {
                for _, tt := range values {
                        sink = convertLeafOp(tt.n)
                }
        }
        if sink == nil {
                b.Fatal("Benchmark wasn't run")
        }               
        sink = nil
}
```
The results are dramatic, the code can now run in nanoseconds instead of microseconds, ~80% faster, with ~>=80% less allocations
```shell
name             old time/op    new time/op    delta
ConvertLeafOp-8    4.08µs ± 1%    0.83µs ± 0%  -79.59%  (p=0.000 n=9+9)

name             old alloc/op   new alloc/op   delta
ConvertLeafOp-8    5.18kB ± 0%    0.77kB ± 0%  -85.19%  (p=0.000 n=10+10)

name             old allocs/op  new allocs/op  delta
ConvertLeafOp-8       120 ± 0%        24 ± 0%  -80.00%  (p=0.000 n=10+10)
```

Kindly cc-ing @alessio @erikgrinaker @marbar3778

odeke-em:
Noticed while auditing and profiling dependencies of cosmos-sdk, that
convertVarIntToBytes, while reusing already implemented code, it
was expensively creating a bytes.Buffer (40B on 64-bit architectures)
returning a result and discarding it, yet that code was called
3 times successively at least.

By reusing a byte array (not a slice, to ensure bounds checks
eliminations by the compiler), we are able to dramatically improve
performance, taking it from ~4µs down to 850ns (~4.5X reduction),
reduce allocations by >=~80% in every dimension:

```shell
name             old time/op    new time/op    delta
ConvertLeafOp-8    3.90µs ± 1%    0.85µs ± 4%  -78.12%  (p=0.000 n=10+10)

name             old alloc/op   new alloc/op   delta
ConvertLeafOp-8    5.18kB ± 0%    0.77kB ± 0%  -85.19%  (p=0.000 n=10+10)

name             old allocs/op  new allocs/op  delta
ConvertLeafOp-8       120 ± 0%        24 ± 0%  -80.00%  (p=0.000 n=10+10)
```

Fixes #344

### DIFF
--- a/encoding_test.go
+++ b/encoding_test.go
@@ -86,3 +86,37 @@ func TestDecodeBytes_invalidVarint(t *testing.T) {
 	_, _, err := decodeBytes([]byte{0xff})
 	require.Error(t, err)
 }
+
+// sink is kept as a global to ensure that value checks and assignments to it can't be
+// optimized away, and this will help us ensure that benchmarks successfully run.
+var sink interface{}
+
+func BenchmarkConvertLeafOp(b *testing.B) {
+	var versions = []int64{
+		0,
+		1,
+		100,
+		127,
+		128,
+		1 << 29,
+		-0,
+		-1,
+		-100,
+		-127,
+		-128,
+		-1 << 29,
+	}
+
+	b.ReportAllocs()
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		for _, version := range versions {
+			sink = convertLeafOp(version)
+		}
+	}
+	if sink == nil {
+		b.Fatal("Benchmark wasn't run")
+	}
+	sink = nil
+}
--- a/proof_ics23.go
+++ b/proof_ics23.go
@@ -1,7 +1,7 @@
 package iavl
 
 import (
-	"bytes"
+	"encoding/binary"
 	"fmt"
 
 	ics23 "github.com/confio/ics23/go"
@@ -94,10 +94,11 @@ func convertExistenceProof(p *RangeProof, key, value []byte) (*ics23.ExistencePr
 }
 
 func convertLeafOp(version int64) *ics23.LeafOp {
+	var varintBuf [binary.MaxVarintLen64]byte
 	// this is adapted from iavl/proof.go:proofLeafNode.Hash()
-	prefix := convertVarIntToBytes(0)
-	prefix = append(prefix, convertVarIntToBytes(1)...)
-	prefix = append(prefix, convertVarIntToBytes(version)...)
+	prefix := convertVarIntToBytes(0, varintBuf)
+	prefix = append(prefix, convertVarIntToBytes(1, varintBuf)...)
+	prefix = append(prefix, convertVarIntToBytes(version, varintBuf)...)
 
 	return &ics23.LeafOp{
 		Hash:         ics23.HashOp_SHA256,
@@ -114,13 +115,15 @@ func convertInnerOps(path PathToLeaf) []*ics23.InnerOp {
 	// lengthByte is the length prefix prepended to each of the sha256 sub-hashes
 	var lengthByte byte = 0x20
 
+	var varintBuf [binary.MaxVarintLen64]byte
+
 	// we need to go in reverse order, iavl starts from root to leaf,
 	// we want to go up from the leaf to the root
 	for i := len(path) - 1; i >= 0; i-- {
 		// this is adapted from iavl/proof.go:proofInnerNode.Hash()
-		prefix := convertVarIntToBytes(int64(path[i].Height))
-		prefix = append(prefix, convertVarIntToBytes(path[i].Size)...)
-		prefix = append(prefix, convertVarIntToBytes(path[i].Version)...)
+		prefix := convertVarIntToBytes(int64(path[i].Height), varintBuf)
+		prefix = append(prefix, convertVarIntToBytes(path[i].Size, varintBuf)...)
+		prefix = append(prefix, convertVarIntToBytes(path[i].Version, varintBuf)...)
 
 		var suffix []byte
 		if len(path[i].Left) > 0 {
@@ -147,12 +150,7 @@ func convertInnerOps(path PathToLeaf) []*ics23.InnerOp {
 	return steps
 }
 
-func convertVarIntToBytes(orig int64) []byte {
-	buf := new(bytes.Buffer)
-	err := encodeVarint(buf, orig)
-	// write should not fail
-	if err != nil {
-		panic(err)
-	}
-	return buf.Bytes()
+func convertVarIntToBytes(orig int64, buf [binary.MaxVarintLen64]byte) []byte {
+	n := binary.PutVarint(buf[:], orig)
+	return buf[:n]
 }


encoding_test.go
```diff
@@ -86,3 +86,33 @@ func TestDecodeBytes_invalidVarint(t *testing.T) {
...
+	1 << 29,
+	-0,
+	-1,
+	-100,
+	-127,
+	-128,
+	-1 << 29,
+}
+
+var sink interface{}
```
erikgrinaker, line 105: We should move these inside the benchmark function scope.

erikgrinaker (approved):
Thanks for the PR!

encoding_test.go
odeke-em: @erikgrinaker versions yes, we can move inside, but sink we have to leave as a global to ensure the benchmark runs successfully. I've added a comment to sink to indicate why, and then moved versions inside. Please take another look, and thank you for the review!

odeke-em:
@erikgrinaker please take a look, I’ve addressed your comments. Thank you!

erikgrinaker:
Thanks @odeke-em!

odeke-em:
Thank you @erikgrinaker for the review! Could you please help me cut a new release too? I ask because it would be nice for us to get the cosmos-sdk team to use the latest code as well as the changes you made over the weekend fixing identified issues.

erikgrinaker:
We cut 0.15.2 on Monday, the only change which isn't included is this one.

odeke-em:
Gotcha, thanks for the update. If you are able to, I kindly request that you help me cut a point release with this code included, as when I (memory and cpu) profile gaiad, iavl code shows up, and am investigating.

erikgrinaker:
Sure https://github.com/cosmos/iavl/releases/tag/v0.15.3

odeke-em:
Thank you @erikgrinaker!

Issue #354, opened 2020-12-26
GRPC support for GetByIndex


jlandrews:
For my use case (in Haskell via [Kepler](https://github.com/f-o-a-m/kepler)), it would be useful to be able to get the nearest populated key/value for a given search key. This is possible in IAVL with Get returning the index of the next populated key, and then using GetByIndex, but the GRPC layer doesn't have this functionality. The Get method returns an error when the requested key is not found, and there is no GetByIndex method.

I think this could be relatively easily addressed by changing the behavior of the GRPC Get method to better match that of ImmutableTree.Get, and by adding a GRPC GetByIndex method which just directly maps to ImmutableTree.GetByIndex.

erikgrinaker:
This sounds reasonable to me, feel like submitting a PR?

jlandrews:
:+1: #355

jlandrews:
Closed with merge of #355

PR #355, opened 2021-01-04
grpc/proto changes for GetByIndex

jlandrews:
#354 

Is there any documentation on how to do the setup for the gogoproto generation stuff? I see the `proto-gen` make action, but when I tried to set things up manually to do that I ended up with a different version of `protoc-gen-grpc-gateway` which caused a bunch of extra changes on the generated go files.


marbar3778:
There doesn't seem to be. Ill open a PR to fix that now.

server/server.go
```diff
@@ -79,13 +79,30 @@ func (s *IAVLServer) Get(_ context.Context, req *pb.GetRequest) (*pb.GetResponse
 
 	idx, value := s.tree.Get(req.Key)
 	if value == nil {
-		s := status.New(codes.NotFound, "the key requested does not exist")
-		return nil, s.Err()
+		e := status.New(codes.NotFound, "the key requested does not exist")
+		return &pb.GetResponse{Index: idx, Value: nil}, e.Err()
```
erikgrinaker, line 83: I don't believe Protobuf differentiates between `[]byte{}` and `nil`, so there's no way for a client to determine whether the key was missing or empty. There's [experimental support](https://github.com/protocolbuffers/protobuf/blob/v3.12.0/docs/field_presence.md) for field presence starting with Protobuf 3.12, but we may not want to rely on that just yet.

Maybe we just add a boolean `missing` field or something for non-existing keys?

erikgrinaker (commented):
Thanks! This is a breaking change, so it'll have to wait for 0.16.

server/server.go
jlandrews: I was letting the error serve as that indicator. Was I wrong to assume that the error code/message would be transmitted across gRPC along with the protobuf response? 

If so I can add in a missing field, and set the error to nil in both cases.

erikgrinaker: Oh, hm. I don't really know how gRPC handles error codes and response messages, or how the Go implementation does it. Worth trying it out (i.e. writing a test for it).

jlandrews: Okay, after some investigation, it is sending the error codes/message over the wire. But using that on the other side is a lot messier than using a value in the protobuf response. So I agree that it makes more sense to add a `Missing` field.

erikgrinaker (approved):
LGTM, thanks for the contribution! Would you mind adding an entry to `CHANGELOG.md` as well, in particular describing the breaking change to `Get()`?

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+
+## Unreleased
+
+### Breaking Changes
+- [\#355](https://github.com/cosmos/iavl/pull/355) `Get` in `iavlServer` no longer returns an error if the requested key does not exist. `GetResponse` now contains a `missing` boolean to indicate that a key does not exist, and the returned index will be that of the next occupied key.
```
robert-zaremba, line 7: Why do we need that breaking change?

erikgrinaker: I don't believe there is a supported way of returning structured error information in gRPC, so in order to return the index of the nearest key when accessing a missing key we have to change from an error response to a regular response. See original issue #354.

server/server.go
```diff
@@ -70,20 +70,37 @@ func (s *IAVLServer) Has(_ context.Context, req *pb.HasRequest) (*pb.HasResponse
...
 
 	idx, value := s.tree.Get(req.Key)
 	if value == nil {
-		s := status.New(codes.NotFound, "the key requested does not exist")
-		return nil, s.Err()
+		return &pb.GetResponse{Index: idx, Value: nil, Missing: true}, nil
 	}
 
-	return &pb.GetResponse{Index: idx, Value: value}, nil
+	return &pb.GetResponse{Index: idx, Value: value, Missing: false}, nil
```
robert-zaremba, line 86: you could merge the lines above to one line;
```
return &pb.GetResponse{Index: idx, Value: value, Missing: value==nil}, nil
```

proto/iavl/iavl_api.proto
```diff
@@ -283,6 +296,12 @@ message HasResponse {
 message GetResponse {
   int64 index = 1;
   bytes value = 2;
+  bool missing = 3;
```
robert-zaremba, line 299: how about changing to more standard `not_found` (like in HTTP) or `exists` (like in SQL)?

robert-zaremba (commented):
Let's consider renaming `missing`.

proto/iavl/iavl_api.proto
erikgrinaker: Sure, `not_found` sounds good. I think it makes sense for the zero value to default to the typical case where the key exists.

robert-zaremba (approved):
ukACK

PR #356, opened 2021-01-05
proto: use buf gen with docker

marbar3778:
- update docs and make file to use buf.gen.yaml with a docker file. This avoids the need to download protoc and various compilers


erikgrinaker (approved):
Great stuff, thanks!

PR #363, opened 2021-02-02
document MutableTree#Set return value

melekes:
While using the API, I assumed the returned value indicates whenever `Set` operation was successful or not. I was pointed out later that `Set` returns `true` if the value was updated, `false` if it is new.


alexanderbez (approved):
🎉

robert-zaremba (approved):
looks good, thanks.

Issue #373, opened 2021-03-05
Too large genesis file committed to panic with out of memory


KamiD:
## Summary
When the blockchain which based on cosmos-sdk exported a very large `genesis.json` file (over 5GB) , import and execute in new height with this huge block, it panicked cause ”out of memory“ when committing into IAVL tree.

## Environment
* Machine: 16G memory, 4 cores CPU
* OS: ubuntu 18.04

## Suggestion 
* Don't using levelDB batch to commit genesis block

PR #374, opened 2021-03-05
fix large genesis file commit

KamiD:
Closes: #373


nodedb.go
```diff
@@ -168,12 +169,31 @@ func (ndb *nodeDB) SaveBranch(node *Node) []byte {
...
+	var err error
+	if ndb.opts.Sync {
+		err = ndb.batch.WriteSync()
+	} else {
+		err = ndb.batch.Write()
+	}
+	if err != nil {
+		panic(err)
+	}
+	ndb.batch.Close()
```
robert-zaremba, line 193: Are we sure we don't loos data in an `async` mode here?

robert-zaremba (commented):
Thanks for contribution.
Could you add a test for reset sync in an async mode please?

robert-zaremba:
Hi @KamiD , we are going to tag a new release. Could you respond on the comment above so we can merge this PR?

nodedb.go
```diff
@@ -168,12 +169,31 @@ func (ndb *nodeDB) SaveBranch(node *Node) []byte {
 	node._hash()
 	ndb.SaveNode(node)
 
+	//resetBatch only working on generate a genesis block
+	if node.version == genesisVersion {
```
robert-zaremba, line 173:
```suggestion
	if node.version <= genesisVersion {
```

robert-zaremba (approved):
tACK

PR #385, opened 2021-04-26
fix: Make GetVersioned() available after doing LazyLoadVersion

egonspace:
I recreate a PR again in place of https://github.com/cosmos/iavl/pull/358.

We can use LazyLoadVersion by #148.
But if we start app with LazyLoadVersion we cannot query app state by specific height because MutableTree.GetVersioned returns nil for other previous version after doing LazyLoadVersion.

Our nodes are facing a slow booting up problem because of the large app data, and I want to hear your opinion about whether we should modify the code like this.

### DIFF
--- a/mutable_tree.go
+++ b/mutable_tree.go
@@ -27,6 +27,7 @@ type MutableTree struct {
 	lastSaved      *ImmutableTree   // The most recently saved tree.
 	orphans        map[string]int64 // Nodes removed by changes to working tree.
 	versions       map[int64]bool   // The previous, saved versions of the tree.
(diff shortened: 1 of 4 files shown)


robert-zaremba (approved):
Thanks. 
utACK.

egonspace:
Test ci is not running. Is there anything else I need to do?

robert-zaremba:
> Test ci is not running. Is there anything else I need to do?

I've just re-run the tests and will merge it right after.

robert-zaremba:
we will do a new release after merging this

PR #396, opened 2021-05-18
add "prefix" arg to iaviewer

summerpro:
The pr can support access to the data in the iavl tree of different modules of cosmos-sdk

In cosmos-sdk, the different modules under x/ use different iavl trees and are stored in the same leveldb (application.db). Different db prefixes are used to distinguish different modules, such as "s/k: gov/", "s/k:staking/", "s/k:evm/".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 
 ## Unreleased
-
+- [\#396](https://github.com/cosmos/iavl/pull/396) Add "prefix" arg to iaviewer.
 
 
 ## 0.16.0 (May 04, 2021)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 
 ## Unreleased
+
+### CLI Breaking Changes
+
 - [\#396](https://github.com/cosmos/iavl/pull/396) Add "prefix" arg to iaviewer.
 
 
--- a/cmd/iaviewer/main.go
+++ b/cmd/iaviewer/main.go
@@ -109,7 +109,7 @@ func ReadTree(dir string, version int, prefix []byte) (*iavl.MutableTree, error)
 	if err != nil {
 		return nil, err
 	}
-	if len(prefix) != nil {
+	if len(prefix) != 0 {
 		db = dbm.NewPrefixDB(db, prefix)
 	}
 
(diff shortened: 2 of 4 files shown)


cmd/iaviewer/main.go
```diff
@@ -101,11 +103,15 @@ func PrintDBStats(db dbm.DB) {
 
 // ReadTree loads an iavl tree from the directory
 // If version is 0, load latest, otherwise, load named version
-func ReadTree(dir string, version int) (*iavl.MutableTree, error) {
+func ReadTree(dir string, version int, prefix []byte) (*iavl.MutableTree, error) {
```
robert-zaremba, line 106: Please update the function comment.
The `iaviwer` will always set a prefix.

cmd/iaviewer/main.go
```diff
@@ -101,11 +103,15 @@ func PrintDBStats(db dbm.DB) {
...
 	db, err := OpenDB(dir)
 	if err != nil {
 		return nil, err
 	}
+	if prefix != nil {
```
robert-zaremba, line 111:
```suggestion
	if len(prefix) != 0 {
```

robert-zaremba: this will also catch a case when prefix is an empty string.

robert-zaremba (commented):
Thanks for contribution. Left a suggestion and we need to update the function godoc as well.

cmd/iaviewer/main.go
summerpro: good case

cmd/iaviewer/main.go
summerpro: ok

robert-zaremba (approved):
Please update the CHANGELOG as well.

summerpro:
> Please update the CHANGELOG as well.

done

summerpro:
hi @robert-zaremba , can this pr be merged?

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 
 ## Unreleased
-
+- [\#396](https://github.com/cosmos/iavl/pull/396) Add "prefix" arg to iaviewer.
```
robert-zaremba, line 5:
```suggestion

### CLI Breaking Changes

- [\#396](https://github.com/cosmos/iavl/pull/396) Add "prefix" arg to iaviewer.
```

robert-zaremba (approved):
Thanks. This is a good UX improvement.

PR #420, opened 2021-08-17
Reduce encodeVarint memory usage

### LINKED ISSUE #419: Optimize encodeVarint
Profiling full benchmarks, I notice `encodeVarint` shows up in top list of memory:

```
Showing nodes accounting for 247.53GB, 89.82% of 275.58GB total
Dropped 242 nodes (cum <= 1.38GB)
Showing top 20 nodes out of 73
      flat  flat%   sum%        cum   cum%
   36.26GB 13.16% 13.16%    36.26GB 13.16%  github.com/cosmos/iavl.(*KeyFormat).KeyBytes
   34.76GB 12.61% 25.77%    34.76GB 12.61%  github.com/cosmos/iavl.makeHashes
   23.04GB  8.36% 34.13%    23.04GB  8.36%  bytes.makeSlice
   22.93GB  8.32% 42.45%    22.93GB  8.32%  github.com/cosmos/iavl.convertLeafOp
   15.16GB  5.50% 47.95%    15.16GB  5.50%  github.com/cosmos/iavl.(*MutableTree).prepareOrphansSlice
   13.21GB  4.79% 52.74%    24.46GB  8.87%  github.com/cosmos/iavl.encodeVarint
```

For each call, `encodeVarint` allocate a new array for temporary store encoded data:

https://github.com/cosmos/iavl/blob/fb52f6c6a77b113e931509c6a1da0297be63d450/encoding.go#L111

We can use a sync.Pool to reduce allocation needed.

cuonglm:
By using a sync.Pool to shared temporary buffer between calls.

```
name                             old time/op    new time/op    delta
Export-8                           7.88ms ± 2%    7.91ms ± 1%     ~     (p=0.356 n=9+10)
Import-8                           9.01ms ± 7%    8.88ms ± 1%     ~     (p=0.529 n=10+10)
Node_WriteBytes/NoPreAllocate-8     340ns ± 5%     235ns ± 5%  -30.80%  (p=0.000 n=10+10)
Node_WriteBytes/PreAllocate-8       274ns ± 5%     204ns ± 0%  -25.49%  (p=0.000 n=10+8)

name                             old alloc/op   new alloc/op   delta
Export-8                           3.74MB ± 0%    3.74MB ± 0%     ~     (p=0.796 n=10+10)
Import-8                           9.84MB ± 0%    9.05MB ± 0%   -7.98%  (p=0.000 n=10+10)
Node_WriteBytes/NoPreAllocate-8      368B ± 0%      320B ± 0%  -13.04%  (p=0.000 n=10+10)
Node_WriteBytes/PreAllocate-8        224B ± 0%      176B ± 0%  -21.43%  (p=0.000 n=10+10)

name                             old allocs/op  new allocs/op  delta
Export-8                            69.6k ± 0%     69.6k ± 0%     ~     (all equal)
Import-8                             164k ± 0%      115k ± 0%  -29.88%  (p=0.000 n=10+10)
Node_WriteBytes/NoPreAllocate-8      9.00 ± 0%      6.00 ± 0%  -33.33%  (p=0.000 n=10+10)
Node_WriteBytes/PreAllocate-8        8.00 ± 0%      5.00 ± 0%  -37.50%  (p=0.000 n=10+10)
```

Fixes #419

### DIFF
--- a/encoding.go
+++ b/encoding.go
@@ -7,7 +7,14 @@ import (
 	"fmt"
 	"io"
 	"math/bits"
+	"sync"
 )
+
+var varintPool = &sync.Pool{
+	New: func() interface{} {
+		return &[binary.MaxVarintLen64]byte{}
+	},
+}
 
 // decodeBytes decodes a varint length-prefixed byte slice, returning it along with the number
 // of input bytes read.
@@ -108,7 +115,17 @@ func encodeUvarintSize(u uint64) int {
 
 // encodeVarint writes a varint-encoded integer to an io.Writer.
 func encodeVarint(w io.Writer, i int64) error {
-	var buf [binary.MaxVarintLen64]byte
+	// Use a pool here to reduce allocations.
+	//
+	// Though this allocates just 10 bytes on the stack, doing allocation for every calls
+	// cost us a huge memory. The profiling show that using pool save us ~30% memory.
+	//
+	// Since when we don't have concurrent access to the pool, the speed will nearly identical.
+	// If we need to support concurrent access, we can accept a *[binary.MaxVarintLen64]byte as
+	// input, so the caller can allocate just one and pass the same array pointer to each call.
+	buf := varintPool.Get().(*[binary.MaxVarintLen64]byte)
+	defer varintPool.Put(buf)
+
 	n := binary.PutVarint(buf[:], i)
 	_, err := w.Write(buf[0:n])
 	return err


cuonglm:
cc @odeke-em

odeke-em (approved):
LGTM, thank you @cuonglm! Funnily I had a similar change but directly prepared for cosmos-sdk/store/*

marbar3778 (approved):

encoding.go
```diff
@@ -106,9 +107,17 @@ func encodeUvarintSize(u uint64) int {
...
+	New: func() interface{} {
+		return &[binary.MaxVarintLen64]byte{}
+	},
+}
+
 // encodeVarint writes a varint-encoded integer to an io.Writer.
 func encodeVarint(w io.Writer, i int64) error {
-	var buf [binary.MaxVarintLen64]byte
+	buf := encodeVarintPool.Get().(*[binary.MaxVarintLen64]byte)
+	defer encodeVarintPool.Put(buf)
```
robert-zaremba, line 119: is there a real difference here? We are talking about allocating 10 bytes on on the stack. I think it's faster then going through a sync pool.

cuonglm: @robert-zaremba Yes, if you look at the issue for profile mem:

```
13.21GB  4.79% 52.74%    24.46GB  8.87%  github.com/cosmos/iavl.encodeVarint
```

Reduce 30% of 24GB will be a huge win.

What we deal is actually the number of allocations, since when we nearly don't have concurrently access to the pool, so the speed will be nearly identical.

If someday, we need to support concurrently call to encodeVarint, then we can change it to accept a `*[binary.MaxVarintLen64]` as input, so the caller can allocate just one and pass the same pointer each call to encodeVarint.

robert-zaremba: Good tip. Could you add a comment about this in the code, please?

robert-zaremba (approved):
Please add a comment to the `encodeVarint` before merging. Thanks for benchmarking!

encoding.go
cuonglm: Done

cuonglm:
cc @odeke-em @robert-zaremba for re-running the workflow.

robert-zaremba:
@cuonglm  - you need to merge master into this branch

cuonglm:
@robert-zaremba Done!

cuonglm:
Kindly ping @marbar3778 @robert-zaremba to merge this, so we can continue with optimizing `encodeUvarint`.

PR #447, opened 2021-11-30
Refactor Get() to work with new faster cache

jtieri:
Created a new cache in `nodedb.go` that would be maintained for our newer `FastNode` structs and implemented the pieces necessary for changing `Get()` as per [this](https://github.com/cosmos/iavl/commit/75d6e9829a4a29b1b2fe02d0f2ead3dd7ae9388d#diff-f53b00cecfca17dc100d06e79d2ce350e3b92c09d33fe69d384feaba56d2aa40R154-R160)

### DIFF
--- a/fast_node.go
+++ b/fast_node.go
@@ -15,15 +15,16 @@ type FastNode struct {
 }
 
 // NewFastNode returns a new fast node from a value and version.
-func NewFastNode(value []byte, version int64) *FastNode {
+func NewFastNode(key []byte, value []byte, version int64) *FastNode {
 	return &FastNode{
+		key:                  key,
 		versionLastUpdatedAt: version,
 		value:                value,
 	}
 }
 
-// MakeFastNode constructs an *FastNode from an encoded byte slice.
-func MakeFastNode(buf []byte) (*FastNode, error) {
+// DeserializeFastNode constructs an *FastNode from an encoded byte slice.
+func DeserializeFastNode(buf []byte) (*FastNode, error) {
 	val, n, cause := decodeBytes(buf)
 	if cause != nil {
 		return nil, errors.Wrap(cause, "decoding fastnode.value")
@@ -35,7 +36,10 @@ func MakeFastNode(buf []byte) (*FastNode, error) {
 		return nil, errors.Wrap(cause, "decoding fastnode.version")
 	}
 
-	fastNode := NewFastNode(val, ver)
+	fastNode := &FastNode{
+		versionLastUpdatedAt: ver,
+		value:                val,
+	}
 
 	return fastNode, nil
 }
--- a/fast_node.go
+++ b/fast_node.go
@@ -1,0 +1,41 @@
+package iavl
+
+import (
+	"github.com/pkg/errors"
+)
+
+// NOTE: This file favors int64 as opposed to int for size/counts.
+// The Tree on the other hand favors int.  This is intentional.
+
+type FastNode struct {
+	key                  []byte
+	versionLastUpdatedAt int64
+	value                []byte
+	leafHash             []byte // TODO: Look into if this would help with proof stuff.
+}
+
+// NewFastNode returns a new fast node from a value and version.
+func NewFastNode(value []byte, version int64) *FastNode {
+	return &FastNode{
+		versionLastUpdatedAt: version,
+		value:                value,
+	}
+}
+
+// MakeFastNode constructs an *FastNode from an encoded byte slice.
+func MakeFastNode(buf []byte) (*FastNode, error) {
+	val, n, cause := decodeBytes(buf)
+	if cause != nil {
+		return nil, errors.Wrap(cause, "decoding fastnode.value")
+	}
+	buf = buf[n:]
+
+	ver, _, cause := decodeVarint(buf)
+	if cause != nil {
+		return nil, errors.Wrap(cause, "decoding fastnode.version")
+	}
+
+	fastNode := NewFastNode(val, ver)
+
+	return fastNode, nil
+}
--- a/immutable_tree.go
+++ b/immutable_tree.go
@@ -164,9 +164,8 @@ func (t *ImmutableTree) Get(key []byte) (index int64, value []byte) {
 
 	if fastNode.versionLastUpdatedAt > t.version {
 		return t.root.get(t, key)
-	} else {
-		return 0, value // TODO determine index and adjust this appropriately
 	}
+	return 0, value // TODO determine index and adjust this appropriately
 }
 
 // GetByIndex gets the key and value at the specified index.
--- a/immutable_tree.go
+++ b/immutable_tree.go
@@ -146,19 +146,26 @@ func (t *ImmutableTree) Export() *Exporter {
 // otherwise. The returned value must not be modified, since it may point to data stored within
 // IAVL.
 // TODO: Understand what is this index? Index on its own isn't well defined
-// index across all leaves?
+// index = index of the leaf node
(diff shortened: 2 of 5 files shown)


immutable_tree.go
```diff
@@ -146,19 +146,26 @@ func (t *ImmutableTree) Export() *Exporter {
 // otherwise. The returned value must not be modified, since it may point to data stored within
 // IAVL.
 // TODO: Understand what is this index? Index on its own isn't well defined
-// index across all leaves?
+// index = index of the leaf node
```
ValarDragon, line 149: But index amongst what? Like between the `m` branches in a BTree?

fast_node.go
```diff
@@ -0,0 +1,41 @@
...
+
+type FastNode struct {
+	key                  []byte
+	versionLastUpdatedAt int64
+	value                []byte
+	leafHash             []byte // TODO: Look into if this would help with proof stuff.
+}
+
+// NewFastNode returns a new fast node from a value and version.
+func NewFastNode(value []byte, version int64) *FastNode {
```
ValarDragon, line 18:
```suggestion
func NewFastNode(key []byte, value []byte, version int64) *FastNode {
```

I think it has to have key & value

fast_node.go
```diff
@@ -0,0 +1,41 @@
...
+	return &FastNode{
+		versionLastUpdatedAt: version,
+		value:                value,
+	}
+}
+
+// MakeFastNode constructs an *FastNode from an encoded byte slice.
+func MakeFastNode(buf []byte) (*FastNode, error) {
```
ValarDragon, line 26: Should we call this `DeserializeFastNode` (its a naming issue in existing code atm)

nodedb.go
```diff
@@ -113,6 +120,40 @@ func (ndb *nodeDB) GetNode(hash []byte) *Node {
 	return node
 }
 
+func (ndb *nodeDB) GetFastNode(key []byte) (*FastNode, error) {
+	ndb.mtx.Lock()
+	defer ndb.mtx.Unlock()
```
ValarDragon, line 125: We can do this as RLock

nodedb.go
```diff
@@ -113,6 +120,40 @@ func (ndb *nodeDB) GetNode(hash []byte) *Node {
...
+
+	// Check the cache.
+	if elem, ok := ndb.fastNodeCache[string(key)]; ok {
+		// Already exists. Move to back of fastNodeCacheQueue.
+		ndb.fastNodeCacheQueue.MoveToBack(elem)
+		return elem.Value.(*FastNode), nil
+	}
+
+	// Doesn't exist, load.
+	buf, err := ndb.db.Get(key)
```
ValarDragon, line 139:
```suggestion
	// Doesn't exist, load.
	// TODO: Revisit if this even needs to be under mutex
	buf, err := ndb.db.Get(key)
```

immutable_tree.go
jtieri: "The index is the index in the list of leaf nodes sorted lexicographically by key. The leftmost leaf has index 0. It's neighbor has index 1 and so on." I noticed this in the docs found [here](https://github.com/cosmos/iavl/blob/master/docs/tree/immutable_tree.md#get)

jtieri: I'll add this in a comment in the codebase so that is immediately clear what index represents.

fast_node.go
jtieri: Yeah I noticed that the original node struct was initialized and then had it's hash set so followed that [pattern](https://github.com/cosmos/iavl/blob/577c9a8615e9edd17f2bba1c9c2b0547f8e96631/node.go#L44) with our fast node but looking into the call-sites for the original node struct constructor, I don't see any reason why that would be necessary.

fast_node.go
jtieri: I think this is a much better naming convention since that is what is actually going on here.

fast_node.go
jtieri: You're right, I overlooked this after adding the key field to the struct.

nodedb.go
jtieri: On [L134](https://github.com/cosmos/iavl/pull/447/files#diff-6df49839605c94e5caeb7d4549afe697ad97a00144ef26ec6746d9b5e8a3019cR134) there is the chance that the cache-queue is updated which would still need to be protected for writes

immutable_tree.go
ValarDragon: Sweet, tysm

nodedb.go
ValarDragon: oh I see. We should probably add a note to make a second Write lock just for that queue later then. (Difference between an RLock and WLock for performance here will be notable, since we want/need parallel queries)

Can write a TODO for this, and then come back to it later?

jtieri: Sounds good to me, added a TODO to circle back around to later.

immutable_tree.go
```diff
@@ -145,20 +145,27 @@ func (t *ImmutableTree) Export() *Exporter {
...
+	fastNode, err := t.ndb.GetFastNode(key)
+	if fastNode == nil || err != nil {
+		return t.root.get(t, key)
+	}
+	value = fastNode.value
+	if value == nil {
+		return t.root.get(t, key)
+	}
+
+	if fastNode.versionLastUpdatedAt > t.version {
```
ValarDragon, line 165:
```suggestion
	// cache node is too new, so read from historical tree
	if fastNode.versionLastUpdatedAt > t.version {
```

ValarDragon (approved):

immutable_tree.go
```diff
@@ -145,20 +145,28 @@ func (t *ImmutableTree) Export() *Exporter {
...
 	}
-	// IMPLEMENT FOLLOWING PSUEDOCODE
-	// value, version := t.nodeDb.fastGet(key)
-	// if value == nil { return t.root.get(t, key)}
-	// if version > t.version { return t.root.get(t, key)}
-	// else: return value
-	// TODO: Figure out what index is
 
-	return t.root.get(t, key)
+	fastNode, err := t.ndb.GetFastNode(key)
```
robert-zaremba, line 156: Should we log error here? When this operation will fail? Would be good to add a comment. @ValarDragon

nodedb.go
```diff
@@ -113,6 +120,41 @@ func (ndb *nodeDB) GetNode(hash []byte) *Node {
 	return node
 }
 
+func (ndb *nodeDB) GetFastNode(key []byte) (*FastNode, error) {
+	ndb.mtx.Lock()
+	defer ndb.mtx.Unlock()
+
+	if len(key) == 0 {
+		panic("nodeDB.GetFastNode() requires key")
+	}
```
robert-zaremba, line 129: please do not panic!

nodedb.go
robert-zaremba: we can move the mutex operation to line 130

nodedb.go
jtieri: Good catch Robert, I meant to replace all the panics used with the traditional node logic with errors so that failure to fetch a fast node simply falls back to the original logic in IAVL.

I'll get this taken care of in the next PR

immutable_tree.go
jtieri: The idea is that if this `GetFastNode()` call fails for any reason, we assume there is no `FastNode` equivalent for this key in the nodedb and fall back to the original, yet slower, IAVL logic in place. 

I'll add debug logging here and add a comment to make this more clear.

robert-zaremba: I think the `FastNode` is confusing. Maybe we should rename to `CachedNode`?

PR #448, opened 2021-12-01
change `RLock` to `Lock` for MutableTree.VersionExists

YunSuk-Yeo:
We monitored, sometimes we met concurrent read write map access panic and found there is some miss-configured lock.

`MutableTree.VersionExists` has write operation, so we have to change to use write lock.

https://github.com/cosmos/iavl/blob/577c9a8615e9edd17f2bba1c9c2b0547f8e96631/mutable_tree.go#L64-L79


marbar3778 (approved):
thanks for the PR!!

robert-zaremba:
@odeke-em it turns out that the Mutex improvements are mis-targetted. I'm really curious why we can't use RWMutex. Do you have an idea?

marbar3778:
they can be used and are used in places, just can't be used here

PR #456, opened 2021-12-16
perf: use a reusable bytes.Buffer to speed up MutableTree.String

### LINKED ISSUE #455: MutableTree.String and nodeDB.String consume unnecessary memory and CPU time using naive string concatenation
## String concatenation
If we examine the code for (nodeDB.String) it constructs a string naively by plain old `str += "..."` but unfortunately this is quite inefficient, if we examine
https://github.com/cosmos/iavl/blob/258c3375dd78a8ee610edc51dcbacd494e778d66/nodedb.go#L705
https://github.com/cosmos/iavl/blob/258c3375dd78a8ee610edc51dcbacd494e778d66/nodedb.go#L710
https://github.com/cosmos/iavl/blob/258c3375dd78a8ee610edc51dcbacd494e778d66/nodedb.go#L716-L726

## Unnecessary byteslice->string conversion even in fmt.Sprintf("%s", string(byteslice)
The fmt "%s" verb knows how to format byteslices as a strings, meanwhile this code performs an unnecessary conversion
https://github.com/cosmos/iavl/blob/258c3375dd78a8ee610edc51dcbacd494e778d66/nodedb.go#L704-L706
https://github.com/cosmos/iavl/blob/258c3375dd78a8ee610edc51dcbacd494e778d66/nodedb.go#L709-L711

odeke-em:
The prior code used a naive string concatenation
str += "..."

which is very slow and inefficient especially when iterating over
all the keys, but also it performed an unnecessary byteslice->string
conversion for all arguments inside (nodeDB).traverse* using

    str += fmt.Sprintf("%s: %x\n", string(key), value)

notice the `string(key)` passed into fmt.Sprintf("%s"?
At bare minimum that code should just simply be:

    str += fmt.Sprintf("%s: %x\n", key, value)

per https://twitter.com/orijtech/status/1462100381803102213?s=20
which saves a whole lot of cycles and RAM.

This change uses:
* (*bytes.Buffer) in combination with fmt.Fprintf
* A sync.Pool with (*bytes.Buffer) values to reuse buffers

and the results are profound:

```shell
name          old time/op    new time/op    delta
TreeString-8     111ms ± 8%       4ms ± 4%  -96.01%  (p=0.000 n=20+20)

name          old alloc/op   new alloc/op   delta
TreeString-8     734MB ± 0%       2MB ± 1%  -99.72%  (p=0.000 n=20+20)

name          old allocs/op  new allocs/op  delta
TreeString-8     37.0k ± 0%     28.7k ± 0%  -22.40%  (p=0.000 n=20+20)
```

Fixes #455

### DIFF
--- a/nodedb.go
+++ b/nodedb.go
@@ -698,33 +698,36 @@ func (ndb *nodeDB) traverseNodes(fn func(hash []byte, node *Node)) {
 }
 
 func (ndb *nodeDB) String() string {
-	var str string
+	buf := bufPool.Get().(*bytes.Buffer)
+	defer bufPool.Put(buf)
+	buf.Reset()
+
 	index := 0
 
 	ndb.traversePrefix(rootKeyFormat.Key(), func(key, value []byte) {
-		str += fmt.Sprintf("%s: %x\n", string(key), value)
+		fmt.Fprintf(buf, "%s: %x\n", key, value)
 	})
-	str += "\n"
+	buf.WriteByte('\n')
 
 	ndb.traverseOrphans(func(key, value []byte) {
-		str += fmt.Sprintf("%s: %x\n", string(key), value)
+		fmt.Fprintf(buf, "%s: %x\n", key, value)
 	})
-	str += "\n"
+	buf.WriteByte('\n')
 
 	ndb.traverseNodes(func(hash []byte, node *Node) {
 		switch {
 		case len(hash) == 0:
-			str += "<nil>\n"
+			buf.WriteByte('\n')
 		case node == nil:
-			str += fmt.Sprintf("%s%40x: <nil>\n", nodeKeyFormat.Prefix(), hash)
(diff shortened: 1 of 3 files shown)


marbar3778 (approved):
lgtm,

ValarDragon (approved):
Nice catch! 

I'm a bit skeptical of the sync pool being helpful here since the data _should_ be small. (a question of CAS ops + defer + reflection type casting taking less time than a heap alloc)

However, this change is clearly much more helpful than not having it! And it'll be pretty easy to see if theres any overheads due to sync pool in the future

marbar3778:
@odeke-em can you update this branch with master then I can merge it

odeke-em:
> @odeke-em can you update this branch with master then I can merge it

Done, thank you @marbar3778!

Thank you @ValarDragon @marbar3778 for the reviews!

robert-zaremba:
Probably we could optimize it even further with `strings.Join`.

PR #459, opened 2022-01-12
WIP: FastNode Get and Set

akhtariev:


akhtariev:
Opened by mistake in the wrong repo. Reopened in [Osmosis](https://github.com/osmosis-labs/iavl/pull/5)

robert-zaremba:
If this makes  sense, would be good to push it here as well ;)

marbar3778:
this will be pushed upstream touched base with @ValarDragon today.

PR #460, opened 2022-01-14
chore: introduce internal and kill common

marbar3778:
introduce internal folder to clean up the root, kill the common folder.


robert-zaremba:
Why do you want to hide encoding functions? Maybe someone is using it, I would prefer to not break that part of the API.

marbar3778:
> Why do you want to hide encoding functions? Maybe someone is using it, I would prefer to not break that part of the API.

it was all private, there is no api breaking in this pr. This is meant to reduce root level files for legibility reasons.

ValarDragon (approved):

PR #466, opened 2022-01-29
Update docker.yml and Dockerfile, fix ci

faddat:
Use github container registry


Dockerfile
```diff
@@ -1,26 +1,12 @@
-FROM golang:1.14 as build
-
-WORKDIR /iavl
+# This docker image is designed to be used in CI for benchmarks and also by developers wanting an environment that always has the lastest rocksdb and cleveldb. 
+FROM faddat/archlinux
```
robert-zaremba, line 2: how about using one of the official arch or alpine linux distros?

robert-zaremba (approved):
utACK

Dockerfile
robert-zaremba: ping

.github/workflows/docker.yml
```diff
@@ -11,45 +12,30 @@ on:
...
+        name: Build and push
+        uses: docker/build-push-action@v2
         with:
-          file: ./Dockerfile
-          push: ${{ github.event_name != 'pull_request' }}
-          tags: ${{ steps.prep.outputs.tags }}
+          context: .
+          platforms: linux/amd64,linux/arm64
+          push: true
+          tags: ghcr.io/${{ github.repository_owner }}/iavl:latest
```
marbar3778, line 41: love this!! we should migrate everything we have to this

Dockerfile
faddat: hey sure thing

.github/workflows/docker.yml
faddat: so, ok-- this is a package deal with the above docker image, at least in my experience the multiarch stuff.  Let me see what I can do.

marbar3778:
we removed this docker file

Issue #482, opened 2022-03-11
Support deletion of latest saved version


yihuang:
When we try to implement the rollback command in cosmos-sdk, we want to delete the latest saved version, but that's not allowed in the current version of iavl, what's the reason for that, and is it possible to relax that restriction?

marbar3778:
@yihuang do you still need this?

yihuang:
> @yihuang do you still need this?

I think LoadVersionForOverwriting do just that, although it's very slow to run in a big db.

PR #483, opened 2022-03-11
relax the restriction of delete latest version

yihuang:
Closes: #482

This is required for the rollback command to work in cosmos-sdk (https://github.com/cosmos/cosmos-sdk/issues/11333#issuecomment-1064672605).
I'm not sure if there are other consequences of this, but my local test with rollback command seems fine using this change.

### DIFF
--- a/mutable_tree.go
+++ b/mutable_tree.go
@@ -566,13 +566,10 @@ func (tree *MutableTree) deleteVersion(version int64) error {
 	if version <= 0 {
 		return errors.New("version must be greater than 0")
 	}
-	if version == tree.version {
-		return errors.Errorf("cannot delete latest saved version (%d)", version)
-	}
 	if !tree.VersionExists(version) {
 		return errors.Wrap(ErrVersionDoesNotExist, "")
 	}
-	if err := tree.ndb.DeleteVersion(version, true); err != nil {
+	if err := tree.ndb.DeleteVersion(version, false); err != nil {
 		return err
 	}
 
--- a/nodedb.go
+++ b/nodedb.go
@@ -281,8 +281,8 @@ func (ndb *nodeDB) DeleteVersionsRange(fromVersion, toVersion int64) error {
 	defer ndb.mtx.Unlock()
 
 	latest := ndb.getLatestVersion()
-	if latest < toVersion {
-		return errors.Errorf("cannot delete latest saved version (%d)", latest)
+	if latest+1 < toVersion {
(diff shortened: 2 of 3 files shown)


nodedb.go
```diff
@@ -281,8 +281,8 @@ func (ndb *nodeDB) DeleteVersionsRange(fromVersion, toVersion int64) error {
 	defer ndb.mtx.Unlock()
 
 	latest := ndb.getLatestVersion()
-	if latest < toVersion {
-		return errors.Errorf("cannot delete latest saved version (%d)", latest)
+	if latest+1 < toVersion {
+		return errors.Errorf("cannot delete future versions (..%d)", toVersion)
```
yihuang, line 285:
```suggestion
		return errors.Errorf("cannot delete future versions (%d..%d)", latest+1, toVersion)
```

mutable_tree.go
```diff
@@ -566,13 +566,10 @@ func (tree *MutableTree) deleteVersion(version int64) error {
...
 		return errors.New("version must be greater than 0")
 	}
-	if version == tree.version {
-		return errors.Errorf("cannot delete latest saved version (%d)", version)
-	}
 	if !tree.VersionExists(version) {
 		return errors.Wrap(ErrVersionDoesNotExist, "")
 	}
-	if err := tree.ndb.DeleteVersion(version, true); err != nil {
+	if err := tree.ndb.DeleteVersion(version, false); err != nil {
```
ValarDragon, line 572: Are you sure this works? You have to move every node in the merkle tree for the historical state at the prior height, to being in the live state section of node db

yihuang: I'm not sure, only did basic unit test in cosmos-sdk side here:https://github.com/cosmos/cosmos-sdk/pull/11361, I need more understanding on internals of iavl I think. 
Do you think there is an efficient way to achieve this rollback feature?

yihuang: I think we don't need to patch iavl at all, there's `LoadVersionForOverwriting` which is perfect for rollback state.

robert-zaremba: We should also check if it will be pruned correctly.

yihuang: https://github.com/cosmos/cosmos-sdk/pull/11361/files#diff-54fe31c47e1d9d86bbaf2face4c464c1f5a74e3c81ca5ac9a27a25a157752592R56
it's checked in this unit test, that one can rollback and recreate new states.

PR #493, opened 2022-04-21
pebbledb

faddat:
PebbleDB is about twice as fast as the rest, sometimes more or less of course.

The acceptance test we are using:

1) passes all tm-db tests
2) runs osmo from genesis for 3 hours
3) runs gaia from genesis for 3 hours


marbar3778:
Hey, not sure why we are adding new dbs without fixing the ones we have. I have talked with the get team and they said pebbledb is significantly slower than goleveldb if used properly. 

Going to close this for now

Issue #212, opened 2020-02-23
Don't panic


erikgrinaker:
IAVL mostly uses panics for error handling. It should return errors instead.

p4u:
I am experiencing this issue. Instead of Panic IAVL should return errors in order to be able to handle them on the upper layer.

```
panic: Value missing for hash 5c762dd92f621bd3df4f93cdd153d5639590de63f8be117c317615faab68c0c6 corresponding to nodeKey n\v-�/b�O���S�c���c��|1v��h��

goroutine 1452118376 [running]:
github.com/tendermint/iavl.(*nodeDB).GetNode(0xc001fd0600, 0xc0724f3ea0, 0x20, 0x20, 0x0)
	/go/pkg/mod/github.com/tendermint/iavl@v0.13.3/nodedb.go:109 +0x7bf
github.com/tendermint/iavl.(*Node).getRightNode(...)
	/go/pkg/mod/github.com/tendermint/iavl@v0.13.3/node.go:417
github.com/tendermint/iavl.(*Node).get(0xc071dd45a0, 0xc0c6762b40, 0xc021c80c60, 0x81, 0x90, 0x81, 0xffffffffffffffff, 0xc0c934ad40, 0xc0475daff2)
	/go/pkg/mod/github.com/tendermint/iavl@v0.13.3/node.go:174 +0x2f2
github.com/tendermint/iavl.(*Node).get(0xc071dd4500, 0xc0c6762b40, 0xc021c80c60, 0x81, 0x90, 0x81, 0x1, 0x1, 0x30ab980)
	/go/pkg/mod/github.com/tendermint/iavl@v0.13.3/node.go:172 +0x1ac
github.com/tendermint/iavl.(*Node).get(0xc071c8bf40, 0xc0c6762b40, 0xc021c80c60, 0x81, 0x90, 0x81, 0xffffffffffffffff, 0xc017393410, 0xe8b266)
	/go/pkg/mod/github.com/tendermint/iavl@v0.13.3/node.go:175 +0x26f
github.com/tendermint/iavl.(*Node).get(0xc071f69cc0, 0xc0c6762b40, 0xc021c80c60, 0x81, 0x90, 0x81, 0x1, 0xf, 0xc0054bea00)
	/go/pkg/mod/github.com/tendermint/iavl@v0.13.3/node.go:172 +0x1ac
[...]

PR #500, opened 2022-05-05
refactor!: Remove panics

facundomedica:
WIP, removing some panics by returning errors instead. Looking to solve #212


ValarDragon:
Thanks for working on this!

facundomedica:
From the starting 49 panics we are now at 25.
7 are in *_test.go so won't be paying too much attention on those
4 are in `ProofOp` has to comply with an interface
2 there are tests in place to ensure that these panic (not sure if we want to remove those)

So 12 panics to go 🥳 

I think I should be able to finish this up tomorrow, but I expect a lot of back and forth. I assume we might encounter cases in which we might go back to panics, but I can't tell.

robert-zaremba:
I'm in a full support for reducing panics. However, this will create lot of breaking changes in Cosmos SDK. So this has to be well coordinated.

node.go
```diff
@@ -122,9 +122,9 @@ func (node *Node) String() string {
 }
 
 // clone creates a shallow copy of a node with its hash set to nil.
-func (node *Node) clone(version int64) *Node {
+func (node *Node) clone(version int64) (*Node, error) {
 	if node.isLeaf() {
-		panic("Attempt to copy a leaf node")
+		return nil, fmt.Errorf("attempt to copy a leaf node")
```
robert-zaremba, line 127: let's create `const` errors and return them.
eg:
```
const ErrCloneLeafNode = fmt.Errorf("attempt to copy a leaf node")
```

robert-zaremba (approved):
utACK

cmd/iaviewer/main.go
```diff
@@ -47,7 +47,12 @@ func main() {
 	switch args[0] {
 	case "data":
 		PrintKeys(tree)
-		fmt.Printf("Hash: %X\n", tree.Hash())
+		hash, err := tree.Hash()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Error hashing tree: %s\n", err)
```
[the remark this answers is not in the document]
facundomedica: I think that's because it's a command that prints stuff to the command line output. So it uses fmt.Print everywhere

node.go
facundomedica: Added them everywhere I replaced a panic (and the error is not dynamic). More work will be needed on this later on

marbar3778:
this looks amazing!!

PR #505, opened 2022-06-10
add Statistics for cache miss rate

luohaha:
For a better user experience, add optional statistics for cache miss rate.

### DIFF
--- a/nodedb.go
+++ b/nodedb.go
@@ -173,7 +173,7 @@ func (ndb *nodeDB) GetFastNode(key []byte) (*FastNode, error) {
 	// Check the cache.
 	if elem, ok := ndb.fastNodeCache[string(key)]; ok {
 		if ndb.opts.Stat != nil {
-			ndb.opts.Stat.IncCacheHitCnt()
+			ndb.opts.Stat.IncFastCacheHitCnt()
 		}
 		// Already exists. Move to back of fastNodeCacheQueue.
 		ndb.fastNodeCacheQueue.MoveToBack(elem)
@@ -181,7 +181,7 @@ func (ndb *nodeDB) GetFastNode(key []byte) (*FastNode, error) {
 	}
 
 	if ndb.opts.Stat != nil {
-		ndb.opts.Stat.IncCacheMissCnt()
+		ndb.opts.Stat.IncFastCacheMissCnt()
 	}
 
 	// Doesn't exist, load.
--- a/nodedb.go
+++ b/nodedb.go
@@ -125,9 +125,16 @@ func (ndb *nodeDB) GetNode(hash []byte) (*Node, error) {
 
 	// Check the cache.
 	if elem, ok := ndb.nodeCache[string(hash)]; ok {
+		if ndb.opts.Stat != nil {
+			ndb.opts.Stat.IncCacheHitCnt()
+		}
 		// Already exists. Move to back of nodeCacheQueue.
 		ndb.nodeCacheQueue.MoveToBack(elem)
 		return elem.Value.(*Node), nil
+	}
+
+	if ndb.opts.Stat != nil {
+		ndb.opts.Stat.IncCacheMissCnt()
 	}
 
 	// Doesn't exist, load.
@@ -165,9 +172,16 @@ func (ndb *nodeDB) GetFastNode(key []byte) (*FastNode, error) {
 
 	// Check the cache.
 	if elem, ok := ndb.fastNodeCache[string(key)]; ok {
+		if ndb.opts.Stat != nil {
+			ndb.opts.Stat.IncCacheHitCnt()
+		}
 		// Already exists. Move to back of fastNodeCacheQueue.
 		ndb.fastNodeCacheQueue.MoveToBack(elem)
 		return elem.Value.(*FastNode), nil
+	}
+
+	if ndb.opts.Stat != nil {
+		ndb.opts.Stat.IncCacheMissCnt()
 	}
 
 	// Doesn't exist, load.
--- a/options.go
+++ b/options.go
@@ -1,5 +1,37 @@
 package iavl
 
+import "sync/atomic"
+
+// Statisc about db runtime state
+type Statistics struct {
+	// Each time GetNode and GetFastNode operation hit cache
+	cacheHitCnt uint64
+
+	// Each time GetNode and GetFastNode operation miss cache
+	cacheMissCnt uint64
+}
+
+func (stat *Statistics) IncCacheHitCnt() {
+	atomic.AddUint64(&stat.cacheHitCnt, 1)
+}
+
+func (stat *Statistics) IncCacheMissCnt() {
+	atomic.AddUint64(&stat.cacheMissCnt, 1)
+}
+
+func (stat *Statistics) GetCacheHitCnt() uint64 {
+	return atomic.LoadUint64(&stat.cacheHitCnt)
+}
+
+func (stat *Statistics) GetCacheMissCnt() uint64 {
+	return atomic.LoadUint64(&stat.cacheMissCnt)
+}
+
+func (stat *Statistics) Reset() {
+	atomic.StoreUint64(&stat.cacheHitCnt, 0)
+	atomic.StoreUint64(&stat.cacheMissCnt, 0)
+}
+
 // Options define tree options.
 type Options struct {
 	// Sync synchronously flushes all writes to storage, using e.g. the fsync syscall.
@@ -10,6 +42,9 @@ type Options struct {
 	// this, an error is returned when loading the tree. Only used for the initial SaveVersion()
 	// call.
 	InitialVersion uint64
+
+	// When Stat is not nil, statistical logic needs to be executed
+	Stat *Statistics
 }
 
 // DefaultOptions returns the default options for IAVL.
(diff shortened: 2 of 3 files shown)


p0mvn:
Hi @luohaha 

Thank you for this change. This is certainly useful to keep track of.

I have one question though - how do you expect this new statistic to be read or queried?

luohaha:
@p0mvn After new mutable tree, we can get a statistic with it. We can periodic read this statistic and set to a new metric, and then report to prometheus or print to log.

p0mvn:
Since `nodeDb` is an unexported field of the mutable tree, how can we read the statistic periodically? Is it possible to expose a method for doing so that would also synchronize access to the statistics?

options.go
```diff
@@ -10,6 +29,9 @@ type Options struct {
 	// this, an error is returned when loading the tree. Only used for the initial SaveVersion()
 	// call.
 	InitialVersion uint64
+
+	// When Stat is not nil, statistical logic needs to be executed
+	Stat *Statistics
```
p0mvn, line 34: Is it possible to define this on `nodeDb` instead?

Options are used for configuring iavl. On the other hand, this seems non-configuration specific

luohaha: https://github.com/facebook/rocksdb/blob/main/include/rocksdb/options.h#L609
Usage is just like rocksdb, eg:
```
Options options;
options.statistics = ROCKSDB_NAMESPACE::CreateDBStatistics();
Status s = DB::Open(options, kDBPath, &db);
...
options.statistics->getTickerCount(NUMBER_BLOCK_COMPRESSED);
```
Our usage:
```
opt.Statistics = MyStatistics
NewMutableTreeWithOpts(db, 100, &opt);
...
print(MyStatistics)
```

p0mvn: Thanks for the example.

Although it might be convenient in C++, I don't think it would be as pretty in Go

Relying on `opts.Statistics` pointer is prone to data races. In the linked example, the synchronization of the counters is likely handled in `getTickerCount(...)`

In our case, we currently don't have such a method. However, even if we had one, relying on the user on calling that method instead of accessing `opt.Statistics` directly is dangerous. Someone could easily misuse it and access one of the counters directly without a lock (e.g. `opts.Statistics.CacheHitCnt`) causing a data race.

So I think we could do one of the following:
- avoid exporting `CacheHitCnt` and `CacheMissCnt` and create methods like:
```
func (s *Statistics) GetCacheHitCnt() {...}
```
in which we atomically get the counters

OR

- define the Statistics on `nodeDb` instead of `Options`, expose getters on the mutable tree and atomically get the counters

Please let me know what you think

luohaha: I prefer expose getters on the Statistics than mutable tree, because more statistics item may be added in the future, and their getters will be more concise to be hidden on the Statistics.
And I also prefer to add Statistics to option, because statistical behavior could cause performance damage and no every one can accept this. So it is better to be optional.

p0mvn: Good call, sounds good to me

nodedb.go
```diff
@@ -125,11 +125,18 @@ func (ndb *nodeDB) GetNode(hash []byte) (*Node, error) {
 
 	// Check the cache.
 	if elem, ok := ndb.nodeCache[string(hash)]; ok {
+		if ndb.opts.Stat != nil {
```
p0mvn, line 128: Could you please also add a couple of unit tests with and without the `nil` `Stat`. 

Also, asserting that the counters get increased would be great.

options.go
```diff
@@ -1,5 +1,37 @@
 package iavl
 
+import "sync/atomic"
+
+// Statisc about db runtime state
+type Statistics struct {
+	// Each time GetNode and GetFastNode operation hit cache
+	cacheHitCnt uint64
```
p0mvn, line 8: We mainly have 2 types of nodes:
1. Regular (that you get with `GetNode(...)`
2. Fast (`GetFastNode(...)`

They each have their own cache (`nodeCache`, `fastNodeCache`):
https://github.com/cosmos/iavl/blob/640ad78af7f34486f4793b3e442c8d20c751daf0/nodedb.go#L79-L83

What do you think about having a separate counter for each cache?

nodedb.go
luohaha: Done

options.go
luohaha: That is great. Done

tree_test.go
```diff
@@ -1982,3 +1982,69 @@ func Benchmark_GetByIndex(b *testing.B) {
 		}
 	})
 }
+
+func TestNodeCacheStatisic(b *testing.T) {
```
p0mvn, line 1986: nit: suggest changing to `t` since it is idiomatic to name benchmarks as `b` but regular tests as `t`

```suggestion
func TestNodeCacheStatisic(t *testing.T) {
```

tree_test.go
```diff
@@ -1982,3 +1982,69 @@ func Benchmark_GetByIndex(b *testing.B) {
...
+	db, err := db.NewDB("test", db.MemDBBackend, "")
+	require.NoError(b, err)
+
+	// with cache enable
+	b.Run("with_cache", func(sub *testing.T) {
```
p0mvn, line 1991: Seems like a good candidate to make this table-driven to avoid repetition. Do you mind making it so?

tree_test.go
```diff
@@ -1982,3 +1982,69 @@ func Benchmark_GetByIndex(b *testing.B) {
...
+			key := []byte(strconv.Itoa(i))
+			val, err := it.Get(key)
+			require.NoError(b, err)
+			require.NotNil(b, val)
+			require.NotEmpty(b, val)
+		}
+
+		require.Equal(b, 0, int(opts.Stat.GetFastCacheHitCnt()))
+		require.Equal(b, numKeyVals, int(opts.Stat.GetFastCacheMissCnt()))
+		require.Equal(b, 1, int(opts.Stat.GetCacheMissCnt()))
```
p0mvn, line 2048: Can we make an assertion on `GetCacheHitCnt` too, please?

tree_test.go
```diff
@@ -1982,3 +1982,69 @@ func Benchmark_GetByIndex(b *testing.B) {
...
+		for i := 0; i < numKeyVals; i++ {
+			key := []byte(strconv.Itoa(i))
+			val, err := it.Get(key)
+			require.NoError(b, err)
+			require.NotNil(b, val)
+			require.NotEmpty(b, val)
+		}
+		require.Equal(b, numKeyVals, int(opts.Stat.GetFastCacheHitCnt()))
+		require.Equal(b, 0, int(opts.Stat.GetFastCacheMissCnt()))
+		require.Equal(b, 1, int(opts.Stat.GetCacheHitCnt()))
```
p0mvn, line 2017: Let's make an assertion on miss as well?

p0mvn (commented):
Great work so far, a couple more comments. Let me know what you think

tree_test.go
luohaha: Done.

p0mvn: Table-driven tests look like the following:
```
func TestSplit(t *testing.T) {
    tests := map[string]struct {
        input string
        sep   string
        expected  []string
    }{
        "simple":       {input: "a/b/c", sep: "/", expected: []string{"a", "b", "c"}},
        "wrong sep":    {input: "a/b/c", sep: ",", expected: []string{"a/b/c"}},
        "no sep":       {input: "abc", sep: "/", expected: []string{"abc"}},
        "trailing sep": {input: "a/b/c/", sep: "/", expected: []string{"a", "b", "c"}},
    }

    for name, tc := range tests {
        t.Run(name, func(t *testing.T) {
            got := Split(tc.input, tc.sep)
            if !reflect.DeepEqual(tc.expected, got) {
                t.Fatalf("expected: %v, got: %v", tc.want, got)
            }
        })
    }
}
```

where there is a `tests` slice and then you iterate over each and run it. Each test case has a name, inputs, and expected values.

Source: https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

A good example in this repo: https://github.com/cosmos/iavl/blob/abaaacdcf919da1fb9fb780862346dcb99b60fd4/internal/encoding/encoding_test.go#L11

tree_test.go
```diff
@@ -1982,3 +1982,49 @@ func Benchmark_GetByIndex(b *testing.B) {
...
 	})
 }
+
+func TestNodeCacheStatisic(t *testing.T) {
+	db, err := db.NewDB("test", db.MemDBBackend, "")
+	require.NoError(t, err)
+	const numKeyVals = 100000
+	testcases := map[string]struct {
+		cacheSize int
+		expect    []int
```
p0mvn, line 1992: Instead of having this as `[]int`, can we split it into a field for each.
E.g.
```
expectFastHit int
expectFastMiss int
...
```

Currently, it's not intuitive which value is represented by which index in `expect`

p0mvn (approved):
LGTM once the final comment is addressed. Great work

tree_test.go
luohaha: Done.

marbar3778:
Thanks for the reviews @p0mvn

luohaha:
@p0mvn @marbar3778 When will we publish IAVL-v0.18.1 ? After this, I can add some metrics about cache miss rate in cosmos-sdk rely on IAVL-v0.18.1

marbar3778:
it is part of our current sprint, should happen over the next two weeks

PR #506, opened 2022-06-14
refactor: implement cache abstraction and unit test

p0mvn:
Backport of: https://github.com/osmosis-labs/iavl/pull/38

**Context**
- See here: https://github.com/osmosis-labs/osmosis/issues/1163
- To sum up:
  * better abstraction for cache to eventually enable the limit by bytes as opposed to the number of nodes
  * unit-tested cache
  * avoid caching nodes with the same key (the old one is now replaced)
  * cache adapted from: https://github.com/golang/groupcache/blob/master/lru/lru.go

This change has been running on Osmosis for almost 2 months so it is relatively low-risk.

**Original Benchstat from Osmosis fork**
```
name                                                                 old time/op    new time/op    delta
Medium/goleveldb-100000-100-16-40/query-no-in-tree-guarantee-fast-4    4.22µs ± 9%    4.15µs ± 6%     ~     (p=1.000 n=5+5)
Medium/goleveldb-100000-100-16-40/query-no-in-tree-guarantee-slow-4    15.3µs ±11%    16.6µs ± 6%     ~     (p=0.056 n=5+5)
Medium/goleveldb-100000-100-16-40/query-hits-fast-4                     566ns ± 8%     613ns ±11%     ~     (p=0.222 n=5+5)
Medium/goleveldb-100000-100-16-40/query-hits-slow-4                    21.2µs ± 5%    21.0µs ± 4%     ~     (p=0.690 n=5+5)
Medium/goleveldb-100000-100-16-40/iteration-fast-4                     78.0ms ± 6%    81.1ms ±11%     ~     (p=0.548 n=5+5)
Medium/goleveldb-100000-100-16-40/iteration-slow-4                      1.79s ±13%     1.90s ±11%     ~     (p=0.222 n=5+5)
Medium/goleveldb-100000-100-16-40/update-4                              233µs ±15%     203µs ± 8%  -12.78%  (p=0.008 n=5+5)
Medium/goleveldb-100000-100-16-40/block-4                              29.0ms ± 4%    25.8ms ±10%  -11.04%  (p=0.016 n=5+5)

name                                                                 old alloc/op   new alloc/op   delta
Medium/goleveldb-100000-100-16-40/query-no-in-tree-guarantee-fast-4      814B ± 0%      814B ± 0%     ~     (all equal)
Medium/goleveldb-100000-100-16-40/query-no-in-tree-guarantee-slow-4    1.41kB ± 1%    1.41kB ± 0%     ~     (p=1.000 n=5+5)
Medium/goleveldb-100000-100-16-40/query-hits-fast-4                     0.00B          0.00B          ~     (all equal)
Medium/goleveldb-100000-100-16-40/query-hits-slow-4                    1.99kB ± 1%    2.00kB ± 1%     ~     (p=0.460 n=5+5)
Medium/goleveldb-100000-100-16-40/iteration-fast-4                     29.3MB ± 0%    29.3MB ± 0%     ~     (p=0.381 n=5+5)
Medium/goleveldb-100000-100-16-40/iteration-slow-4                      276MB ± 0%     276MB ± 0%   +0.02%  (p=0.029 n=4+4)
Medium/goleveldb-100000-100-16-40/update-4                             47.4kB ± 2%    46.5kB ± 4%     ~     (p=0.151 n=5+5)
Medium/goleveldb-100000-100-16-40/block-4                              5.65MB ± 2%    5.41MB ± 2%   -4.27%  (p=0.008 n=5+5)

name                                                                 old allocs/op  new allocs/op  delta
Medium/goleveldb-100000-100-16-40/query-no-in-tree-guarantee-fast-4      16.0 ± 0%      16.0 ± 0%     ~     (all equal)
Medium/goleveldb-100000-100-16-40/query-no-in-tree-guarantee-slow-4      24.0 ± 0%      24.0 ± 0%     ~     (all equal)
Medium/goleveldb-100000-100-16-40/query-hits-fast-4                      0.00           0.00          ~     (all equal)
Medium/goleveldb-100000-100-16-40/query-hits-slow-4                      34.0 ± 0%      34.0 ± 0%     ~     (all equal)
Medium/goleveldb-100000-100-16-40/iteration-fast-4                       523k ± 0%      523k ± 0%     ~     (p=0.484 n=5+5)
Medium/goleveldb-100000-100-16-40/iteration-slow-4                      4.71M ± 0%     4.71M ± 0%   +0.01%  (p=0.029 n=4+4)
Medium/goleveldb-100000-100-16-40/update-4                                502 ± 9%       482 ± 6%     ~     (p=0.421 n=5+5)
Medium/goleveldb-100000-100-16-40/block-4                               62.1k ± 3%     59.6k ± 2%   -4.05%  (p=0.016 n=5+5)
```


marbar3778 (approved):
amazing, thank you!

robert-zaremba:
Did anyone make an in depth review?

cache/cache.go
```diff
@@ -0,0 +1,96 @@
...
+	// Len returns the cache length.
+	Len() int
+}
+
+// lruCache is an LRU cache implementation.
+type lruCache struct {
+	dict       map[string]*list.Element // FastNode cache.
+	cacheLimit int                      // FastNode cache size limit in elements.
+	ll         *list.List               // LRU queue of cache elements. Used for deletion.
+}
```
robert-zaremba, line 37: I would prefer if we use https://github.com/hashicorp/golang-lru or https://github.com/hashicorp/golang-lru/tree/master/simplelru ? We are already using it in various places.

nodedb.go
```diff
@@ -98,18 +93,14 @@ func newNodeDB(db dbm.DB, cacheSize int, opts *Options) *nodeDB {
...
-		fastNodeCacheSize:  100000,
-		fastNodeCacheQueue: list.New(),
-		versionReaders:     make(map[int64]uint32, 8),
-		storageVersion:     string(storeVersion),
+		db:             db,
+		batch:          db.NewBatch(),
+		opts:           *opts,
+		latestVersion:  0, // initially invalid
+		nodeCache:      cache.New(cacheSize),
+		fastNodeCache:  cache.New(fastNodeCacheLimit),
```
robert-zaremba, line 101: why do we still need `fastNodeCache`?

robert-zaremba (commented):
Let's consider using https://github.com/hashicorp/golang-lru/tree/master/simplelru rather than implementing our own lRU.

cache/cache.go
p0mvn: I've looked into these options and, unfortunately, they do not allow for enough customization necessary for future work.

I'm planning to further improve the custom cache by adding the ability to have 2 kinds of limits:
- in terms of bytes
- in terms of the number of nodes

I have this work started here that I'm hoping to move to this repository: https://github.com/osmosis-labs/iavl/pull/40

More background about why I'm introducing this cache and its limits can be found here: https://github.com/osmosis-labs/osmosis/issues/1163

In short, we found it not intuitive configuring the limits in terms of nodes so we would like to provide the ability to do so in bytes only for fast cache

nodedb.go
p0mvn: According to benchmarks, removing it significantly degrades the performance: https://github.com/osmosis-labs/osmosis/issues/1041

p0mvn:
@robert-zaremba thanks for the review. Addressed all comments in threads. Please let me know what you think

cache/cache.go
robert-zaremba: Maybe let's add a short comment in the code, before someone else will come and propose to remove this and reuse other library.

nodedb.go
robert-zaremba: I looked at it too fast. I don't see a benchmark with `nodeCache and no-fastNodeCache`. The benchmark you have is `noCache` vs `with fastNodeCache`. Sorry, maybe I'm not interpreting it correctly as I don't see the whole setting.

p0mvn: I think I just wasn't consistent with how I was referring to the benchmarks, sorry for the confusion.

The comparison is most certainly between `both node caches` and `regular cache but no-fastNodeCache`.
When I referred to:
`cache.log` = `both node caches`
`nocache.log` = `regular cache but no-fastNodeCache`

Looking [at the benchstat](https://github.com/osmosis-labs/osmosis/issues/1041#issuecomment-1058788599), the delta is coming mostly from `query-hits-fast-16` and `iteration-fast-16`, proving that the regular cache was untouched. If it was touched, we would see delta in `query-hits-slow-16` and `iteration-slow-16 `

To explain the reasons for trying this, our node operators were getting OOM'ed and pprof results were showing memory growth only specific to fast cache. There was never a motivation for trying things w/o regular cache.

To mitigate the memory growth we decided to hardcode the value for the fast cache to `100000`
https://github.com/cosmos/iavl/blob/d07ef28ff2421163d95092fa4b4d27c41080de2f/nodedb.go#L33

The value was chosen by spinning up a bunch of nodes at 10k, 50k, 100k, 150 and 200k. 100k proved to be the most optimal.

Spinning up several nodes to find a good value seemed like an unnecessarily complex solution. That's why I decided to do this refactor to eventually introduce byte cache for fast nodes. I'm also hoping to expose a parameter in config to let node operators choose this byte value.

p0mvn:
gobencher is failing for unrelated to this PR reasons

cache/cache_bench_test.go
```diff
@@ -0,0 +1,67 @@
...
+			key: key,
+		})
+	}
+
+	randSeed := 498727689 // For deterministic tests
+	r := rand.New(rand.NewSource(int64(randSeed)))
+
+	for i := 0; i < b.N; i++ {
+		key := existentKeyMirror[r.Intn(len(existentKeyMirror))]
+		b.ResetTimer()
```
odeke-em, line 64: This benchmark doesn't make any sense. You keep resetting the the timer in the most important of the benchmark for every b.N iteration. This looks like a mistake.

cache/cache_bench_test.go
```diff
@@ -0,0 +1,67 @@
...
+		existentKeyMirror = append(existentKeyMirror, key)
+
+		cache.Add(&testNode{
+			key: key,
+		})
+	}
+
+	randSeed := 498727689 // For deterministic tests
+	r := rand.New(rand.NewSource(int64(randSeed)))
+
```
odeke-em, line 61: I think you should have instead invoked b.ResetTimer() before the loop.

odeke-em (commented):
Thank you @p0mvn for the work! Thanks @marbar3778 for the ping!

@p0mvn I've added some comments, please take a look.

cache/cache.go
p0mvn: @odeke-em thanks for the suggestion. I've investigated using ristretto by swapping it for fast nodes but ran into issues. 

The major one is a ~17% increase in time/op. I suspect that is due to ristretto having overhead from concurrency guarantees which we might not really need.

Also, some randomized tests started failing for reasons I could not explain.

The current state of the attempt, summary, and benchmarks can be found here: https://github.com/cosmos/iavl/pull/508

robert-zaremba:
@p0mvn any chance you will finish this PR this week?

robert-zaremba:
thanks!

cache/cache.go
```diff
@@ -0,0 +1,107 @@
...
+	GetKey() []byte
+}
+
+// Cache is an in-memory structure to persist nodes for quick access.
+// Please see lruCache for more details about why we need a custom
+// cache implementation.
+type Cache interface {
+	// Adds node to cache. If full and had to remove the oldest element,
+	// returns the oldest, otherwise nil.
+	Add(node Node) Node
```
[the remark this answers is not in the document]
p0mvn: There should never be a nil Node added. Would it be sufficient to add a comment saying that the contract for this method is that `node` is never nil?

From reading [this discussion](https://stackoverflow.com/questions/13476349/check-for-nil-and-nil-interface-in-go), I could either:
- use reflection to check against nil in the `Add` method
- define `IsNil()` method for the `Node` interface and implement it for every node struct.

However, if we were to go with one of the approaches, I think that this would be trading performance for safety unnecessarily at such a low database layer.

cache/cache.go
```diff
@@ -0,0 +1,107 @@
...
+
+	// Returns Node for the key, if exists. nil otherwise.
+	Get(key []byte) Node
```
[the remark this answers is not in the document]
p0mvn: It should never be possible for the value to be nil.

Currently, I added the `CONTRACT` to the spec of `Add(node Node)` explaining that `node` must never be nil.

Given the contract, it is impossible to have a nil value

cache/cache.go
```diff
@@ -0,0 +1,107 @@
...
+		c.ll.MoveToFront(e)
+		old := e.Value
+		e.Value = node
+		return old.(Node)
+	}
+
+	elem := c.ll.PushFront(node)
+	c.dict[string(node.GetKey())] = elem
+
+	if c.ll.Len() > c.cacheLimit {
```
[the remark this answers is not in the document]
p0mvn: When I named this `limit`, I meant `maximum`, including in tests. I changed the names to reflect that so that the condition is kept as is.

cache/cache_bench_test.go
```diff
@@ -0,0 +1,67 @@
...
+			keySize:    30,
+		},
+	}
+
+	for name, tc := range testcases {
+		cache := cache.New(tc.cacheLimit)
+		b.Run(name, func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				_ = cache.Add(&testNode{
+					key: randBytes(tc.keySize),
```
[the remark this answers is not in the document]
p0mvn: Went with option 1, thanks

p0mvn:
@robert-zaremba @odeke-em addressed your comments, please take a look when you have a chance

robert-zaremba:
test is still failing

p0mvn:
Hey everyone. This is ready for review. Please take a look when you can

marbar3778 (approved):

PR #525, opened 2022-07-20
perf: use fast unsafe bytes->string convertion

robert-zaremba:
Avoid unnecessary allocation by using fast unsafe bytes -> string conversion

Notes:
+ I didn't add conversions to the tests - I'm planning other PR there to cleanup key / values repetitions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,16 @@
 
 ## Unreleased
 
+### Improvements
+
+- [#525](https://github.com/cosmos/iavl/pull/525) Optimization: use fast unsafe bytes->string convertion.
+
 ## 0.19.0 (July 6, 2022)
 
 ### Breaking Changes
 
 - [\514](https://github.com/cosmos/iavl/pull/514) Downgrade Tendermint to 0.34.x
-- [\500](https://github.com/cosmos/iavl/pull/500) Return errors instead of panicking. 
+- [\500](https://github.com/cosmos/iavl/pull/500) Return errors instead of panicking.
 
 ### Improvements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Improvements
 
-- [#525](https://github.com/cosmos/iavl/pull/525) Optimization: use fast unsafe bytes->string convertion.
+- [#525](https://github.com/cosmos/iavl/pull/525) Optimization: use fast unsafe bytes->string conversion.
 
 ### Bug Fixes
 
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -61,7 +61,8 @@ func New(maxElementCount int) Cache {
 }
 
 func (c *lruCache) Add(node Node) Node {
-	if e, exists := c.dict[ibytes.UnsafeBytesToStr(node.GetKey())]; exists {
+	keyStr := ibytes.UnsafeBytesToStr(node.GetKey())
+	if e, exists := c.dict[keyStr]; exists {
 		c.ll.MoveToFront(e)
 		old := e.Value
 		e.Value = node
@@ -69,11 +70,10 @@ func (c *lruCache) Add(node Node) Node {
 	}
 
 	elem := c.ll.PushFront(node)
-	c.dict[ibytes.UnsafeBytesToStr(node.GetKey())] = elem
+	c.dict[keyStr] = elem
 
 	if c.ll.Len() > c.maxElementCount {
 		oldest := c.ll.Back()
-
 		return c.remove(oldest)
 	}
 	return nil
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -2,6 +2,8 @@ package cache
 
 import (
 	"container/list"
+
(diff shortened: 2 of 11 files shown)


robert-zaremba:
Benchmark:
```
benchmark                                old ns/op     new ns/op     delta
BenchmarkGetNonMembership/fast-16        143368        129285        -9.82%
BenchmarkGetNonMembership/regular-16     10305208      8466036       -17.85%
BenchmarkImmutableAvlTreeMemDB-16        150397        139207        -7.44%

benchmark                             old allocs     new allocs     delta
BenchmarkImmutableAvlTreeMemDB-16     437            392            -10.30%

benchmark                             old bytes     new bytes     delta
BenchmarkImmutableAvlTreeMemDB-16     33513         32100         -4.22%
```

robert-zaremba:
turns out that it's significant optimization

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Improvements
+
+- [#525](https://github.com/cosmos/iavl/pull/525) Optimization: use fast unsafe bytes->string convertion.
```
p0mvn, line 7:
```suggestion
- [#525](https://github.com/cosmos/iavl/pull/525) Optimization: use fast unsafe bytes->string conversion.
```

internal/bytes/string.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"reflect"
+	"unsafe"
+)
+
+// UnsafeStrToBytes uses unsafe to convert string into byte array. Returned bytes
+// must not be altered after this function is called as it will cause a segmentation fault.
+func UnsafeStrToBytes(s string) []byte {
+	var buf []byte
```
p0mvn, line 11: What do you think about the following to avoid copying the slice internals by directly casting the string:
```
func unsafeGetBytes(s string) []byte {
    return (*[0x7fff0000]byte)(unsafe.Pointer(
        (*reflect.StringHeader)(unsafe.Pointer(&s)).Data),
    )[:len(s):len(s)]
}
```
I'm not sure if there are any constraints preventing that but this could be a further optimization

cache/cache.go
```diff
@@ -95,14 +97,14 @@ func (nc *lruCache) Len() int {
 }
 
 func (c *lruCache) Remove(key []byte) Node {
-	if elem, exists := c.dict[string(key)]; exists {
+	if elem, exists := c.dict[ibytes.UnsafeBytesToStr(key)]; exists {
```
p0mvn, line 100: I'm surprised that this is helpful because, from my understanding, copying during `[]byte` to `string` conversion when accessing a map by key should be optimized by the Go compiler.

Sources:
- https://github.com/golang/go/issues/3512
- https://stackoverflow.com/questions/43470284/bytestring-vs-bytestring/43470344#43470344
- https://github.com/golang/go/blob/f5f5a8b6209f84961687d993b93ea0d397f5d5bf/src/cmd/gc/order.c#L943-L953

p0mvn: I'm wondering if the benchmark would remain the same as it is right now if the `map[string]` changes are reverted

p0mvn (commented):
Hi @robert-zaremba . Great work on this optimization. I had a couple of questions. Please let me know what you think

internal/bytes/string.go
robert-zaremba: What is `[0x7fff0000]byte`? Looks like a big static array. Where did you get that solution from?

cache/cache.go
robert-zaremba: I can check, however I for consistency we can keep the casting here.

internal/bytes/string.go
p0mvn: It is the maximum [process address space](https://stackoverflow.com/questions/27675550/what-is-the-address-space-in-golang) value

I grabbed this suggestion from here:
https://stackoverflow.com/questions/59209493/how-to-use-unsafe-get-a-byte-slice-from-a-string-without-memory-copy

Original source: https://groups.google.com/g/golang-nuts/c/Zsfk-VMd_fU/m/O1ru4fO-BgAJ

robert-zaremba:
After removing the fast convertion, there is a slight performance degradation:
* before = with unsafe string in cache.go maps
* after = with normal `string(key)` lookup in cache.go maps.

```
benchmark                             old ns/op     new ns/op     delta
BenchmarkImmutableAvlTreeMemDB-16     138184        140417        +1.62%

benchmark                             old allocs     new allocs     delta
BenchmarkImmutableAvlTreeMemDB-16     392            419            +6.89%

benchmark                             old bytes     new bytes     delta
BenchmarkImmutableAvlTreeMemDB-16     32154         32928         +2.41%
```

I repeated it few times and the memory allocation results were similar (there was more variation in the ns/op)

p0mvn (approved):
LGTM. Thanks for trying the map benchmarks.

Let's go with the approach that benchmarks show is best.

We can investigate [this](https://github.com/cosmos/iavl/pull/525#discussion_r934901067) further separately

PR #531, opened 2022-08-10
upgrade to fast IAVL with multi batches

giskook:
When we auto upgrade our huge (289G) data. We met OOM. so splite to multi batches to upgrade the fast storage.

### DIFF
--- a/mutable_tree.go
+++ b/mutable_tree.go
@@ -15,6 +15,8 @@ import (
 	"github.com/cosmos/iavl/internal/logger"
 )
 
+var commitGap uint64 = 10000000
+
 // ErrVersionDoesNotExist is returned if a requested version does not exist.
 var ErrVersionDoesNotExist = errors.New("version does not exist")
 
@@ -627,12 +629,6 @@ func (tree *MutableTree) IsUpgradeable() (bool, error) {
 // from latest tree.
 // nolint: unparam
 func (tree *MutableTree) enableFastStorageAndCommitIfNotEnabled() (bool, error) {
-	shouldForceUpdate, err := tree.ndb.shouldForceFastStorageUpgrade()
-	if err != nil {
-		return false, err
-	}
-	isFastStorageEnabled := tree.ndb.hasUpgradedToFastStorage()
-
 	isUpgradeable, err := tree.IsUpgradeable()
 	if err != nil {
 		return false, err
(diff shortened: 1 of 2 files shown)


mutable_tree.go
```diff
@@ -15,6 +15,8 @@ import (
 	"github.com/cosmos/iavl/internal/logger"
 )
 
+var commitGap uint64 = 10000000
```
robert-zaremba, line 18: let's add a comment here

mutable_tree.go
```diff
@@ -712,10 +718,15 @@ func (tree *MutableTree) enableFastStorageAndCommit() error {
...
 	itr := NewIterator(nil, nil, true, tree.ImmutableTree)
 	defer itr.Close()
+	var upgradedFastNodes uint64
 	for ; itr.Valid(); itr.Next() {
+		upgradedFastNodes++
 		if err = tree.ndb.SaveFastNodeNoCache(NewFastNode(itr.Key(), itr.Value(), tree.version)); err != nil {
 			return err
 		}
+		if upgradedFastNodes%commitGap == 0 {
+			tree.ndb.Commit()
```
robert-zaremba, line 728: why doing a commit here?

mutable_tree.go
giskook: had add a comment

mutable_tree.go
giskook: when we update `commitGap` fastNodes, we commit it, or when we met a huge IAVL tree, the memory cost will be huge too.

faddat:
very interesting!

robert-zaremba:
@p0mvn you have been in charge for the fast cache - would you have time to review this PR?

p0mvn (approved):
utACK.

Thank you for this change. That makes sense.

As a side note for future work, do you think we can now remove this garbage collection logic:
https://github.com/cosmos/iavl/blob/807f8c542e48369d35622d582f56da5187d128b3/mutable_tree.go#L688-L712

That logic was added to mitigate OOM but I think the solution in this PR might remove the need for the garbage collection.

We suspect that this garbage collection might be increasing the setup time in our unit tests in Osmosis.

ref: https://github.com/osmosis-labs/osmosis/issues/2148#issuecomment-1233453076

giskook:
> utACK.
> 
> Thank you for this change. That makes sense.
> 
> As a side note for future work, do you think we can now remove this garbage collection logic:
> 
> https://github.com/cosmos/iavl/blob/807f8c542e48369d35622d582f56da5187d128b3/mutable_tree.go#L688-L712
> 
> That logic was added to mitigate OOM but I think the solution in this PR might remove the need for the garbage collection.
> 
> We suspect that this garbage collection might be increasing the setup time in our unit tests in Osmosis.
> 
> ref: [osmosis-labs/osmosis#2148 (comment)](https://github.com/osmosis-labs/osmosis/issues/2148#issuecomment-1233453076)

@p0mvn

mutable_tree.go
```diff
@@ -642,19 +639,29 @@ func (tree *MutableTree) enableFastStorageAndCommitIfNotEnabled() (bool, error)
 		return false, nil
 	}
 
-	if isFastStorageEnabled && shouldForceUpdate {
```
marbar3778, line 645: I think we should leave this as some users may not want to upgrade at this time

mutable_tree.go
marbar3778: have you tried different numbers to see results?

giskook: > have you tried different numbers to see results?

not yet. maybe I can use a different value to comare with this. do you have any sugguestion? maybe 5,000,000?

mutable_tree.go
giskook: Once users update the iavl code version, it will auto uggrade to fast IAVL.

marbar3778: there is another PR to make it optional since some dbs may take too long. If a node doesn't upgrade then would this still work?

mutable_tree.go
marbar3778: yea lets try it

mutable_tree.go
giskook: I glanced the pr#547. It seems if `skipFastStorageUpgrade` is true. it will not go here. so we do not need to readd the `isFastStorageEnabled && shouldForceUpdate`

mutable_tree.go
giskook: with 5m, when got lower memory usage and faster speed ;P
image

marbar3778: amazing, lets run with it

giskook: ok, had change the commitGap to 5000000, and remove the timing-gc related code.

marbar3778 (approved):
thank you for this pr

PR #538, opened 2022-08-22
cosmos db

faddat:
This is a draft PR for cosmos db

-- basically it should just wait on cutting a release of cosmos-db


marbar3778 (approved):

marbar3778:
@faddat could you resolve conflicts

Makefile
```diff
@@ -46,18 +46,18 @@ lint:
 # bench is the basic tests that shouldn't crash an aws instance
 bench:
 	cd benchmarks && \
-		go test $(LDFLAGS) -tags cleveldb,rocksdb,boltdb,badgerdb -run=NOTEST -bench=Small . && \
-		go test $(LDFLAGS) -tags cleveldb,rocksdb,boltdb,badgerdb -run=NOTEST -bench=Medium . && \
```
robert-zaremba, line 50: why removing `badgerdb`?

robert-zaremba (commented):
Le'ts keep `badgerdb`

Makefile
faddat: it is no longer in cosmos-db because it's abandoned.

faddat:
* https://github.com/cosmos/iavl/pull/535

It'd be a good idea to merge #535.  The reason that golangci-lint is passing in 3 seconds is that we don't check out the repository in ci and #535 fixes that.

PR #546, opened 2022-09-03
ci: Enable benchmarks in Github Actions

faddat:
Automates one run of the iavl benchmarks, and does so with each commit.

It also removes old benchmark records, that are likely no longer relevant, as well as the setup script for those benchmarks.


faddat:
@robert-zaremba hey, I ended up finding the benchmarking issue that you described. 

Indeed, they do not run "stock".

Since this is an important library, I **think** that these changes should fix benchmarks.  Basically just explicitly installed clang and gcc here:

https://github.com/notional-labs/containers/blob/1bbdb5894ac2d293651a2bffeefa831de0eb8448/cosmos/Dockerfile#L9

We're missing some kind of pthread support header.

marbar3778:
@odeke-em how do we make gobencher work. This should already be handled no?

faddat:
Not sure, but I can check that out

faddat:
Latest commit - since we run the benchmarks in CI, there's not really a need to have the results kept here, ci will keep those for us.

odeke-em:
@marbar3778 yup, bencher already runs with every commit on this repository, please see https://dashboard.bencher.orijtech.com/?repo=cosmos%2Fiavl but seems like something inside iavl ain't properly running

faddat:
@odeke-em hi :D

so, I think that this is .... wierd.  Where could I see the gobencher config?

If I had to guess, it's not being built quite right.

faddat:
```Makefile
bench:
	cd benchmarks && \
		go test $(LDFLAGS) -tags cleveldb,rocksdb,pebbledb -run=NOTEST -bench=Small . && \
		go test $(LDFLAGS) -tags cleveldb,rocksdb,pebbledb -run=NOTEST -bench=Medium . && \
		go test $(LDFLAGS) -run=NOTEST -bench=RandomBytes .
.PHONY: bench
```

in order to run the tests, you'll want to build with the different tags.

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,19 @@
...
+  Benchmarks:
+    runs-on: ubuntu-latest
+    container: ghcr.io/notional-labs/cosmos
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with:
+          go-version: '1.19' # The Go version to download (if necessary) and use.
+      - name: run benchmarks
+        run: make bench
```
marbar3778, line 19: where is this being stored? how do we compare differences?

maybe go bencher is the right approach here

faddat: oh, I think we should use both.  This only stores in github actions. 

I do not know how to configure gobencher, though.

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,19 @@
...
+
+jobs:
+  Benchmarks:
+    runs-on: ubuntu-latest
+    container: ghcr.io/notional-labs/cosmos
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with:
+          go-version: '1.19' # The Go version to download (if necessary) and use.
```
robert-zaremba, line 17: let's use cache.

.github/workflows/benchmarks.yml
robert-zaremba, line 19: do we really want to run benchmarks on CI? What's the goal? It takes lot of cycles. If we don't have performance DIFF reporting, then running these benchmark is burining CI cycles.

robert-zaremba (commented):
let's discuss more how to run benchmarks in CI
* do we want to run them on each PR
* who is paying for the CI cycles
* does it make sense if there is no DIFF reporting

marbar3778:
go bencher is now fixed, we should opt to only use that. no need to duplicate jobs

faddat:
How do I see what gobencher produces? 

This is globally visible

robert-zaremba:
I gobencher works then I prefer much more gobencher approach. It gives us diff, and won't consume github cycles.

faddat:
question:

We are spending on these ci jobs?

Who is spending-- and then also maybe, why are we spending?

Thanks

robert-zaremba:
no limit CI is not for free. I'm not sure if we are currently in a free tier for IAVL. For Cosmos-SDK, we are in a paid plan. But with benchamarks enabled in CI then it will be more complicated.

faddat:
I get unlimited CI for free.

 anyhow no worries and I still have no idea who's paying for it

The real issue at this point is that we need to make gobencher test pebble and rocks and cleveldb not just goleveldb.

marbar3778:
@odeke-em how do we run govencher with other dbs?

tac0turtle:
ah nvm we should actually run this to make sure they pass

tac0turtle (approved):

PR #547, opened 2022-09-07
feat: Add `skipFastStorageUpgrade` to `MutableTree` to control fast storage upgrade

adu-crypto:
```go
type MutableTree struct {
	*ImmutableTree                                     // The current, working tree.
	lastSaved                *ImmutableTree            // The most recently saved tree.
	orphans                  map[string]int64          // Nodes removed by changes to working tree.
	versions                 map[int64]bool            // The previous, saved versions of the tree.
	allRootLoaded            bool                      // Whether all roots are loaded or not(by LazyLoadVersion)
	unsavedFastNodeAdditions map[string]*fastnode.Node // FastNodes that have not yet been saved to disk
	unsavedFastNodeRemovals  map[string]interface{}    // FastNodes that have not yet been removed from disk
	ndb                      *nodeDB
	skipFastStorageUpgrade   bool // If true, the tree will work like no fast storage and always not upgrade fast storage

	mtx sync.Mutex
}
```
Add the `skipFastStorageUpgrade` field to `MutalbeTree`.
If `skipFastStorageUpgrade` is true, neither `Load()`, `LazyLoadVersion()`, `LoadVersion()` nor `LoadVersionForOverwriting` would try to upgrade fast storage when loading the tree.
This is needed for archive nodes with extremely big db to avoid taking unacceptable long time to upgrade the  fast storage

### DIFF
--- a/mutable_tree_test.go
+++ b/mutable_tree_test.go
@@ -30,7 +30,7 @@ var (
 
 func setupMutableTree(t *testing.T) *MutableTree {
 	memDB := db.NewMemDB()
-	tree, err := NewMutableTree(memDB, 0)
+	tree, err := NewMutableTree(memDB, 0, false)
 	require.NoError(t, err)
 	return tree
 }
@@ -183,7 +183,7 @@ func TestMutableTree_LoadVersion_Empty(t *testing.T) {
 
 func TestMutableTree_LazyLoadVersion_Empty(t *testing.T) {
 	memDB := db.NewMemDB()
-	tree, err := NewMutableTree(memDB, 0)
+	tree, err := NewMutableTree(memDB, 0, false)
 	require.NoError(t, err)
 
 	version, err := tree.LazyLoadVersion(0)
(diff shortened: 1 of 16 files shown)


mutable_tree_test.go
```diff
@@ -29,7 +29,7 @@ var (
 
 func setupMutableTree(t *testing.T) *MutableTree {
 	memDB := db.NewMemDB()
-	tree, err := NewMutableTree(memDB, 0)
+	tree, err := NewMutableTree(memDB, 0, false)
```
marbar3778, line 32: we should add a couple tests that has it disabled. talked with dev and he mentioned a few more places are needed. The gets was one thing he pointed to

marbar3778 (commented):
left one comment

mutable_tree_test.go
adu-crypto: ok, I would convert this to draft and continue working on other possible affected parts and adding some tests

adu-crypto: added two integration tests to make sure that the tree will not perform the fast storage upgrade when loading the tree, and the not upgraded tree returns correct key/value with `Get` and `Iterate`.

marbar3778 (approved):
LGTM, we will have to push this to a branch without the cosmos-db changes.

adu-crypto:
> LGTM, we will have to push this to a branch without the cosmos-db changes.

is this not compatible with `cosmos-db`? which branch do you suggest pushing to?

marbar3778:
I can handle this part, will start right when this is merged. Would like to get one more review before then

alexanderbez (approved):

PR #550, opened 2022-09-08
benchmarks: on missing/unlinked DB -tags=cleveldb,rocksdb, just log errors instead of failing

### LINKED ISSUE #549: benchmarks: BenchmarkMedium and BenchmarkSmall fail with "unknown db_backend rocksdb, expected one of goleveldb,memdb" or "unknown db_backend cleveldb, expected one of goleveldb,memdb"
Running benchmarks in this repository requires one to add the tags with runtime tags `-tags=cleveldb` or `-tags=rocksdb` but when running benchmarks with Bencher and just on a local machine, unfortunately that fails and causes lots of interruptions. We need to be able to skip this for benchmarking to proceed, please see https://dashboard.bencher.orijtech.com/?repo=cosmos%2Fiavl which has failures all the way and just burn CPU time and precious cloud computing resources like https://dashboard.bencher.orijtech.com/benchmark/929976c03aa04d1b8bf42b13f8652fe0 or
Screen Shot 2022-09-08 at 1 27 12 PM

Just an FYI for @kirbyquerby @elias-orijtech @marbar3778 @faddat; I am sending a PR to fix this.

odeke-em:
To ensure that benchmarks run smoothly by default even if the tags "-tags=cleveldb,rocksdb" are missing, just log errors that result from missing runtime linked databases.

Fixes #549

### DIFF
--- a/benchmarks/bench_test.go
+++ b/benchmarks/bench_test.go
@@ -5,6 +5,7 @@ import (
 	"math/rand"
 	"os"
 	"runtime"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


odeke-em:
Kindly cc-ing @marbar3778

marbar3778 (approved):

PR #554, opened 2022-09-09
quick linter pass and standardization

faddat:
Before beginning on the data locality work, a quick linter so can always see if there's an issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,38 +32,37 @@ tools:
 .PHONY: tools
 
 format:
-	find . -name '*.go' -type f -not -path "*.git*" -not -name '*.pb.go' -not -name '*pb_test.go' | xargs gofmt -w -s
-	find . -name '*.go' -type f -not -path "*.git*"  -not -name '*.pb.go' -not -name '*pb_test.go' | xargs goimports -format
+	@go install mvdan.cc/gofumpt@latest
+	@go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest
+	find . -name '*.go' -type f -not -path "*.git*" -not -name "*.pb.go" -not -name "*pb_test.go"  | xargs gofumpt -w -l
+	golangci-lint run --fix
 .PHONY: format
 
 # look into .golangci.yml for enabling / disabling linters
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -32,38 +32,37 @@ tools:
...
-		go test $(LDFLAGS) -run=NOTEST -bench=RandomBytes . && \
-		go test $(LDFLAGS) -tags cleveldb,rocksdb,pebbledb -run=NOTEST -bench=Small . && \
-		go test $(LDFLAGS) -tags cleveldb,rocksdb,pebbledb -run=NOTEST -bench=Medium . && \
-		go test $(LDFLAGS) -tags cleveldb,rocksdb,pebbledb -run=NOTEST -timeout=30m -bench=Large . && \
-		go test $(LDFLAGS) -run=NOTEST -bench=Mem . && \
-		go test $(LDFLAGS) -run=NOTEST -timeout=60m -bench=LevelDB .
+		go test $(LDFLAGS) -bench=RandomBytes . && \
+		go test $(LDFLAGS) -tags boltdb,badgerdb,pebbledb -bench=Small . && \
+		go test $(LDFLAGS) -tags boltdb,badgerdb,pebbledb -bench=Medium . && \
+		go test $(LDFLAGS) -tags boltdb,badgerdb,pebbledb -timeout=30m -bench=Large . && \
```
robert-zaremba, line 63: so, are we still using boltdb and badger db?

Makefile
```diff
@@ -32,38 +32,37 @@ tools:
...
+		go test $(LDFLAGS) -bench=Mem . && \
+		go test $(LDFLAGS) -timeout=60m -bench=LevelDB .
```
robert-zaremba, line 65: Why do you remove `-run=NOTEST` flag? it's needed to not start any test. We don't want it to impact benchmarks.

robert-zaremba (commented):
We should keep `-run=NOTEST` flag

Makefile
faddat: no

faddat:
in fact the bottom line here is this is just a bad PR.


Sorry!

PR #558, opened 2022-09-10
ci: add lint PR action

julienrbrt:
Adds lint-pr action. This forces the standardization of PR titles.

### DIFF
--- a/.github/workflows/lint-pr.yml
+++ b/.github/workflows/lint-pr.yml
@@ -1,0 +1,22 @@
+name: "Lint PR"
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+
+permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/lint-pr.yml
```diff
@@ -0,0 +1,22 @@
+name: "Lint PR"
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
```
robert-zaremba, line 8: why do we need "synchronize" here?

robert-zaremba (approved):

.github/workflows/lint-pr.yml
julienrbrt: I actually just copied it from the SDK config.

PR #566, opened 2022-09-15
feat: add iterator concurrency tests

facundomedica:
Adding 3 simple tests that will fail if any new change makes the current situation with concurrency worse. These 3 always fail if the fast node is enabled (can be passed to `setupMutableTree`).

### DIFF
--- a/mutable_tree_test.go
+++ b/mutable_tree_test.go
@@ -39,6 +39,9 @@ func setupMutableTree(t *testing.T, skipFastStorageUpgrade bool) *MutableTree {
 
 // TestIterateConcurrency throws "fatal error: concurrent map writes" when fast node is enabled
 func TestIterateConcurrency(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping test in short mode.")
+	}
 	tree := setupMutableTree(t, true)
 
 	for i := 0; i < 100; i++ {
@@ -56,6 +59,9 @@ func TestIterateConcurrency(t *testing.T) {
 // TestConcurrency throws "fatal error: concurrent map iteration and map write" and
 // also sometimes "fatal error: concurrent map writes" when fast node is enabled
 func TestIteratorConcurrency(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping test in short mode.")
+	}
 	tree := setupMutableTree(t, true)
 
 	tree.LoadVersion(0)
@@ -74,6 +80,9 @@ func TestIteratorConcurrency(t *testing.T) {
 
 // TestNewIteratorConcurrency throws "fatal error: concurrent map writes" when fast node is enabled
 func TestNewIteratorConcurrency(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping test in short mode.")
+	}
 	tree := setupMutableTree(t, true)
 	for i := 0; i < 100; i++ {
 		go func(i int) {
--- a/mutable_tree_test.go
+++ b/mutable_tree_test.go
@@ -7,6 +7,7 @@ import (
 	"runtime"
 	"sort"
 	"strconv"
+	"sync"
 	"testing"
 
 	"github.com/cosmos/iavl/fastnode"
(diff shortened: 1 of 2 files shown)


mutable_tree_test.go
```diff
@@ -29,15 +30,65 @@ var (
...
 	require.NoError(t, err)
 	return tree
 }
 
+// TestIterateConcurrency throws "fatal error: concurrent map writes" when fast node is enable
+func TestIterateConcurrency(t *testing.T) {
+	tree := setupMutableTree(t, true)
+
+	for i := 0; i < 100; i++ {
+		go func() {
```
mmsqe, line 45: May I ask if need use local scope `i` to avoid data race, and [wait](https://github.com/mmsqe/iavl/blob/b1f03a903ac6c907a01a0eee042cdc8beda4c62b/mutable_tree_test.go#L40) until all set finish?

marbar3778 (commented):
should we get a test that shows it failing if by miracle we fix it so we know?

mutable_tree_test.go
```diff
@@ -29,15 +30,65 @@ var (
...
+			for j := 0; j < 1000000; j++ {
+				tree.Set([]byte(fmt.Sprintf("%d%d", i, j)), rand.Bytes(1))
+			}
+		}()
```
robert-zaremba, line 49:
```suggestion
		}(i)
```

mutable_tree_test.go
robert-zaremba, line 45: `i` is changing, so you need to bind it in the function

robert-zaremba:
```suggestion
		go func(i int) {
```

mutable_tree_test.go
```diff
@@ -29,15 +30,65 @@ var (
...
+	tree := setupMutableTree(t, true)
+
+	tree.LoadVersion(0)
+	// So much slower
+	for i := 0; i < 100; i++ {
+		go func() {
+			for j := 0; j < 100000; j++ {
+				tree.Set([]byte(fmt.Sprintf("%d%d", i, j)), rand.Bytes(1))
+			}
+		}()
```
robert-zaremba, line 68: ditto

mutable_tree_test.go
facundomedica: If we wait then the tests will timeout; we just want to throw stuff at it enough that we are sure that it would fail.

facundomedica:
> should we get a test that shows it failing if by miracle we fix it so we know?

I thought this was going to be possible but apparently there's no way to catch a fatal exit :(

Btw, I'm getting data race warnings; which is obvious... I'm not sure what we should do here, remove the -race flag doesn't seem like a sensible option to me

marbar3778:
> > should we get a test that shows it failing if by miracle we fix it so we know?
> 
> I thought this was going to be possible but apparently there's no way to catch a fatal exit :(
> 
> Btw, I'm getting data race warnings; which is obvious... I'm not sure what we should do here, remove the -race flag doesn't seem like a sensible option to me

want to separate them via a build flag and run it separately? Think this may be the cleanest

marbar3778 (approved):
Lgtm

mmsqe (commented):
Test is undetermined without wait goroutine done, just think we could lower iterator since performance will degrade if do concurrent safe, free to [try](https://github.com/mmsqe/iavl/pull/1).

facundomedica:
> Test is undetermined without wait goroutine done, just think we could lower iterator since performance will degrade if do concurrent safe, free to [try](https://github.com/mmsqe/iavl/pull/1).

Sorry, I don't understand what you mean. Could you please rephrase it for me? 🙏 Thank you!

mmsqe:
> Sorry, I don't understand what you mean. Could you please rephrase it for me? 🙏 Thank you!

We might just add this [wait](https://github.com/cosmos/iavl/pull/570/files) for the test.

facundomedica:
@mmsqe lowering the amount of iterations + the wait group seems to ensure that the test fails when the fast cache is enabled. I'll merge your PR :)

Ywmet (commented):
maybe ci should add for test -race

Issue #573, opened 2022-09-27
iaviewer doesn't work


robert-zaremba:
Can't use `iaviewer` with the app db.

```
iaviewer versions application.db/ ""
Error reading data: cannot cut paths on application
```

facundomedica:
I got it to run doing `iaviewer versions ./application.db`. the code is looking for that initial slash

facundomedica:
But yeah, a silly bug anyway

PR #574, opened 2022-09-28
fix: accept both relative and absolute paths in iaviewer

facundomedica:
Closes: #573


robert-zaremba (approved):

robert-zaremba:
Thanks for a quick fix!

marbar3778 (approved):

PR #578, opened 2022-10-06
chore: don't use deprecated linters

faddat:
This PR:

* removes deprecated linters
* does a linting pass


marbar3778 (approved):
LGTM, hopefully returning errors doesn't handle logic differently

PR #586, opened 2022-10-11
fix: remove `RangeProofs`

### LINKED ISSUE #284: Construct ICS proofs directly from tree
Currently, the IC proof is converting a series of range proofs into ICS proof using public methods.

Taking advantage of internal methods of the tree to construct ICS proof should improve efficiency

### LINKED ISSUE #584: Proof: Remove RangeProofs in favour of only ics23
The current proof system of IAVL is hard to understand and led to a recent hack for Binance. 

Since this code is not actively used in the cosmos-sdk, we should not keep it around. We should look at adopting ics23 natively onto the tree and remove the current proof system.

cool-develope:
Closes #584 
Closes #284


cool-develope:
@tac0turtle, this is the initial PR, please review it. If it is OK for you, I will remove the necessary ones.
Benchmark results
``` bash
# original
goos: linux
goarch: amd64
pkg: github.com/cosmos/iavl
cpu: Intel(R) Core(TM) i5-9400 CPU @ 2.90GHz
BenchmarkGetNonMembership/fast-6         	    4908	    214837 ns/op	    8729 B/op	     124 allocs/op
BenchmarkGetNonMembership/regular-6      	     162	   7052272 ns/op	 1337791 B/op	   22250 allocs/op

# refactored
goos: linux
goarch: amd64
pkg: github.com/cosmos/iavl
cpu: Intel(R) Core(TM) i5-9400 CPU @ 2.90GHz
BenchmarkGetNonMembership/fast-6         	    6990	    212281 ns/op	    3709 B/op	      15 allocs/op
BenchmarkGetNonMembership/regular-6      	    6045	    213444 ns/op	    3747 B/op	      16 allocs/op
```

tac0turtle:
cc @AdityaSripal @colin-axner

proof_ics23.go
```diff
@@ -29,79 +29,52 @@ GetNonMembershipProof will produce a CommitmentProof that the given key doesn't
...
+			},
 		},
-	}
-	return proof, nil
+	}, nil
 }
 
-func createExistenceProof(tree *ImmutableTree, key []byte) (*ics23.ExistenceProof, error) {
-	value, proof, err := tree.GetWithProof(key)
+func createExistenceProof(tree *ImmutableTree, key []byte, leftMost bool) (*ics23.ExistenceProof, error) {
```
colin-axner, line 59: Just thinking out loud for code hygiene improvement. I think it could make sense to make `createExistenceProof` a func of `ImmutableTree` so you call `t.createExistenceProof(key)`

I think it could make sense to order this function below `GetMembershipProof`, the functions names make sense, since you either want:
- MembershipProof (one existence proof)
- NonMembershipProof(two existence proofs)

proof_ics23.go
```diff
@@ -29,79 +29,52 @@ GetNonMembershipProof will produce a CommitmentProof that the given key doesn't
...
 // convertExistenceProof will convert the given proof into a valid
 // existence proof, if that's what it is.
 //
 // This is the simplest case of the range proof and we will focus on
 // demoing compatibility here
-func convertExistenceProof(p *RangeProof, key, value []byte) (*ics23.ExistenceProof, error) {
-	if len(p.Leaves) != 1 {
-		return nil, fmt.Errorf("existence proof requires RangeProof to have exactly one leaf")
-	}
+func convertExistenceProof(p PathToLeaf, node *Node) (*ics23.ExistenceProof, error) {
```
colin-axner, line 72: Can probably remove the error return now

proof_ics23.go
```diff
@@ -29,79 +29,52 @@ GetNonMembershipProof will produce a CommitmentProof that the given key doesn't
...
-	}
-	return convertExistenceProof(proof, key, value)
+	return convertExistenceProof(proof, node)
 }
 
 // convertExistenceProof will convert the given proof into a valid
 // existence proof, if that's what it is.
 //
 // This is the simplest case of the range proof and we will focus on
 // demoing compatibility here
```
colin-axner, line 71: Godoc can be updated

proof_ics23.go
colin-axner, line 72: I would recommend renaming `node` to `leafNode` as I think the root node version and leaf node version can be different

colin-axner, line 72: I think it could make sense to remove this function entirely. We should just construct the ics23 existence proof within `createExistenceProof`

proof.go
```diff
@@ -219,26 +222,34 @@ func proofLeafNodeFromProto(pbLeaf *iavlproto.ProofLeafNode) (ProofLeafNode, err
...
+	path := new(PathToLeaf)
+	val, err := node.pathToLeaf(t, key, leftMost, path)
 	return *path, val, err
 }
 
 // pathToLeaf is a helper which recursively constructs the PathToLeaf.
 // As an optimization the already constructed path is passed in as an argument
 // and is shared among recursive calls.
-func (node *Node) pathToLeaf(t *ImmutableTree, key []byte, path *PathToLeaf) (*Node, error) {
+func (node *Node) pathToLeaf(t *ImmutableTree, key []byte, leftMost bool, path *PathToLeaf) (*Node, error) {
```
colin-axner, line 239: what is leftMost doing?

proof.go
```diff
@@ -219,26 +222,34 @@ func proofLeafNodeFromProto(pbLeaf *iavlproto.ProofLeafNode) (ProofLeafNode, err
...
+		return node, ErrKeyNotExist
 	}
 
 	// Note that we do not store the left child in the ProofInnerNode when we're going to add the
 	// left node as part of the path, similarly we don't store the right child info when going down
 	// the right child node. This is done as an optimization since the child info is going to be
 	// already stored in the next ProofInnerNode in PathToLeaf.
-	if bytes.Compare(key, node.key) < 0 {
+	result := bytes.Compare(key, node.key)
+	if (leftMost && result < 0) || (!leftMost && result >= 0) {
```
colin-axner, line 252: This change I don't understand? The code was correct as is (though I think `bytes.Compare(key, node.key) < 0` should be put into a helper function similar to [this](https://github.com/confio/ics23/blob/master/go/ics23.go#L169)

ie, in this if statement, we want to indicate that key is on the left side, so we construct the hash of the right side for the proof

colin-axner (commented):
Thanks for cleaning up the code! I don't understand the `leftMost` addition, it seems unnecessary and potentially incorrect? Could you explain that to me? I haven't looked into the non membership proof code yet

proof.go
cool-develope: OK, I reverted it

tac0turtle:
could we remove the proto messages https://github.com/cosmos/iavl/blob/master/proto/iavl/proof.proto for things that aren't needed

proof_ics23.go
```diff
@@ -166,3 +178,38 @@ func convertVarIntToBytes(orig int64, buf [binary.MaxVarintLen64]byte) []byte {
 	n := binary.PutVarint(buf[:], orig)
 	return buf[:n]
 }
+
+// GetWithProof gets the proof for the given key.
+func (t *ImmutableTree) GetWithProof(key []byte) (*ics23.CommitmentProof, error) {
```
colin-axner, line 183: Maybe rename to `GetProof`?

proof_ics23.go
```diff
@@ -166,3 +178,38 @@ func convertVarIntToBytes(orig int64, buf [binary.MaxVarintLen64]byte) []byte {
...
+		return nil, fmt.Errorf("cannot generate the proof with nil root")
+	}
+	value, err := t.Get(key)
+	if err != nil {
+		return nil, err
+	}
+	if value != nil {
+		return t.GetMembershipProof(key)
+	}
+	return t.GetNonMembershipProof(key)
```
colin-axner, line 194: Is this correct? I think a non membership proof is about the key not existing, not the value. If the key doesn't exist, we will return an err on 189 instead of a non membership proof

colin-axner (commented):
LGTM see minor comments, mostly just reviewed the ics23 changes. It might make sense to split this pr up? It seems to me you could have a pr fixing up the ics23 file and then a followup pr which removes all the unnecessary range proof stuff

I'm just a fan of splitting up prs since it is easier to feel confident approving less code diffs at a time

Also would appreciate a few approvals on these changes

proof_ics23.go
cool-develope: make sense!

proof_ics23.go
cool-develope: OK, I added `isExist` param to proof of both types (existence and non-existence).

proof_ics23.go
```diff
@@ -179,16 +179,12 @@ func convertVarIntToBytes(orig int64, buf [binary.MaxVarintLen64]byte) []byte {
...
+func (t *ImmutableTree) GetProof(key []byte, isExist bool) (*ics23.CommitmentProof, error) {
 	if t.root == nil {
 		return nil, fmt.Errorf("cannot generate the proof with nil root")
 	}
-	value, err := t.Get(key)
-	if err != nil {
-		return nil, err
-	}
-	if value != nil {
+	if isExist {
```
colin-axner, line 187: sorry for not being more specific, I think you can use the [`Has`](https://github.com/cosmos/iavl/blob/master/immutable_tree.go#L143) function

```go
exist, err := t.Has(key)
if err != nil {
    return nil, err
}

if exist {
    return t.GetMembershipProof(key)
}

return t.GetNonMembershipProof(key)
```

proof_test.go
```diff
@@ -6,69 +6,47 @@ import (
...
 
-	proto "github.com/cosmos/gogoproto/proto"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	iavlrand "github.com/cosmos/iavl/internal/rand"
-	iavlproto "github.com/cosmos/iavl/proto"
 )
 
 func TestTreeGetWithProof(t *testing.T) {
```
colin-axner, line 15:
```suggestion
func TestTreeGetProof(t *testing.T) {
```

colin-axner (approved):
Nice work! Love to see this cleanup! Would appreciate at least another approval from someone familiar with ics23

proof_ics23.go
colin-axner: I guess when the SDK queries for this value, it should be performing the `Has` query then?

cool-develope:
> Nice work! Love to see this cleanup! Would appreciate at least another approval from someone familiar with ics23

cc: @tac0turtle

proof_ics23.go
```diff
@@ -166,3 +178,34 @@ func convertVarIntToBytes(orig int64, buf [binary.MaxVarintLen64]byte) []byte {
 	n := binary.PutVarint(buf[:], orig)
 	return buf[:n]
 }
+
+// GetProof gets the proof for the given key.
+func (t *ImmutableTree) GetProof(key []byte, isExist bool) (*ics23.CommitmentProof, error) {
```
AdityaSripal, line 183: Seems weird to me that we would expose the API to take in an `isExist` bool.

Doesn't it make more sense to just check if key exists in the tree.

If it does, return an existence proof otherwise return a nonexistence proof

AdityaSripal (approved):
Changes look good. I have one remaining question about the API that can be addressed in subsequent PR.

Thanks for the cleanup!

proof_ics23.go
AdityaSripal: Seems like the correct fix is:

```go
value, err := t.Get(key)
if value != nil {
    return t.GetMembershipProof(key)
}
return t.GetNonMembershipProof(key)
```

cool-develope:
OK, I removed the `isExist` param in GetProof, because we have already `GetMembershipProof` and `GetNonMembershipProof`.
cc: @AdityaSripal @colin-axner

PR #589, opened 2022-10-14
refactor: disable remaining fast node logic in mutable tree

p0mvn:
Skip fast node removals when fast node is disabled to avoid any possibility of those removals being read, leading to inconsistent state between nodes that have fast nodes and the nodes that do not.

### DIFF
--- a/mutable_tree.go
+++ b/mutable_tree.go
@@ -194,13 +194,11 @@ func (tree *MutableTree) Iterate(fn func(key []byte, value []byte) bool) (stoppe
 		return tree.ImmutableTree.Iterate(fn)
 	}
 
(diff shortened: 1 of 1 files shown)


faddat:
thank you from Notional and evmos

tac0turtle (approved):

yihuang:
> leading to inconsistent state between nodes that have fast nodes and the nodes that do not.

does this cause observable inconsistency, like app-hash mismatch, or just a preventive change?

p0mvn:
> > leading to inconsistent state between nodes that have fast nodes and the nodes that do not.
> 
> does this cause observable inconsistency, like app-hash mismatch, or just a preventive change?

Just a preventive change

Issue #592, opened 2022-10-18
Modify the key format of node in DB to a sequential integer ID instead of the hash value.


cool-develope:
## Context
As we discussed in #571, I am introducing the `nonce` mechanism here.

- [x] add `nodeKey` field to replace the key of a node in DB
- [x] replace  a key in DB  from `node.hash` to `nodeKey` 
- [ ] investigate the possibility of removing `version` in the node structure

## Design Decision
- immutable tree structure
```go
type ImmutableTree struct {
	root                   *Node
	ndb                    *nodeDB
	version                int64
	nonce                  int64             // new field to track the current ID
	skipFastStorageUpgrade bool
}
```

- node structure
```go
type Node struct {
	key           []byte
	value         []byte
	hash          []byte

	leftHash      []byte        
	rightHash     []byte    

	nodeKey       int64         // new field which reflects the current node key
	leftNodeKey   int64       // new field
	rightNodeKey  int64      // new field

	version       int64           // possible to remove ???

	size          int64
	leftNode      *Node
	rightNode     *Node
	subtreeHeight int8
	persisted     bool
}
```

cool-develope:
> Before the final merge we should have an adr on the chnage

Yeah, I will prepare the adr when finishing all subtasks.

catShaark:
I don't think we should remove leftHash and rightHash. 
How can I collab with you on this ? I already had the code with the changes mentions here, I've modified methods like balance, delete, save branch ..... so that their logic matches with the new node struct, node saving mechanism.

Issue #137, opened 2019-05-13
Use immutable key for nodes to avoid excessive LevelDB reorgs


mdyring:
Follow-up of https://github.com/cosmos/cosmos-sdk/issues/2131.

After discussion on Slack, a suggestion to implement a fix:

1) For each new node in IAVL tree, assign an sequential integer ID. This ID never changes for the node.
2) Use this integer ID as "key" in the <key, value> pair stored in LevelDB.
3) The value part stored in LevelDB contains the node hash (existing LevelDB key).
4) All parent/child references must be through integer IDs.
4) A special "root" key stored in LevelDB has an value that contains the integer ID of the root node.

mossid:
Implementation suggestion:

I think using the key(or the hash of the key)will be better, sequentially assigned integer key cannot be obtained from the node solely and have to be stored separately.

Also we have to separate the hashbytes and keybytes of the nodes. Currently, the full hash of the node is used both for calculating root hash and leveldb key. We can maintain the current format of the hash, only change the leveldb key format to something immutable.
 
I suggest `hash(node.key)|bigendian(node.version)` format. The `node.version` part will be empty in case of the mutable tree, and be filled in case of the immutable key. This will preserve the backward compatibility(since the root hash calculation is not changed) and compatibility between the nodes using immutable/mutable versions of IAVL(since the transition between them can be done simply emptying/filling the `node.version`).

ethanfrey:
One issue is that we can never overwrite nodes as long as we want to maintain history.
This is why iavl uses the unique hashes (including version) for the keys.

If history can be sacrificed, we can definitely make major speed-ups. But I think it is very important to allow some historical queries for consistent reads on fast-moving block times. On the other hand, we may be able to use a persistence layer that handles multiple snapshots very efficiently, and then avoid this orphan and garbage collection overhead.

ethanfrey:
Are there concrete performance numbers (benchmarks and pprof) that show this root issue

ebuchman:
> I suggest `hash(node.key)|bigendian(node.version)` format.

This sounds reasonable. Fully derived from the node, unique, immutable. The `node.key` will always be the same (right?) but the `node.version` will change any time the value does, so the nodes are still immutable. 

Here's some related info not about IAVL but Tendermint DB performance.

For one, Tendermint seems to slow down over time: https://github.com/tendermint/tendermint/issues/1835

[...]

PR #593, opened 2022-10-18
feat: add nodeKey integer field

cool-develope:
ref: #592
ref: #137


tac0turtle (approved):
utACK

cool-develope:
> utACK

I think the tests are already covered by the original tests because we are triggering errors when decode and encode the node.

tac0turtle:
sorry, meant I didnt pull and test it.

Issue #598, opened 2022-10-27
The immutableTree clone() func is missing skipFastStorageUpgrade field --> wrong Block.Header.AppHash. error in the blockchain


dungtran8tiki:
When we upgrade iavl in our blockchain from iavl v0.19.0 to v0.19.3, the chain got the error: "wrong Block.Header.AppHash.". 

Deep dive the iavl source code, we found that the immutable tree clone() func missing the skipFastStorageUpgrade field. This cause the "skipFastStorageUpgrade" field auto set to false, while it is initiated "true" --> immutable iterator() func return the wrong list, then mismatch in db of validators.

To fix this issue, we need to add "skipFastStorageUpgrade" into the clone() func as follows:
```
func (t *ImmutableTree) clone() *ImmutableTree {
	return &ImmutableTree{
		root:                   t.root,
		ndb:                    t.ndb,
		version:                t.version,
		skipFastStorageUpgrade: t.skipFastStorageUpgrade,
	}
}
```
@tac0turtle please give your comment

tac0turtle:
this sounds good to me, if you submit a pr I can test on chains

tac0turtle:
the main issue people seem to be running into is when the fast nodes exist and then turn it off. Ill test this on some chains, but this could elude to a different issue

dungtran8tiki:
> the main issue people seem to be running into is when the fast nodes exist and then turn it off. 

It's true, actually we upgrade cosmos-sdk v0.45.7 to v0.45.9, it sets "const iavlDisablefastNodeDefault = true" by default, other blockchains set "const iavlDisablefastNodeDefault = false" then they don't face this issue.

PR #599, opened 2022-10-27
fix: add field to struct

tac0turtle:
closes: #598


alexanderbez (approved):

tac0turtle:
tested this fixes the issue we ran into

yihuang:
> tested this fixes the issue we ran into

is it to fix an app hash mismatch issue?

tac0turtle:
yes

PR #604, opened 2022-10-30
chore: remove unneeded deps

tac0turtle:
remove final tm dep form iavl


Makefile
```diff
@@ -93,46 +93,3 @@ exploremem:
...
-
-proto-lint:
-	@$(DOCKER_BUF) check lint --error-format=json
-.PHONY: proto-lint
-
-proto-check-breaking:
-	@$(DOCKER_BUF) check breaking --against-input $(HTTPS_GIT)#branch=master
-.PHONY: proto-check-breaking
-
-proto-gen:
```
julienrbrt, line 135: Do you intend to remove this or is it a mistake?

Makefile
```diff
@@ -93,46 +93,3 @@ exploremem:
...
-	rm -f $(CERTSTRAP) $(PROTOBUF) $(GOX) $(GOODMAN)
-	rm -rf /usr/local/include/google/protobuf
-	rm -f /usr/local/bin/protoc
-.PHONY: tooks-clean
-
-###
-# Non Go tools
-###
-
-protoVer=0.10.0
```
julienrbrt, line 122: I feel like this is needed (but it could be upgraded to 0.11.0)

Makefile
tac0turtle: there is no more proto in this repo.

julienrbrt: Ah ok 👍 I think we can remove the buf config then right?

julienrbrt (approved):
lgtm, pending https://github.com/cosmos/iavl/pull/604#discussion_r1008846603

PR #597, opened 2022-10-25
feat: replace the node key

cool-develope:
ref: #592

PR #608, opened 2022-11-01
feat: add adr-001 for node key refactoring

cool-develope:
ref: #592, #593, #597


docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,81 @@
...
+	unsavedFastNodeAdditions map[string]*fastnode.Node
+	unsavedFastNodeRemovals  map[string]interface{}   
+	ndb                      *nodeDB
+	skipFastStorageUpgrade   bool 
+
+	mtx sync.Mutex
+}
+```
+
+## Consequences
```
tac0turtle, line 63: we should touch on migration paths here. I think this is the biggest question on how to upgrade existing nodes

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,81 @@
...
+- Remove the `version` field from the node structure.
+- Remove the `orphans` from the tree.
+
+New node structure
+
+```go
+type Node struct {
+	key           []byte
+	value         []byte
+	hash          []byte
```
tac0turtle, line 29: are hash and key here the same value? a hash of the value?

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,81 @@
...
+
+Using the sequenced integer ID, we take advantage of data locality in the bTree and it leads to performance improvements.
+
+Removing orphans also provides performance improvements including memory and storage saving. Also, it makes it easy to rollback the tree. Because we will keep the sequenced segment IDs for the specific version, and we can remove all nodes for which the `nodeKey` is greater than the specified integer value.
+
+### Negative
+
+It requires extra storage to store the node because it should keep `leftNodeKey` and `rightNodeKey` to iterate the tree. Instead, we can delete the `version` in the node and reduce the key size.
+
+It can't delete the old nodes for the specific version due to removing orphans. But it makes `rollback` easier and it makes it possible to remove old nodes through `import` and `export` functionalities. The `export` will restruct the tree to make node IDs to a sequenced segment like (1 ... node_sieze).
```
tac0turtle, line 75: could you elaborate on cant delete old nodes? how will this effect the current pruning behaviour used in the sdk. 

using import and export to remove old nodes seems like the wrong approach. 

if a user wants to prune every 500 versions every 1000 blocks it seems this would not be supported based on the above framing.

docs/architecture/adr-001-node-key-refactoring.md
cool-develope: OK, added the migration part

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,96 @@
...
+
+Removing orphans also provides performance improvements including memory and storage saving. Also, it makes it easy to rollback the tree. Because we will keep the sequenced segment IDs for the specific version, and we can remove all nodes for which the `nodeKey` is greater than the specified integer value.
+
+### Negative
+
+It requires extra storage to store the node because it should keep `leftNodeKey` and `rightNodeKey` to iterate the tree. Instead, we can delete the `version`, `leftHash`, and `rightHash` fields in the node and reduce the key size.
+
+It can't delete the old nodes for the specific version due to removing orphans. We introduce a new way to prune old versions.
+
+For example, when a user wants to prune the previous 500 versions every 1000 blocks
```
cool-develope, line 87: @tac0turtle, I added the explanation of how to prune old nodes.

docs/architecture/adr-001-node-key-refactoring.md
cool-develope: here, the `hash` is the hash value of node.

tac0turtle:
a good question by celestia, how does this effect the commitment structure? Are proofs still the same?

cool-develope:
> a good question by celestia, how does this effect the commitment structure? Are proofs still the same?

it doesn't affect, as you can see in #597 it doesn't update any proof part and still passing tests

tzdybal:
> > a good question by celestia, how does this effect the commitment structure? Are proofs still the same?
> 
> it doesn't affect, as you can see in #597 it doesn't update any proof part and still passing tests

#597 still contains `version` field, which is still hashed (see: https://github.com/cosmos/iavl/blob/36a150eddd8f8bd7c6b93bcccadfb4391fa48ea1/node.go#L364). Removing `version` will affect all the hashes, and therefore migration will invalidate all existing proofs.

cool-develope:
> Removing version will affect all the hashes, and therefore migration will invalidate all existing proofs.
@tzdybal , good catch. @tac0turtle how about your though? Will we keep the version or refactor the ics23 part?

tac0turtle:
> @tzdybal , good catch. @tac0turtle how about your thougt? Will we keep the version or refactor the ics23 part?

what do you mean by refactor ics23?

So there are two paths, introduce this in a breaking release of the sdk which a coordinated upgraded is needed or leave it alone. for testing purposes and data collection we probably want to leave it but in the future we could break it. BUT we should check with IBC about this.

cool-develope:
> what do you mean by refactor ics23?

I mean update the validation part in ics23. 

> So there are two paths, introduce this in a breaking release of the sdk which a coordinated upgraded is needed or leave it alone. for testing purposes and data collection we probably want to leave it but in the future we could break it. BUT we should check with IBC about this.

Sounds good, then let's keep the version at this time. I will create the issue and we can revisit in the future

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,96 @@
...
+	unsavedFastNodeAdditions map[string]*fastnode.Node
+	unsavedFastNodeRemovals  map[string]interface{}   
+	ndb                      *nodeDB
+	skipFastStorageUpgrade   bool 
+
+	mtx sync.Mutex
+}
+```
+
+## Consequences
```
tac0turtle, line 64: nit: sorry consequences should come after migration and then we can add a pruning section talking about what you touched on below

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Migration
+
+We can migrate nodes one by one by iterating the version.
+
+- Iterate the version in order, and get the root node for the specific version.
```
tac0turtle, line 70: this may need a larger discussion but should it be lazily or stop the node migrate then restart? Any thoughts on that?

cool-develope: Re-assign should be done after stopping the node. Maybe we can delete only nodes which not appear in iterating. It can be done without breaking off. I will introduce both ways.

tac0turtle: I can grab some snapshots to test this migration and the time it would take. It's gonna be a long one.

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Status
+
+Proposed
+
+## Context
+
+The original node key of IAVL is a hash of the node and it does not take advantage of data locality on disk. The nodes are stored in a random location of the disk due to the random hash value, so it needs to do a random search of the disk to find the node.
+
+The `orphans` are used to manage the removed nodes in the current version and allow to deletion of the removed nodes for the specific version from the disk through the `DeleteVersion`. It needs to track every time when updating the tree and also requires extra storage to store `orphans`, but there are not many use cases of `DeleteVersion`. There are two use cases, the first one is the rollback of the tree and the second one is to remove the unnecessary old nodes.
```
alexanderbez, line 15:
```suggestion
The `orphans` are used to manage node removal in the current design and allow deletion of removed nodes for the specific version from the disk through the `DeleteVersion` API. It needs to track every time when updating the tree and also requires extra storage to store `orphans`, but there are not many use cases of `DeleteVersion`. There are two use cases:

1. Rollback of the tree to a previous version
2. Remove unnecessary old nodes
```

docs/architecture/adr-001-node-key-refactoring.md
cool-develope: Sorry, I was confused with pruning!
I think lazy loading is possible but we should restart the node to switch database after loading, right?

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,96 @@
...
+```go
+type Node struct {
+	key           []byte
+	value         []byte
+	hash          []byte    // keep this field in the storage
+	leftHash      []byte    // will remove
+	rightHash     []byte    // will remove
+	nodeKey       int64     // new field, use as a node key
+	leftNodeKey   int64     // new field, need to store in the storage
+	rightNodeKey  int64     // new field, need to store in the storage
```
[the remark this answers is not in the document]
cool-develope: `leftNode` and `rightNode` is only meaningful on memory side, we need these fields to get children from the storage side.

docs/architecture/adr-001-node-key-refactoring.md
tac0turtle: hmm, maybe lazy migration is not possible. cause it needs to start from the lowest version in the db then move upwards, right?

aaronc (commented):
I'm unclear as to how an integer improves data locality. Can you explain this a bit more? It would seem that it puts nodes that are created at similar times near each other in the B-tree, but this doesn't actually improve things like range scans other than by reducing the size of the node key which is significant.

docs/architecture/adr-001-node-key-refactoring.md
cool-develope: I think it is possible, but this should be done at the top level, not within iavl.
For example, we load it to the new IAVL lazily through `Export` of the original version and the new `Import` functionality, after syncing and block confirmation, we can switch the store to the new version.

tac0turtle: yes that also works. not sure how that would work with go.mods though unless we call this v1 so we can import both?

cool-develope:
> I'm unclear as to how an integer improves data locality. Can you explain this a bit more? It would seem that it puts nodes that are created at similar times near each other in the B-tree, but this doesn't actually improve things like range scans other than by reducing the size of the node key which is significant.

Reducing key size itself improves data locality. I am not sure if leveldb uses compressed trie or general bTree, but we can reduce the child count and improve the tree density.

aaronc:
> > I'm unclear as to how an integer improves data locality. Can you explain this a bit more? It would seem that it puts nodes that are created at similar times near each other in the B-tree, but this doesn't actually improve things like range scans other than by reducing the size of the node key which is significant.
> 
> 
> 
> Reducing key size itself improves data locality. I am not sure if leveldb uses compressed trie or general bTree, but we can reduce the node count and improve the tree density.

I'm pretty sure it's an LSM tree

cool-develope:
> I'm pretty sure it's an LSM tree

It will improve the performance to find the key in the LSM tree, because each chunk file will contain the continuous section and will find the exact level of file which contains the specific key (integer), right?

yihuang:
Keeping the key order the same as insertion order should be great for lsm tree compaction, the merge sort would be simply concatenation.

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,101 @@
...
+For example, when a user wants to prune the previous 500 versions every 1000 blocks
+- We assume the pruning is completed for `n`th version and the last nonce of `n`th version is `x`.
+- We iterate the tree from the `n+501`th root node and pick only nodes which the nodeKey is in `[(n+1)th version first nonce, (n+500)th version the last nonce]`.
+- We can delete missing nodes instantly or re-assign the nodeKey from `x+1` in order for those nodes. Re-assign should be done after stopping the node but it can lead to improving the data locality.
+
+## Consequences
+
+### Positive
+
+Using the sequenced integer ID, we take advantage of data locality in the bTree and it leads to performance improvements. Also, it can reduce the node size in the storage.
```
yihuang, line 87: Nowadays, all the "mainstream" db backends are lsm tree, but it should help there too.

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,101 @@
...
+The original key format of IAVL nodes is a hash of the node. It does not take advantage of data locality on disk. Nodes are stored in random locations on disk due to the random hash value, so it needs to scan the disk to find the corresponding node which can be very inefficient.
+
+The `orphans` are used to manage node removal in the current design and allow deletion of removed nodes for the specific version from the disk through the `DeleteVersion` API. It needs to track every time when updating the tree and also requires extra storage to store `orphans`, but there are not many use cases of `DeleteVersion`. There are two use cases:
+
+1. Rollback of the tree to a previous version
+2. Remove unnecessary old nodes
+
+## Decision
+
+- Use the sequenced integer ID as a node key like `bigendian(nodeKey)` format.
```
yihuang, line 22: How is nodeKey computed, I remember it's sth like "version+path"?

cool-develope: no, it is a just sequenced integer. 
for example, we assign the `nodeKey` of the new node as `tree.nonce + 1` every time when create node

yihuang: It seems better to use "version/seq", where seq is only unique inside the version.

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,101 @@
...
+We will implement the `Import` functionality for the original version.
+
+### Pruning
+
+We introduce a new way to prune old versions.
+
+For example, when a user wants to prune the previous 500 versions every 1000 blocks
+- We assume the pruning is completed for `n`th version and the last nonce of `n`th version is `x`.
+- We iterate the tree from the `n+501`th root node and pick only nodes which the nodeKey is in `[(n+1)th version first nonce, (n+500)th version the last nonce]`.
+- We can delete missing nodes instantly or re-assign the nodeKey from `x+1` in order for those nodes. Re-assign should be done after stopping the node but it can lead to improving the data locality.
```
yihuang, line 81: I don't follow how pruning works, how do we know the node is not referenced by later versions so it's safe to delete?

aaronc: I was wondering that as well, but not sure if I just don't understand

docs/architecture/adr-001-node-key-refactoring.md
cool-develope: here, the `nodeKey` is unique globally.

docs/architecture/adr-001-node-key-refactoring.md
cool-develope: we traverse all nodes based on `n+501`th root, and pick the nodes which the nodeKey is greater than `x`. 
then sort those nodes by nodeKey and can find the missing nodes.

for example, x = 500, nodes= [501, 502, 504, 506 ...], then we can assume 503 and 505 nodes are missed, right?

docs/architecture/adr-001-node-key-refactoring.md
yihuang: "version/seq" is globally unique as well, seq itself is local unique with the version. For each version, the nodes are written in a batch anyway, we only need to maintain the seq in memory.

docs/architecture/adr-001-node-key-refactoring.md
yihuang: I see, use the n+501 tree as the gc root, and delete the missing ones. If a node is not referenced by n+501, it won't be referenced by later versions as well.

cool-develope: exactly

docs/architecture/adr-001-node-key-refactoring.md
cool-develope: we keep `version => (root node key = first nonce, last nonce)`, and ensure the nodeKey is unique globally

yihuang: I know, I was just suggesting to use "version/local-seq" as the nodeKey, so we don't need to store this nonce mapping at all.

cool-develope: I know there are some advantages of `version|seq`, I am just worried it means we need to update `leftNodeKey` and `rightNodeKey` following this update and it leads to extra encode/decode executions and finally requires more storage to store the node data

yihuang: If we constrain both version and seq to be 32bits, we can squeeze them into 8bytes :)

yihuang: Another thing is "version|seq" can utilize dupsort feature of some db backends.

https://lmdb.readthedocs.io/en/release/

docs/architecture/adr-001-node-key-refactoring.md
aaronc: I think this needs to be made a lot clearer and I'm still having trouble understanding how this prunes all the nodes. How do we know that nodes <= x are all pruned properly? Can't nodes less than x still be referenced by later versions of the tree and be eligible for pruning after the nth version has been pruned?

Also, generally this seems like a somewhat roundabout way to do pruning that could become inefficient with large trees because it requires traversing the whole tree and then storing nodes in memory in a sorted list to find out which nodes in the sequence are missing right?

Why don't we just do reference counting?

docs/architecture/adr-001-node-key-refactoring.md
aaronc: > I know there are some advantages of version|seq, I am just worried it means we need to update leftNodeKey and rightNodeKey following this update and it leads to extra encode/decode executions and finally requires more storage to store the node data

Why do we need to update the left and right node keys?

cool-develope: here, `leftNodeKey` and `rightNodeKey` refer to the children node key in the db. If we update the node key as `version|seq`, we should also update left and right node keys, right?

aaronc: Meaning that every time we have a new version we update all referenced node keys to refer to that version rather than the original version where they were created? That seems really complex. I feel like I'm maybe not understanding something really basic here

docs/architecture/adr-001-node-key-refactoring.md
aaronc: In my mind reference counting GC would make it simpler. It wouldn't require traversing all the nodes and would use very little memory. Has anyone explored that?

cool-develope: The main idea of this story is to remove orphans, for me keeping orphans every time when updating the tree is very expensive. I'd like to separate pruning with main CRUD operations.

aaronc: > I guess the cost of reference counting on disk is non-trivial though.

It's just an integer. I think we wouldn't know without benchmarks. Maybe we can explore it in the future.

Anyway, removing orphans makes sense.

JayT106: traversing the whole tree is time-consuming. Shouldn't we try to avoid that?

cool-develope:
@tac0turtle @ValarDragon 
We have two options `version | local nonce` and `global nonce` for the node key format. What about your thoughts?

docs/architecture/adr-001-node-key-refactoring.md
tac0turtle: pruning is handled from the application now, if we can keep the same pruning style, I understand there will need to be minor changes, then it would be good. Later on we should explore native pruning to avoid  write

docs/architecture/adr-001-node-key-refactoring.md
cool-develope: @yihuang , I agree with you. If we keep the local nonce as `int32`, there is not much increase in storage size because we can remove the `version` from node body writes.

docs/architecture/adr-001-node-key-refactoring.md
cool-develope: https://github.com/golang/go/wiki/Modules#version-selection
Maybe, we should create v2 of iavl.

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,101 @@
...
+
+New node structure
+
+```go
+type Node struct {
+	key           []byte
+	value         []byte
+	hash          []byte    // keep this field in the storage
+	nodeKey       int64     // new field, use as a node key
+	leftNodeKey   int64     // new field, need to store in the storage
```
p0mvn, line 36: If we are using nonce as `nodeKey`, how are we getting data locality when traversing or writing the tree to disk?

I thought that the whole idea with the path encoded as a key was that writes would be sequential, contributing to lower frequency of compactions.

yihuang: Without pruning, the nodes are all immutable, the `nodeKey` is sequential, so all the new nodes are written sequentially.

docs/architecture/adr-001-node-key-refactoring.md
yihuang: There are several places storage size would change, assuming version is `int64`:
- node key, 8bytes vs 8+4bytes, but with prefix compression, I think the real difference is minimal.
- node body, we can remove the version field, save a `varint(version)`.
- `leftNodeKey` and `rightNodeKey`, `varint(global nonce)` vs `varint(version << 32 + local nonce)`, there'll be a few more bytes here.

yihuang:
If we come out of the box, a simple file format like this would work?

```
# filename: block-{N}
version: Int
offsets: [Int]
nodes: [Node]
```

`offsets` and `nodes` are continuous array.
node is identified by `(version, seq)`, `seq` is the index of the `offsets` and `nodes` arraies.

docs/architecture/adr-001-node-key-refactoring.md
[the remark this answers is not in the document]
cool-develope: OK, I will update this part and add more description of LSM

docs/architecture/adr-001-node-key-refactoring.md
cool-develope: I know the tree path describes the data locality well. The problem is how to keep the path when rotating the tree.

yihuang: we need to split the `leftNodeKey` into two fields, `leftNodeVersion` and `leftNodeNonce`, right?

yihuang: Aren't we need to split the `leftNodeKey` into `leftNodeVersion` and `leftNodeNonce`, likewise for `rightNodeKey`.

yihuang (changes requested):
I think we need to split `leftNodeKey` into `leftNodeVersion` and `leftNodeNonce`, LGTM other than that

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,99 @@
...
+- Remove the `leftHash` and `rightHash` fields, and instead store `hash` field.
+- Remove the `orphans` from the tree.
+
+New node structure
+
+```go
+type Node struct {
+	key           []byte
+	value         []byte
+	hash          []byte    // keep this field in the storage
```
yihuang, line 33: I guess one consequence is proof generation is slower, previously we can get child hash directly, now we need to load the node?

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,99 @@
...
+	mtx sync.Mutex
+}
+```
+
+### Migration
+
+We can migrate nodes one by one by iterating the version.
+
+- Iterate the version in order, and get the root node for the specific version.
+- Iterate the tree and assign the `nodeKey` to nodes which the node version equals.
```
yihuang, line 69: Migrate a archive node and pruned node will end up with different node keys, but it's probably ok.

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,99 @@
...
+	unsavedFastNodeAdditions map[string]*fastnode.Node
+	unsavedFastNodeRemovals  map[string]interface{}   
+	ndb                      *nodeDB
+	skipFastStorageUpgrade   bool 
+
+	mtx sync.Mutex
+}
+```
+
+### Migration
```
yihuang, line 64: An alternative way to migrate is using the state streamer, recreate the new iavl db with the state changes.

docs/architecture/adr-001-node-key-refactoring.md
yihuang: Inconsistent node nonces could make the state-sync snapshot inconsistent as well?

tac0turtle: On the global nonce they would be different but version | local nonce should be same, right?

yihuang: Hmm, in migration, it depends on the order traversal, this is easy to specify. but when handling the insertion, how do we specify the logic, to allow alternative implementations be compatible with each other.

yihuang:
> we can remove the root store (version => root key)

I think we'll still need the root store, in case there's no change in some versions, where the root node is not updated, we need a root store to reference to the old root node?

docs/architecture/adr-001-node-key-refactoring.md
cool-develope: right, I updated the node struct

docs/architecture/adr-001-node-key-refactoring.md
cool-develope: We are using the ics23 proof, which requires size, version, and height so I believe it doesn't lead to any delays.

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Positive
+
+Using the version and the sequenced integer ID, we take advantage of data locality in the LSM tree. Since we commit the sorted data, it can reduce compactions and makes easy to find the key. Also, it can reduce the key and node size in the storage.
+
+Removing orphans also provides performance improvements including memory and storage saving. Also, it makes it easy to rollback the tree. Because we can remove all nodes for which the `nodeKey` is greater than the specified version.
+
+### Negative
+
+It requires extra storage to store the node because it should keep `leftNodeKey` and `rightNodeKey` to iterate the tree. Instead, we can delete`leftHash` and `rightHash` fields in the node and reduce the key size.
```
tac0turtle, line 96: could we add how much extra per key would be added?

docs/architecture/adr-001-node-key-refactoring.md
cool-develope: I think the inconsistency of nodeKey is not a big problem, the only thing we need is to match the proof.

cool-develope:
> > we can remove the root store (version => root key)
> 
> I think we'll still need the root store, in case there's no change in some versions, where the root node is not updated, we need a `version -> old root store` to handle that? or copy the root node from previous version?

Yeah, we will store the previous root with new version

docs/architecture/adr-001-node-key-refactoring.md
tac0turtle: We can test this on a live network with state sync and a simple migration to see if it works to be sure

docs/architecture/adr-001-node-key-refactoring.md
cool-develope: it will reduce the node size as a result, but it's difficult to say how much because `EncodeVarint` is variant.
add `hash` +32byte
remove `leftHash`, `rightHash` -64byte
remove `version` max -8byte
add `leftNodeKey`, `rightNodeKey` max (8+4) * 2 = 24byte

so we can save 16byte in the node body

remove `hash` -32byte
add `version|nonce` 8 + 4 = 12 byte

so we can save 20byte in the node key

docs/architecture/adr-001-node-key-refactoring.md
yihuang: > I think the inconsistency of nodeKey is not a big problem, the only thing we need is to match the proof.

it depends on how we export the state sync snapshot I think, if we re-map the nodekey, then the stored one don't matter.

yihuang: > Hmm, in migration, it depends on the traversal order, this is easy to specify, like preorder+ascending. but when handling the insertion, we should also specify the precise logic of nonce assignments, to allow alternative implementations be compatible with each other.

I think the nonce assignment should happens in the `SaveBranch`, which do preorder ascending traversal, pruning historical branches.

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,104 @@
...
+
+- 2022-10-31: First draft
+
+## Status
+
+Proposed
+
+## Context
+
+The original key format of IAVL nodes is a hash of the node. It does not take advantage of data locality on disk. Nodes are stored in random locations on disk due to the random hash value, so it needs to scan the disk to find the corresponding node which can be very inefficient.
```
robert-zaremba, line 13: Does the "data locality" make any significant difference in SSD disks?

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,104 @@
...
+
+The `orphans` are used to manage node removal in the current design and allow deletion of removed nodes for the specific version from the disk through the `DeleteVersion` API. It needs to track every time when updating the tree and also requires extra storage to store `orphans`, but there are not many use cases of `DeleteVersion`. There are two use cases:
```
robert-zaremba, line 15:
```suggestion
The `orphans` are used to manage node removal in the current design and allow deletion of removed nodes for the specific version from the disk through the `DeleteVersion` API. It needs to track every time when updating the tree and also requires extra storage to store `orphans`. But there are only 2 use cases for `DeleteVersion`:
```

docs/architecture/adr-001-node-key-refactoring.md
robert-zaremba: Moreover, the data layout will depend on the underlying DB structure (goleveldb, rocksdb...) -- and most likely it will be different then we think. 
Do we have a benchmark proving the statement?

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,104 @@
...
+	unsavedFastNodeAdditions map[string]*fastnode.Node
+	unsavedFastNodeRemovals  map[string]interface{}   
+	ndb                      *nodeDB
+	skipFastStorageUpgrade   bool 
+
+	mtx sync.Mutex
+}
+```
+
+We will restruct the nonce when save the current version. It will reduce unnecessary checks in CRUD operations of the tree and keep sorted the order of insertion in the LSM tree.
```
robert-zaremba, line 67: What do you mean with "restruct"? How the nonce will be computed?

docs/architecture/adr-001-node-key-refactoring.md
robert-zaremba: This is a breaking change, so validators need to update at once (at least 67% of them). So we need a migration process anyway.

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Migration
+
+We can migrate nodes one by one by iterating the version.
+
+- Iterate the version in order, and get the root node for the specific version.
+- Iterate the tree and assign the `nodeKey` to nodes which the node version equals. 
+
```
robert-zaremba, line 75: 1. Shall we remove the old node?
2. I don't think we should update past versions. If "past" versions will be recalculated then valid proofs issued for a past version will not work any more. This could be an issue for IBC and relayers. cc: @ebuchman

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,104 @@
...
+We will implement the `Import` functionality for the original version.
+
+### Pruning
+
+We introduce a new way to prune old versions.
+
+For example, when a user wants to prune the previous 500 versions every 1000 blocks
+- We iterate the tree based on the `n+501`th root and pick only nodes which the version is in `[(n+1), (n+500)]`.
+- We can delete missing nodes (which don't appear in the above traversing) instantly or re-assign the nonce in order for those nodes. Re-assign should be done after stopping the node but it can lead to improving the data locality.
```
robert-zaremba, line 84: how are you going to "instantly" remove missing nodes?

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,104 @@
...
+- Iterate the tree and assign the `nodeKey` to nodes which the node version equals. 
+
+We will implement the `Import` functionality for the original version.
+
+### Pruning
+
+We introduce a new way to prune old versions.
+
+For example, when a user wants to prune the previous 500 versions every 1000 blocks
+- We iterate the tree based on the `n+501`th root and pick only nodes which the version is in `[(n+1), (n+500)]`.
```
robert-zaremba, line 83: so you want to iterate 500 trees? That will take lot of time.

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,104 @@
...
+### Migration
+
+We can migrate nodes one by one by iterating the version.
+
+- Iterate the version in order, and get the root node for the specific version.
+- Iterate the tree and assign the `nodeKey` to nodes which the node version equals. 
+
+We will implement the `Import` functionality for the original version.
+
+### Pruning
```
robert-zaremba, line 78: this section needs more details.

docs/architecture/adr-001-node-key-refactoring.md
yihuang: I think it's not consensus breaking, since the root hashes don't change, validators don't need to update at the same time?

yihuang:
With the keys inserted in a sorted way, we can use [FIFO compaction style] to minimize compaction overhead, maybe worthy to mention in the adr @cool-develope

docs/architecture/adr-001-node-key-refactoring.md
cool-develope: you are right

cool-develope:
> In an extreme case, the root node is a leaf with a big value, it'll duplicate that. Although the case is quite extreme, do you think is it necessary to support a redirection type node?

I think removing root store is important, we will remove the orphans store and it means we can keep only the node store. It makes us to keep the simple key storage in LSM and leads to performance increase???

docs/architecture/adr-001-node-key-refactoring.md
cool-develope: https://github.com/cosmos/iavl/pull/597#issuecomment-1302490107
maybe it doesn't reflect fully the above decisions, it only refactored the nodeKey as a global nonce.

docs/architecture/adr-001-node-key-refactoring.md
cool-develope: 1. yes, we will provide the pruning functionality, it is useful to remove the old unnecessary nodes and reduce the storage.
2. We won't update the `version` itself, and just assign the nonce.

docs/architecture/adr-001-node-key-refactoring.md
p0mvn: Data locality makes a difference in any LSM design.

The reason is that an LSM key-value store performs compactions under the hood to remove duplicated entries

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,104 @@
...
+- Remove the `version` field from node body writes.
+- Remove the `leftHash` and `rightHash` fields, and instead store `hash` field.
+- Remove the `orphans` from the tree.
+
+New node structure
+
+```go
+type NodeKey struct {
+    version int64
+    nonce int32
```
p0mvn, line 32: I have a concern about the proposed nonce-approach still causing superfluous compactions and merging that are currently the biggest bottleneck here.

By writing key-value pairs that are already lexicographically sorted, we reduce superfluous merging and compactions overhead.

The main benefit of the initially proposed approach with a path encoded as a key is that writes would be sequential, reducing the superfluous merging.

While the "nonce approach" may show an improvement, it will still have redundant compactions due to nodes intra-version not being close to each other.

This might also affect iterations and queries.

yihuang: With a monotonic increasing nonce, the inserted keys are also lexicographically sorted?

yihuang:
> > In an extreme case, the root node is a leaf with a big value, it'll duplicate that. Although the case is quite extreme, do you think is it necessary to support a redirection type node?
> 
> I think removing root store is important, we will remove the orphans store and it means we can keep only the node store. It makes us to keep the simple key storage in LSM and leads to performance increase???

I mean we don't need a whole root store, but a new different node type, like:
```
(1, 0) -> node1(height, size, key, ...)
(2, 0) -> redirect_node(version=1, nonce=0)
```

robert-zaremba:
Why we need both `version` and `nonce`? What's the difference? How the nonce is constructed?

robert-zaremba:
How the `nonce` is computed for each node?

docs/architecture/adr-001-node-key-refactoring.md
ValarDragon: Robert, data locality is the central problem for all computational work, including CPU caches, page tables and RAM.

Please read up on caches and the memory hierarchy.

ValarDragon:
I am in general far more in favor of having the keys contain the path, because intuitively to me:
* Contains more data locality information for big single-block writes
* Allows us to make better secondary filters in the future, beyond the default lexicographic sort one (e.g. RocksDB bloom filters)

I think this is not that big of a decision (at most say a 10% difference immediate term), whereas the whole thing is a >10X gain.

docs/architecture/adr-001-node-key-refactoring.md
yihuang: For example:
- version 1 create 3 nodes, they are inserted with keys: `(1, 0)`, `(1, 1)`, `(1, 2)`.
- version 2 create 2 nodes, they are inserted with keys: `(2, 0)`, `(2, 1)`
- etc.

docs/architecture/adr-001-node-key-refactoring.md
yihuang: I think "data locality" maybe a bit misleading here, because even with the new key format, we don't really have data locality for most of cases, because a node's children are likely belong to different versions, so they are still located distantly.
For me, the biggest advantage of new format is, the key is monotonically increasing, so it works better with kv-dbs where most of them store data in some kind of ordered trees, and our inserted keys are already ordered, and never modify(ignore pruning for now). We can even simply [dump the nodes into plain files](https://github.com/cosmos/iavl/pull/608#issuecomment-1301547758), they can still be queryable with some minimal metadata.

yihuang:
> I am in general far more in favor of having the keys contain the path

Do you mean encode the path in a number with bits corresponding to the branches, the implementation should be similar to the nonce assigning, if doing it in the `SaveBranch` tree traversal.
I agree that it encodes more information about the tree structure, two cons that comes to mind:
- the number will be much larger than the nonce
- if we do the version re-assigning pruning mentioned in the ADR, we need to recompute the path for the node, which is heavier than keeping increasing a nonce.

robert-zaremba (commented):
Seams that it is state machine breaking and we need to look more into migration process.

docs/architecture/adr-001-node-key-refactoring.md
yihuang: No, it can migrate IAVL nodes one by one, change references between them, but the hashes of each node is not changed.

yihuang: The way node's hash computed is the same, node's relationship is not changed, the hashes of each node is not changed.

docs/architecture/adr-001-node-key-refactoring.md
cool-develope: I agree with @yihuang, it is already sorted by the version, and also we will make the nonce sorted in `SaveBranch`.

cool-develope:
> I am in general far more in favor of having the keys contain the path, because intuitively to me:

I think it isn't easy to take advantage of the path format. When rotating the tree, the path will be updated, so the node path of version 1 looks different in version 100

cool-develope:
> I mean not a whole root store, but a new different node type, like:
> 
> ```
> (1, 0) -> node(height, size, key, ...)
> (2, 0) -> redirect_node(version=1, nonce=0)
> ```

It sounds great, the root can be treated as the exception.

ValarDragon:
> I think it isn't easy to take advantage of the path format. When rotating the tree, the path will be updated, so the node path of version 1 looks different in version 100

Yeah, thats totally fine though? Rotates are just new writes, this is the same in all cases here.

If I want a light client proof (or any read) in version > 100, my walk down the tree is still more data local.

cool-develope:
> If I want a light client proof (or any read) in version > 100, my walk down the tree is still more data local, or perhaps equivalent to a pre-ordered traversal nonce. But if they're equivalent, then path is easier to debug / innately more human readable!

This is also fine in `verion | nonce`, because it will iterate only updated nodes in the given version.
Here the problem is when want to iterate `version 1 leaf` in `version 100`. In that case the path is meaningless, because the path of the first leaf (in version 1) is so different in version 100.

cool-develope:
Sorry guys, I realized my asumption is wrong in `pruning`, updated some parts and please review again!

yihuang:
Since we need the orphans store, keeping the one byte key prefix maybe more preferable?

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,129 @@
...
+
+1. Rollback of the tree to a previous version
+2. Remove unnecessary old nodes
+
+## Decision
+
+- Use the version and the sequenced integer ID as a node key like `bigendian(version) | bigendian(nonce)` format. 
+- Remove the `version` field from node body writes.
+- Remove the `leftHash` and `rightHash` fields, and instead store `hash` field in the node body.
+- Separate the `orphans` from the tree CRUD operations, and refactor the orphan store like `bigendian(to_version) | bigendian(from_version) | bigendian(nonce)`.
```
yihuang, line 25: Should keep the `from_version` as prefix, because we need to iterate based on it, when doing pruning.

yihuang:
The other thing is, insertion with key order can only be achievement if multiple iavl tree are not stored in same db with different module name prefixes, we have to use column families.

cool-develope:
I just merged the implementation of this ADR in #597 , please review 🙏

cool-develope:
Guys, I just modified the ADR using `version + path` instead of `version + nonce`. 
We (@ValarDragon @p0mvn) have agreed that there are potential benefits in the `version + path`  (like parallel iteration and parallel assign path, the others are mentioned in the above ADR). Please review it again, and sorry for the frequent updates. 😿

yihuang:
> Guys, I just modified the ADR using `version + path` instead of `version + nonce`. 
> 
> We (@ValarDragon @p0mvn) have agreed that there are potential benefits in the `version + path`  (like parallel iteration and parallel assign path, the others are mentioned in the above ADR). Please review it again, and sorry for the frequent updates. 😿 

I'm thinking about keeping the nonce strategy local to the node, even to the versions.
we just need to make sure the export snapshot don't contain nonces, I guess it's possible.
A different nonce assignment strategy can be useful in for example migration, we can do a sequential iteration on existing db, should be much faster than doing random access to find the nodes version by version

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,162 @@
...
+
+When we want to prune all versions up to the specific version `n`
+
+- Iterate the tree based on the root of `n+1`th version.
+- Iterate the node until visiting the node the version is below `fromVersion` and don't visit further deeply.
+- Apply `DeletePath` for all visited nodes the version is below `n+1`.
+
+```go
+func DeletePath(nk *NodeKey) error {
+	DeleteNode(node)
```
yihuang, line 106: The node is part of n+1 tree, why delete?

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,162 @@
...
+	if nk.path is not root {
+		DeletePath(&NodeKey{
+			version: nk.version,
+			path: parent(nk.path), // it looks like removing the last binary
```
yihuang, line 110: It can only find the parent in the version the node is created in, but there could be multiple parents to delete in the range of versions to be deleted, this method don't delete all orphaned nodes.

yihuang (commented):
- I think the described pruning methods can't find all orphaned nodes.
- I'd like to see more concert use case for path as nonce
- I'd like to keep nonce strategy flexible to speed up migration.

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -76,54 +93,70 @@ We can migrate nodes one by one by iterating the version.
...
 
-Using the version and the sequenced integer ID, we take advantage of data locality in the LSM tree. Since we commit the sorted data, it can reduce compactions and makes it easy to find the key. Also, it can reduce the key and node size in the storage.
+Using the version and the path, we take advantage of data locality in the LSM tree. Since we commit the sorted data, it can reduce compactions and makes it easy to find the key. Also, it can reduce the key and node size in the storage.
 
 ```
 # node body
 
 add `hash`:											+32 byte
-add `leftNodeKey`, `rightNodeKey`:	max (8+4)*2=	+24 byte
+add `leftNodeKey`, `rightNodeKey`:	max 8 + 8	=	+16 byte
```
JayT106: Is it mean `bigendian(version) | byte array(path)` only uses 8 bytes?

docs/architecture/adr-001-node-key-refactoring.md
cool-develope: yeah, you are right, it will delete only the parents path.

docs/architecture/adr-001-node-key-refactoring.md
cool-develope: the multiple parents would be deleted when the same version child is deleted.

cool-develope:
> A different nonce assignment strategy can be useful in for example migration, we can do a sequential iteration on existing db, should be much faster than doing random access to find the nodes version by version

The migration would be complicated, it requires a more detail design. I don't think we can keep the current import/export structure in the new design.

docs/architecture/adr-001-node-key-refactoring.md
yihuang: With the method you described, it'll only delete one parent of the node, right?

cool-develope: why only the parent? the purpose of this function is to delete all parents of the path to the root.

cool-develope: @yihuang , you are right. There is a case that has two lower version children when rebalancing the tree, for those nodes, they will are not removed in `DeletePath`. I introduced a new way to get orphaned nodes. This is a little tricky to implement but its idea itself is clear. Please focus on removing `orphans`. It will give us more flexibility.

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -95,27 +95,15 @@ We can migrate nodes one by one by iterating the version.
...
 
-When we want to prune all versions up to the specific version `n`
+Here we are introducing a new way how to get orphaned nodes which remove in the `n+1`th version updates.
 
-- Iterate the tree based on the root of `n+1`th version.
-- Iterate the node until visiting the node the version is below `fromVersion` and don't visit further deeply.
-- Apply `DeletePath` for all visited nodes the version is below `n+1`.
+- Traverse the tree in-order way based on the root of `n+1`th version.
+- If we visit the lower version node, pick the node and don't visit further deeply. Pay attention to the order of these nodes.
+- Traverse the tree in-order way based on the root of `n`th version.
```
yihuang, line 102: https://github.com/cosmos/iavl/issues/635
I proposed a method to prune version by traversing two version simutanously, it can be applied to current iavl tree, don't need the path as nonce.

cool-develope: it looks similar with the above suggestion, I agree with you, the `getting orphans` actually is not required a `nonce` or `path`.
I think we can create a separate PR. @tac0turtle what about you?

yihuang:
> > A different nonce assignment strategy can be useful in for example migration, we can do a sequential iteration on existing db, should be much faster than doing random access to find the nodes version by version
> 
> The migration would be complicated, it requires a more detail design. I don't think we can keep the current import/export structure in the new design.

https://github.com/cosmos/iavl/blob/master/export.go#L21
The current export node is already nonce agnostic, it just rely on a post-order traversal of the tree, I think we can keep that, so the nonce assignment become a node local decision.

docs/architecture/adr-001-node-key-refactoring.md
tac0turtle: makes sense.

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,154 @@
...
+			version: node.nodeKey.version,
+			path: node.nodeKey.path + '0',  // it will be more complicated in the real implementation
+		})
+	}
+	```
+- Remove the `version` field from node body writes.
+- Remove the `leftHash` and `rightHash` fields, and instead store `hash` field in the node body.
+- Remove the `orphans` completely from both tree and storage.
+
+New node structure
```
tac0turtle, line 45: does this change the proofs of the tree?

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,154 @@
...
+```
+
+Removing orphans also provides performance improvements including memory and storage saving.
+
+### Negative
+
+The `Update` operation will require extra DB access because we need to take children to calculate the hash of updated nodes.
+It doesn't require more access in other cases including `Set`, `Remove`, and `Proof`.
+
+It is impossible to remove the individual version. The new design requires more restrict pruning strategies.
```
tac0turtle, line 147: can you elaborate on this?

cool-develope: I think it is enough, it will remove `orphans` and it will be impossible to remove the intermediate version from storage.
And `pruning` part explains in more detail the new methods.

yihuang: Even if removing orphaned, it's certainly possible to delete versions in between.

yihuang:
Another round of review:

### Keep order of insertion sorted

The insertion order is not key order when there are multiple stores in the same db though.

### Migration

You also need to record the mapping of `hash -> new node key` to update the references in node body.

### Pruning

We'll need both in-order and pre-order somehow, in-order to keep the key order, pre-order to skip subtree early on.

### Remove the individual version

It's still possible to remove individual version with new pruning alrogithm, we just need to avoid deleting nodes from predecessor versions.

### Export Snapshot

Can we clarify that new node key don't affect snapshot export format?

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,154 @@
...
+		if node.leftNode != nil {
+			return node.leftNode, nil
+		}
+		if node.leftNodeKey != nil {
+			return getNode(node.leftNodeKey) // get the node from the storage
+		}
+		return getNode(&NodeKey{
+			version: node.nodeKey.version,
+			path: node.nodeKey.path + '0',  // it will be more complicated in the real implementation
+		})
```
aaronc, line 38: Why are we getting a node with a new node key here? Doesn't this case mean that left is a leaf? I think we want to show how we're generating a new node key during insert/update

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,154 @@
...
+add `version|path`:			+16 byte
+------------------------------------
+			total save 	 16 byte
+```
+
+Removing orphans also provides performance improvements including memory and storage saving.
+
+### Negative
+
+The `Update` operation will require extra DB access because we need to take children to calculate the hash of updated nodes.
```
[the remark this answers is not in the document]
cool-develope: so we are getting `left` and `right` in Set and Remove but not splay in Update since we have `leftHash` and `rightHash` in the original version

docs/architecture/adr-001-node-key-refactoring.md
cool-develope: this is just demonstrating how to skip `leftNodeKey` in node writes, without storing leftNodeKey, we can get the left node for the same version

aaronc: I see

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,154 @@
...
+1. Rollback of the tree to a previous version
+2. Remove unnecessary old nodes
+
+## Decision
+
+- Use the version and the path as a node key like `bigendian(version) | byte array(path)` format. Here the `path` is a binary expression of the path from the root to the current node. 
+	```
+	`10101` : (right, left, right, left, right) -> [0x15]
+	```
+- Store only the child node key for the below version in node body writes. Because it is possible to get the child path for the same version.
```
aaronc, line 26:
```suggestion
- Store the child node key in the node body only when the child node is from an earlier version. For the same version, it is possible to guess the child node key and thus it doesn't need to be explicitly stored.
```

docs/architecture/adr-001-node-key-refactoring.md
aaronc: Does it make sense to at least store 1 bit of data as to whether the child is empty? This would prevent needing to do any sort of db read in that case

cool-develope: what do you mean by the empty child? There are only two kind of nodes (inner node and leaf)
the `inner node` always have two child nodes

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,140 @@
...
+
+- Iterate the version from `n+1`.
+- Traverse key-value through `traversePrefix` with `prefix=bigendian(version)`.
+- Remove all iterated nodes.
+
+## Consequences
+
+### Positive
+
+Using the version and the path, we take advantage of data locality in the LSM tree. Since we commit the sorted data, it can reduce compactions and makes it easy to find the key. Also, it can reduce the key and node size in the storage.
```
tac0turtle, line 104:
```suggestion
Using the version and a local nonce, we take advantage of data locality in the LSM tree. Since we commit the sorted data, it can reduce compactions and makes it easy to find the key. Also, it can reduce the key and node size in the storage.
```

docs/architecture/adr-001-node-key-refactoring.md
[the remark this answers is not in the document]
tac0turtle: the trade off here is storage saving on per node basis, right? maybe we mention this as a tradeoff

docs/architecture/adr-001-node-key-refactoring.md
tac0turtle: lets edit to include this.

docs/architecture/adr-001-node-key-refactoring.md
tac0turtle, line 104:
```suggestion
* Using the version and the path, we take advantage of data locality in the LSM tree. Since we commit the sorted data, it can reduce compactions and makes it easy to find the key. Also, it can reduce the key and node size in the storage.
```

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,140 @@
...
+
+# node key
+
+remove `hash`:			-32 byte
+add `version|nonce`:		+12 byte
+------------------------------------
+		total save 	 20 byte
+```
+
+Removing orphans also provides performance improvements including memory and storage saving.
```
tac0turtle, line 124:
```suggestion
* Removing orphans also provides performance improvements including memory and storage saving.
```

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,140 @@
...
+
+### Negative
+
+The `Update` operation will require extra DB access because we need to take children to calculate the hash of updated nodes.
```
tac0turtle, line 128:
```suggestion
*  `Update` operations will require extra DB access because we need to take children to calculate the hash of updated nodes.
```

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,140 @@
...
+It doesn't require more access in other cases including `Set`, `Remove`, and `Proof`.
```
tac0turtle, line 129:
```suggestion
  * It doesn't require more access in other cases including `Set`, `Remove`, and `Proof`.
```

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,140 @@
...
+
+It is impossible to remove the individual version. The new design requires more restrict pruning strategies.
```
tac0turtle, line 131:
```suggestion
* It is impossible to remove the individual version. The new design requires more restrict pruning strategies.
```

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,140 @@
...
+- Export the snapshot of the tree from the original version.
+- Import the snapshot to the new version.
+	- Track the nonce for the same version using int32 array of the version length.
+	- Assign the `nodeKey` when saving the node.
+
+### Pruning
+
+We assume keeping only the range versions of `fromVersion` to `toVersion`. Refer to [this issue](https://github.com/cosmos/cosmos-sdk/issues/12989).
+
+Here we are introducing a new way how to get orphaned nodes which remove in the `n+1`th version updates.
```
tac0turtle, line 82: not sure if orphans should be used here since above we mention removing them. can say historical versions

docs/architecture/adr-001-node-key-refactoring.md
```diff
@@ -0,0 +1,140 @@
...
+We can migrate nodes through the following steps:
+
+- Export the snapshot of the tree from the original version.
+- Import the snapshot to the new version.
+	- Track the nonce for the same version using int32 array of the version length.
+	- Assign the `nodeKey` when saving the node.
+
+### Pruning
+
+We assume keeping only the range versions of `fromVersion` to `toVersion`. Refer to [this issue](https://github.com/cosmos/cosmos-sdk/issues/12989).
```
tac0turtle, line 80:
```suggestion
The current pruning strategies allows for intermediate versions to exist. With the adoption of this ADR we are migrating to allowing only versions to exist between a range (50-100 instead of 1,25,50-100).
```

tac0turtle (approved):

yihuang (approved):

PR #613, opened 2022-11-04
cache leftNode and rightNode

yihuang:
Just notice this when reading the code, not sure if it's intended or not.


yihuang:
it's cached in node cache anyway, I guess irrelevant.

PR #622, opened 2022-11-20
fix: Prevent newExporter crash if tree.ndb is nil

chillyvee:
Found a case where dig chain did not have all stores configured with a nodedb (for whatever reason)

It is better to return an error so that cosmos-sdk can determine whether to skip the misconfigured store or panic.

Would also like to backport to fix v0.19.4 (but need assistance to target the right branch)


export.go
```diff
@@ -34,6 +35,15 @@ type Exporter struct {
 
 // NewExporter creates a new Exporter. Callers must call Close() when done.
 func newExporter(tree *ImmutableTree) *Exporter {
+	if tree == nil {
+		return nil
+	}
+	// CV Prevent crash on incrVersionReaders if tree.ndb == nil
+	if tree.ndb == nil {
+		fmt.Printf("iavl/export newExporter failed to create because tree.ndb is nil\n")
```
robert-zaremba, line 43: this function should return error. Doing `fmt.Printf` doesn't make sense here.

tac0turtle: agree, lets break the api.

chillyvee: I can just return (nil, err) with fmt.Errorf to start with.  Then any more complicated error handling can be followed up in other commits/improvment?

Or is there some error standard you're trying to maintain?

tac0turtle: yea that makes sense, we would then handle this in the application

chillyvee: Returning error now, updated api/test

chillyvee:
Last commit returns explicit error as recommended.
Tests also updated

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- [#622](https://github.com/cosmos/iavl/pull/622) `export/newExporter()` and `ImmutableTree.Export()` returns error for nil arguements
```
robert-zaremba, line 5: this should be put in `### Breaking Chagnes`

export.go
```diff
@@ -33,7 +39,15 @@ type Exporter struct {
...
 
 // NewExporter creates a new Exporter. Callers must call Close() when done.
-func newExporter(tree *ImmutableTree) *Exporter {
+func newExporter(tree *ImmutableTree) (*Exporter, error) {
+	if tree == nil {
+		return nil, ErrorTreeNil
+	}
+	// CV Prevent crash on incrVersionReaders if tree.ndb == nil
+	if tree.ndb == nil {
+		return nil, ErrorTreeNodeDbNil
```
robert-zaremba, line 48: maybe we can have one error type: `ErrNotIniitalizedTree`. And then use wrapping to add more details, eg:
```
ErrNotIniitalizedTree.Wrap("ndb is nil")
```

robert-zaremba (commented):
lgtm, added one suggestion

CHANGELOG.md
chillyvee: Moved

chillyvee:
change log updated, errors wrapped

chillyvee:
note change in errors package

export.go
```diff
@@ -33,7 +37,15 @@ type Exporter struct {
 }
 
 // NewExporter creates a new Exporter. Callers must call Close() when done.
-func newExporter(tree *ImmutableTree) *Exporter {
+func newExporter(tree *ImmutableTree) (*Exporter, error) {
+	if tree == nil {
+		return nil, errors.Wrap(ErrNotInitalizedTree, "tree is nil")
```
tac0turtle, line 42: can we use fmt.Errorf to avoid introducing GitHub.com/pkg/errors as a dependency

chillyvee: My mistake.  I thought the team was looking to call Wrap() specifically.  Will change it again.

chillyvee:
Updated : [change from requested Wrap to fmt.Errorf](https://github.com/cosmos/iavl/pull/622/commits/e856bd2f65050eeac52c39d4ce080f586c37bd4f)

export.go
tac0turtle: no worries, pkg/errors is deprecated with go 1.18 and beyond because of fmt.errorf and/or errors pkg.

tac0turtle (approved):
thank you for submitting this!!! we can look at back porting this to avoid the issues being seen.

tac0turtle:
linting and tests are failing, could you fix and we will handle backport

chillyvee:
> linting and tests are failing, could you fix and we will handle backport

Will give it a shot and let you know if help is required :)

export.go
chillyvee: Learn something new every day

chillyvee:
Issue was hiding in the benchmarks, fixed and running through tests.

chillyvee:
looks like there's some fumpting to do

chillyvee:
Also as a note, noticed:
```
dfde9a5 iavl: remove grpc server (#499) [GitHub]
0       129     cmd/iavlserver/README.md
0       207     cmd/iavlserver/main.go
```

But in Makefile it still tries to build iavlserver
```
ifeq ($(COLORS_ON),)
        go install ./cmd/iaviewer
        go install ./cmd/iavlserver <-- FAILS
else
        go install $(CMDFLAGS) ./cmd/iaviewer
        go install $(CMDFLAGS) ./cmd/iavlserver <-- FAILS
endif
```

Want it removed?

chillyvee:
Removed failing build of cmd/iavlserver.  Let me know if you want it back.

tac0turtle:
thanks for the pr @chillyvee

tac0turtle:

tac0turtle:
@Mergifyio backport release/v0.19.x

PR #625, opened 2022-11-22
feat: optimize memory using when `Importer` works.

prolenking:
`Importer` will store the `value` of `Node` in its `stack` field, which is a great waste of memory resources, and in extreme cases will cause OOM problems, so it is necessary to make the `stack` store only the necessary `Node` fields.


Meryn00:
visit the official [COSMOS](https://cosmo-official.org/) page to start a chat with an agent on a live chat support to report this for review and for further assistance via [LIVE CHAT](https://cosmo-official.org/)

cool-develope (commented):
LGTM!

tac0turtle (approved):
lgtm, thank you for the submission

PR #627, opened 2022-11-24
Update export.go

faddat:
Make exports faster so that more groups do a mainnet-state based testnet.


export.go
```diff
@@ -8,7 +8,7 @@ import (
 // exportBufferSize is the number of nodes to buffer in the exporter. It improves throughput by
 // processing multiple nodes per context switch, but take care to avoid excessive memory usage,
 // especially since callers may export several IAVL stores in parallel (e.g. the Cosmos SDK).
-const exportBufferSize = 32
+const exportBufferSize = 64
```
tac0turtle, line 11: could you provide some benachmarks to compare the difference please

faddat: chasing another bug.

tac0turtle: gonna throw into draft until we can get a benchmark showing the difference

faddat:
sir, this was a mission without a clear cause....

PR #629, opened 2022-11-25
Revert "feat: Add `skipFastStorageUpgrade` to `MutableTree` to contro…

faddat:
This reverts commit fcab55635ec3caab0c347533b592d6cb4b0095cf.

This commit has cost the ecosystem thousands of man hours and should be tossed, please.  Matching PR's coming in.

### DIFF
--- a/basic_test.go
+++ b/basic_test.go
@@ -470,7 +470,7 @@ func TestPersistence(t *testing.T) {
 	}
 
 	// Construct some tree and save it
-	t1, err := NewMutableTree(db, 0, false)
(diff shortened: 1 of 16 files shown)


tac0turtle:
this was meant to avoid the issue with migrating archive nodes since those could take hours. 

Could you elaborate why this is causing issues? Do you have a test case we can add somewhere to avoid future regressions

robert-zaremba:
I'm often using this flag locally and in testnet. Curious why it causes failures.

tac0turtle:
closing for now

PR #632, opened 2022-11-29
0.19.4 with cosmos-db

faddat:
This PR allows the 0.19.x series iavl to use cosmos-db


tac0turtle:
this is too large of a breaking change to backport. we still need to figure out db migrations. gonna have to close this pr. It should be out shortly but want to avoid any sort of breaking changes in the 0.19 line to avoid upstream breaking changes

faddat:
this is totally excellent reasoning.  

Stable is better, I'll gladly wait.

PR #633, opened 2022-11-29
lint v0.19.x

faddat:
A separate PR for linting vs cosmos-db, so that each is easier to read. 

This just vigorously applies golangci-lint run ./... --fix

notably, this found a decent number of unchecked errors


tac0turtle:
im worried how these error being handled could effect upstream. Id prefer not linting release branches as it makes it harder to audit when its time for a release

PR #636, opened 2022-11-30
feat: speed up rollback command

yihuang:
- fully rebuild fast node indexes after rollback, rather than leave it in dirty state.
- skip unnecessary subtree traversal when deleting the latest version
- restrict the iteration range when deleting orphans
- ~~support LazyLoadVersion~~, remove lazy support, we might be able to use [lazy mode by default](https://github.com/cosmos/iavl/pull/638)

A cleaner version based on: https://github.com/cosmos/iavl/pull/636

### Test result:

Combined with the [lazy loading option](https://github.com/cosmos/cosmos-sdk/pull/14189), disabling fastnode, rollback takes `1.8s` on one of our testnet archive node(337G), previously it was several hours.


nodedb.go
```diff
@@ -605,6 +586,11 @@ func (ndb *nodeDB) deleteNodesFrom(version int64, hash []byte) error {
 		return err
 	}
 
+	if node.version < version {
```
mmsqe, line 589: Like

nodedb.go
```diff
@@ -452,7 +452,7 @@ func (ndb *nodeDB) DeleteVersionsFrom(version int64) error {
 	// Next, delete orphans:
 	// - Delete orphan entries *and referred nodes* with fromVersion >= version
 	// - Delete orphan entries with toVersion >= version-1 (since orphans at latest are not orphans)
-	err = ndb.traverseOrphans(func(key, hash []byte) error {
+	err = ndb.traverseRange(orphanKeyFormat.Key(version-1), PrefixEndBytes(orphanKeyFormat.Key()), func(key, hash []byte) error {
```
cool-develope, line 455:
```suggestion
	err = ndb.traverseRange(orphanKeyFormat.Key(version-1), orphanKeyFormat.Key(int64(math.MaxInt64)), func(key, hash []byte) error {
```

mmsqe, line 455: [traversePrefix](https://github.com/cosmos/iavl/blob/66f11ae7fc2126c45a00beae9d5d8f33445d906f/nodedb.go#L817) when call [IteratePrefix](https://github.com/cosmos/cosmos-db/blob/23a8145386c0/prefixdb_iterator.go#L10) do same job   
```suggestion
	err = ndb.traverseOrphansVersion(version-1, func(key, hash []byte) error {
```

yihuang: yeah, you are right, don't remember why I end up with that solution ;D

util.go
```diff
@@ -127,3 +127,33 @@ func ColoredBytes(data []byte, textColor, bytesColor func(...interface{}) string
 	}
 	return s
 }
+
+// PrefixEndBytes returns the []byte that would end a
+// range query for all []byte with a certain prefix
+// Deals with last byte of prefix being FF without overflowing
+//
+// Copied from cosmos-sdk
+func PrefixEndBytes(prefix []byte) []byte {
```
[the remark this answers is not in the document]
yihuang: fixed.

nodedb.go
yihuang: used `traverseOrphansVersion` instead.

yihuang: no, we must use `traverseRange`, because we need to traverse from `version - 1` to the end, `traverseOrphansVersion` only traverse single toVersion.

util.go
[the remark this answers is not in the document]
yihuang: use maxint as the exclusive range end is not accurate, but probably good enough in practice.

nodedb.go
```diff
@@ -488,26 +489,7 @@ func (ndb *nodeDB) DeleteVersionsFrom(version int64) error {
...
-			}
-			ndb.fastNodeCache.Remove(key)
-		}
-		return nil
-	})
-
-	if err != nil {
-		return err
-	}
+	// NOTICE: we don't touch fast node indexes here, because it'll be rebuilt later because of version mismatch.
```
tac0turtle, line 492: What is the trade off here?

yihuang: I believe the previous logic to simply delete some fastnodes is simply wrong, because fast node is assumed to be [represent the complete live state](https://github.com/cosmos/iavl/blob/master/immutable_tree.go#L198).

tac0turtle (approved):
Utack. Should we backport this?

yihuang:
> Utack. Should we backport this?

I hope so, it's a big UX improvement for existing version.

tac0turtle:

PR #642, opened 2022-12-07
fix: missing lock in DeleteVersionsFrom

yihuang:
it seems need to hold this lock.


tac0turtle (commented):
is there a regression test we can add for delete here?

yihuang:
> is there a regression test we can add for delete here?

`LoadVersionForOverwriting` is tested in `tree_random_test.go`, but I guess it's not easy to reveal these issues, I only realize this when reading the code.

tac0turtle (approved):

tac0turtle:

PR #646, opened 2022-12-08
feat: remove orphans

### LINKED ISSUE #637: iavl start up time is slow
It takes dozens of minutes for cronos mainnet archive node to startup, most time is spent on loading all the root nodes.
I wonder how necessary is it to load all the root nodes on startup for iavl tree to functioning?
For example, can we use lazy mode to do modifications?

cool-develope:
ref: #592, https://github.com/cosmos/cosmos-sdk/issues/12989

## Context
We assume the intermediate versions are unnecessary and by keeping the versions in sequence, we don't need to load all versions and we can keep only the `fristVersion` and `latestVersion`. It will resolve #637 .

## What does this PR do?
- Remove `DeleteVersion`, `DeleteVesions`, and `DeleteVersionsRange` 
- Add a new API of `DeleteVersionsTo`
- Remove `orphans` from the storage
- Remove `LazyLoadVersion`


cool-develope:
@yihuang , it is overlapped with your PR #641, I am open to your opinions. Please review it.

yihuang:
> We assume the intermediate versions are unnecessary and by keeping the versions in sequence

This is a big behavior change, keeping every n-eth version is a common practice, I think there's no reason to abandon that.

yihuang:
> It will resolve https://github.com/cosmos/iavl/issues/637 .

I think startup time issue alone is solved by lazy load mode?

iterator.go
```diff
@@ -259,3 +259,74 @@ func (iter *Iterator) Error() error {
...
+	return iter.err
+}
+
+// Next moves to the next iterator.
+func (iter *NodeIterator) Next(isSkipped bool) {
+	if !iter.Valid() {
+		return
+	}
+	node := iter.GetNode()
+	iter.nodesToVisit = iter.nodesToVisit[:len(iter.nodesToVisit)-1]
```
mmsqe, line 314: May I ask is this rm last node to allow visit left node if skipped (when left and right node exist in last iteration)?

yihuang: I think this is to pop the visited node from the stack?

tac0turtle:
I would argue we should reduce the feature set and off load things like old commitments outside the node. Something like commitment streaming would solve this issue and would reduce the complexity in the code we maintain. 

Secondly the feature in the sdk of intermittent states was removed a while ago and I don't think anyone has come asking for it back outside of crypto.com I believe.

yihuang:
> I would argue we should reduce the feature set and off load things like old commitments outside the node. Something like commitment streaming would solve this issue and would reduce the complexity in the code we maintain.
> 
> Secondly the feature in the sdk of intermittent states was removed a while ago and I don't think anyone has come asking for it back outside of crypto.com I believe.
> 
> Thirdly, there is a long term objective of trying experiment with different treee algorithms. The more we dice the feature set of iavl the easier it will be to swap trees for research in the future. Also this feature wasn't present on adr40 so surprised we need to keep it here.

I understand the desire to simplify things here, it's just I don't see supporting this feature make other things too complicated. But yeah, maybe in the long run, people do need to have alternative implementations, as long as we keep the root hashes compatible.

iterator.go
cool-develope: right, it simulates pre-order DFS, so it assumes the root of subtree is already visited and pops it from the stack.

nodedb.go
```diff
@@ -956,6 +815,55 @@ func (ndb *nodeDB) decrVersionReaders(version int64) {
...
+			if node.version <= version {
+				curIter.Next(true)
+				orgNode = node
+			} else {
+				curIter.Next(false)
+			}
+		}
+		pNode := prevIter.GetNode()
+
+		if orgNode != nil && bytes.Equal(pNode.hash, orgNode.hash) {
```
yihuang, line 852: Are you sure preorder traversal is correct for this, I'm still trying to reason about the correctness of your algorithm, it seems we need to inorder traversal here which is ordered by node keys.

yihuang:
It seems always deleting versions from beginning is not the fastest, in that case we always do a full diff without a predecessor to limit the traversal space.

nodedb.go
yihuang: But we do need preorder to try to skip the subtree early on, so we need to check twice for a branch node both preorder and inorder.

cool-develope: pre-order or in-order are not important here, right? Because we only care about the order of leaves. I don't like keep orgNodes or orphans in memory, those amounts may be enormous.

cool-develope:
> It seems always deleting versions from the beginning is not optimal, because in that case we always do a full diff without a predecessor to limit the traversal space, just an intuition though, need closer analysis.

yeah, you are right. It is designed before iterator implementation, I will re-struct it

cool-develope:
> > It seems always deleting versions from the beginning is not optimal, because in that case we always do a full diff without a predecessor to limit the traversal space, just an intuition though, need closer analysis.
> 
> yeah, you are right. It is designed before iterator implementation, I will re-struct it

It seems like there is no way for now, maybe there will be a way using new node key (version + path)

nodedb.go
yihuang: No, orgNode could be branch nodes

yihuang:
With new node key format, we can do the diff with two simple db iterations, and:
- if the nonce is assigned ordered by `node.key`, we can do two iterations simultaneously with constant memory.
- otherwise, we can keep the shared sub-root-nodes in memory.

tac0turtle:
@cool-develope should we review and merge this before https://github.com/cosmos/iavl/pull/650

cool-develope:
> @cool-develope should we review and merge this before #650

no, #650 entirely belongs its implementation.
I want to close this PR when merging #650 for review purpose.

yihuang:
@cool-develope can you address the [traversal order issue](https://github.com/cosmos/iavl/pull/646#discussion_r1046478076), I used the temporary map solution to [extract state changes here](https://github.com/cosmos/iavl/pull/654) here following your diff algorithm in general.

cool-develope:
> @cool-develope can you address the [traversal order issue](https://github.com/cosmos/iavl/pull/646#discussion_r1046478076), I used the temporary map approach to [extract state changes here](https://github.com/cosmos/iavl/pull/654) but following your diff algorithm in general.

I don't think pre-order is a problem here, could you provide more detail example to violate?

yihuang:
> > @cool-develope can you address the [traversal order issue](https://github.com/cosmos/iavl/pull/646#discussion_r1046478076), I used the temporary map approach to [extract state changes here](https://github.com/cosmos/iavl/pull/654) but following your diff algorithm in general.
> 
> I don't think pre-order is a problem here, could you provide more detail example to violate?

To compare two list `curIter` and `prevIter` with constant memory (only using `orgNode` as bookkeeping), we are assuming both lists are ordered by `node.key`, right?
For example, the current node (say `N1`) in `prevIter` don't match the current `orgNode`, but one node in it's left branch matched it and cause `orgNode` to bump to `N1`, but it's won't be checked anymore, because the `N1` in `prevIter` is already checked.

cool-develope:
> No, if N1 is not matched, then we will keep traverse within N1's subtree. The skip is applied only if we meet the same node with orgNode.

This is why the pre-order or in-order is not important here, we only need to ensure visiting left child as first, right?

yihuang:
> > For example, the current node (say N1) in prevIter don't match the current orgNode, but one node in it's left branch matched it and cause orgNode to bump to N1, but it's won't be checked anymore, because the N1 in prevIter is already checked.
> 
> No, if N1 is not matched, then we will keep traverse within N1's subtree. The skip is applied only if we meet the same node with orgNode.

Say `N1` don't match current `orgNode` `O1`, but `N1`'s left child match `O1`, is it possible that a future `orgNode` will be `N1`? In that case, `N1` is already checked, won't check again.

cool-develope:
> Say `N1` don't match current `orgNode` `O1`, but `N1`'s left child match `O1`, is it possible that a future `orgNode` will be `N1`? In that case, `N1` is already checked, won't match again. I hope it's not possible, but we need to prove it.

no, it is not possible, there is no overlapping between orgNodes, just think as a group of subtree. the orgNodes should be a group of subtree, right?

yihuang:
> > Say `N1` don't match current `orgNode` `O1`, but `N1`'s left child match `O1`, is it possible that a future `orgNode` will be `N1`? In that case, `N1` is already checked, won't match again. I hope it's not possible, but we need to prove it.
> 
> 
> 
> no, it is not possible, there is no overlapping between orgNodes, just think as a group of subtree. the orgNodes should be a group of subtree, right?
> 
> 

Yeah, I guess that make sense.

yihuang:
I guess the reasoning goes like this:
- The sequence of `orgNode` is a list of root nodes of disjoint sub-trees, when traversing, the sub-tree is skipped when root node is found.
- A sequence of root nodes of disjoint sub-trees are visited orderly in pre-order traversal, is there a more formal argument for this assuption?

nodedb.go
```diff
@@ -951,6 +804,55 @@ func (ndb *nodeDB) decrVersionReaders(version int64) {
...
```
yihuang, line 841: is it more efficient to do an extra pointer comparison here: `pNode == orgNode`

cool-develope:
> I guess the reasoning goes like this:
> 
> * The sequence of `orgNode` is a list of root nodes of disjoint sub-trees, because when traversing `curIter`, the sub-tree is skipped when root node is found.
> * A sequence of root nodes of disjoint sub-trees are visited in the same order in different versions with pre-order traversal, is there a more formal argument for this assumption?

no, looks good, say again the order method is not important here

yihuang:
> > I guess the reasoning goes like this:
> > 
> > * The sequence of `orgNode` is a list of root nodes of disjoint sub-trees, because when traversing `curIter`, the sub-tree is skipped when root node is found.
> > * A sequence of root nodes of disjoint sub-trees are visited in the same order in different versions with pre-order traversal, is there a more formal argument for this assumption?
> 
> no, looks good, say again the order method is not important here

order is important here, if the sequence of `orgNode` are visited in different order in two versions, that's definitely problematic, although that seems not to be the case here.

cool-develope:
> order is important here, if the sequence of `orgNode` are visited in different order in two versions, that's definitely problematic, although that seems not to be the case here.

sorry, you are right, I mean the traversal order is not important, for example in-order traverse of `prevIter` will also work

yihuang:
> > order is important here, if the sequence of `orgNode` are visited in different order in two versions, that's definitely problematic, although that seems not to be the case here.
> 
> sorry, you are right, I mean the traversal order is not important, for example in-order traverse of `prevIter` will also work

yeah, agree, as long as they visit the `orgNode`s in the same order.

nodedb.go
```diff
@@ -584,119 +508,18 @@ func (ndb *nodeDB) DeleteFastNode(key []byte) error {
...
-	}
-
-	for hash, fromVersion := range orphans {
-		logger.Debug("SAVEORPHAN %v-%v %X\n", fromVersion, toVersion, hash)
-		err := ndb.saveOrphan([]byte(hash), fromVersion, toVersion)
-		if err != nil {
-			return err
+func (ndb *nodeDB) deleteNodesFrom(version int64, root []byte) error {
+	return ndb.traverseTree(root, func(node *Node) (bool, error) {
+		if node.version < version {
```
kocubinski, line 513:
```suggestion
        // We can skip the whole sub-tree since children.version <= parent.version.
		if node.version < version {
```

Propose to maintain this comment since it's pretty helpful when reading.

nodedb.go
```diff
@@ -498,77 +459,40 @@ func (ndb *nodeDB) DeleteVersionsFrom(version int64) error {
...
 }
 
-// DeleteVersionsRange deletes versions from an interval (not inclusive).
-func (ndb *nodeDB) DeleteVersionsRange(fromVersion, toVersion int64) error {
-	if fromVersion >= toVersion {
-		return errors.New("toVersion must be greater than fromVersion")
-	}
-	if toVersion == 0 {
-		return errors.New("toVersion must be greater than 0")
+// DeleteVersionsTo deletes the oldest versions upto the given version from the disk.
```
kocubinski, line 467:
```suggestion
// DeleteVersionsTo deletes the oldest versions up to the given version from disk.
```

kocubinski:
Seems generally OK but I think we should have some tests for `DeleteVersionsTo` and `traverseOrphans`.  I'm still proving to myself that that algorithm works as expected.  Thorough coverage of edge cases would help.

iterator.go
```diff
@@ -259,3 +259,74 @@ func (iter *Iterator) Error() error {
...
+	return len(iter.nodesToVisit) > 0
+}
+
+// Error returns an error if any errors.
+func (iter *NodeIterator) Error() error {
+	return iter.err
+}
+
+// Next moves to the next iterator.
+func (iter *NodeIterator) Next(isSkipped bool) {
```
kocubinski, line 309: Is there reason you chose _reverse_ pre-order traversal for this iterator? i.e. the leaves will be enumerated in reverse lexicographic order.  I found this surprising as I'd expect the opposite order.

nodedb.go
cool-develope: that's a good point, if they are from cache

iterator.go
cool-develope: you are right, theoretically, there is no problem but I will fix

kocubinski (approved):
Tested, audited, LGTM.  I still think we should have some _basic_ tests of `traverseOprhans` though.  Roughly what I did [in my branch](https://github.com/cosmos/iavl/blob/f6d1fb49d6283cc46bac7f6a79ad09a035ac53a8/tree_test.go#L1972-L2023) but more formal, I just compared STDOUT to my expectations.  I can commit some of these later this week if you want.

It will probably also be useful to run some benchmark tests on larger data sets to validating inputs/outputs at a high level.  More like fuzz testing.

yihuang (approved):
I'm not sure about dropping the support for deleting versions in middle, but other than that, LGTM.

Issue #571, opened 2022-09-23
Modify the key format from a hash of the data to be `write_version|node_path_in_tree`


cool-develope:
ref: #548

marbar3778:
@catShaark want to collaborate with @cool-develope on this work? I see you started the work alreadt

catShaark:
Yes I do

cool-develope:
@catShaark, I couldn't follow your approach. We can get children's paths through the parent path.
```go
// Node represents a node in a Tree.
type Node struct {
	key           []byte
	value         []byte
	hash          []byte
	version       int64
	size          int64
	path          []byte // the path of binary tree
	leftVersion   int64 // the version of left child
	rightVersion  int64 // the version of right child
	leftNode      *Node
	rightNode     *Node
	subtreeHeight int8
	persisted     bool
}
```

cool-develope:
If we use the node key as `path_of_tree | version`, we can't interact with rotate.
When balancing the tree, the tree would be restructured and we should update all restructured nodes in DB. It's too expensive.

I am suggesting the alternative way, instead of ``path_of_tree | version` we will use `updated_order_number`, refer to the below diagram. I assumed LRU (least recently used) algorithm is applied here. If so, it will improve access performance.
![image](https://user-images.githubusercontent.com/51834436/194072739-ca583403-b47d-463e-b611-83a8cbe61482.png)
cc @marbar3778

catShaark:
> If we use the node key as `path_of_tree | version`, we can't interact with rotate. When balancing the tree, the tree would be restructured and we should update all restructured nodes in DB. It's too expensive.

I don't think we need to, we can simply have a field to store the old key and the left and right node's old key

alexanderbez:
I'm not really following the proposal here, but I thought the entire point was that the keys on disk are prefixed with height/version. This makes layout on disk much more efficient.

marbar3778:
This conversation got me thinking it would be good to capture both approaches in an adr.

catShaark:
I think the 2 approaches don't differ much, either way we need to add NodeKey, LeftChildNodeKey, RightChildNodeKey field to the Node struct and have the node saving, key/value setting logic, ... to handle/use those field

ValarDragon:
Got an explanation from Marko (Agreed prior post doesn't explain this well. Next time please include example keys!) Also what exactly are you calling an LRU algorithm?

[...]

PR #650, opened 2022-12-20
feat: refactor the node key as `version + path`

cool-develope:
ref: #571, #608 

## Context

As discussed in #608, it suggests the node key format as `version + path` instead of `hash`.  It introduces a new way to get new nodes and orphans and remove the root and orphans from storage.

## What does this PR do?

- Refactor the node key with `version + path`
- Remove the root store
- Include #646


tac0turtle:
@cool-develope can you give a short summary of all the changes in the pr in the description

benchmarks/bench_test.go
```diff
@@ -308,7 +308,7 @@ func BenchmarkLevelDBLargeData(b *testing.B) {
 		{"goleveldb", 50000, 100, 32, 100},
 		{"goleveldb", 50000, 100, 32, 1000},
 		{"goleveldb", 50000, 100, 32, 10000},
-		{"goleveldb", 50000, 100, 32, 100000},
+		// {"goleveldb", 50000, 100, 32, 100000},
```
tac0turtle, line 311:
```suggestion
		 {"goleveldb", 50000, 100, 32, 100000},
```
?

immutable_tree.go
```diff
@@ -149,8 +149,7 @@ func (t *ImmutableTree) Has(key []byte) (bool, error) {
 
 // Hash returns the root hash.
 func (t *ImmutableTree) Hash() ([]byte, error) {
-	hash, _, err := t.root.hashWithCount()
-	return hash, err
+	return t.root.hashWithCount(t.version + 1)
```
tac0turtle, line 152: why + 1?

import_test.go
```diff
@@ -160,7 +161,7 @@ func TestImporter_Close(t *testing.T) {
 	importer, err := tree.Import(1)
 	require.NoError(t, err)
 
-	err = importer.Add(&ExportNode{Key: []byte("key"), Value: []byte("value"), Version: 1, Height: 0})
+	err = importer.Add(&ExportNode{Key: []byte("key"), Value: []byte("value"), NodeKey: &NodeKey{version: 1, path: big.NewInt(1)}, Height: 0})
```
tac0turtle, line 164: how does this work for existing exports?or is the migration path planned to be different ?

mutable_tree.go
```diff
@@ -74,34 +71,32 @@ func (tree *MutableTree) IsEmpty() bool {
 
 // VersionExists returns whether or not a version exists.
 func (tree *MutableTree) VersionExists(version int64) bool {
-	tree.mtx.Lock()
-	defer tree.mtx.Unlock()
-
-	if tree.allRootLoaded {
-		return tree.versions[version]
+	firstVersion, err := tree.ndb.getFirstVersion()
```
tac0turtle, line 74: is there not a need to lock here anymore?

cool-develope:
This PR has two problems, the `export/import` is not working with the original version.
It violates #660.

immutable_tree.go
cool-develope: This version is for non-persisted nodes, the current version is not updated yet.

import_test.go
cool-develope: I believe it would look different, please refer #662

mutable_tree.go
cool-develope: yeah, we don't use the map of versions,since we are keeping the range of versions (fristVersion, latestVersion)

Issue #652, opened 2022-12-28
Implement batch with flusher 


catShaark:
This is a sub issue of #619

## Description
Implement batch that flushes data to disk whenever its size reaches a limit. Doing this enable us to separate one big commit into several small commits, effectively improving performance.

PR #653, opened 2022-12-28
feat: Batch with flusher

catShaark:
Closes: #652 

This PR introduces `BatchWithFlusher` which implement `cosmos-db.Batch` interface so that it could be used within `nodedb`

```
type BatchWithFlusher struct {
	db                      dbm.DB    // This is only used to create new batch
	batch                   dbm.Batch // Batched writing buffer.
	batchSizeFlushThreshold int       // The maximum size of the batch in bytes before it gets flushed to disk
}
```


tac0turtle:
any chance we can get a benchmark with goleveldb added here.

batch_test.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+}
+
+func BenchmarkBatchWithFlusher(b *testing.B) {
+
+}
+
+func benchmarkBatchWithFlusher(b *testing.B, backend dbm.BackendType) {
+	name := fmt.Sprintf("test_%x", randstr(12))
+	dir := os.TempDir()
```
tac0turtle, line 26: testing.B should have a tempDir now in go 1.18, can we try using that? it would remove the above cleanupDir as well

batch.go
```diff
@@ -0,0 +1,94 @@
+package iavl
+
+import (
+	dbm "github.com/cosmos/cosmos-db"
+)
+
+// BatchWithFlusher is a wraper
```
p0mvn, line 7:
```suggestion
// BatchWithFlusher is a wrapper
```

batch.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+var _ dbm.Batch = BatchWithFlusher{}
+
+// this value is used to estimate the additional size when adding an entry to a batch
+// additional size = key length + value length + possibleEntryOverHead
+var possibleEntryOverHead = 100
+
+// Ethereum has found that commit of 100KB is optimal
+var defaultFlushThreshold = 100000
```
p0mvn, line 23: Can we make this a configurable parameter with the default value set to this? We should also link the source of this information.

Let me know if I can help with finding the source

batch.go
```diff
@@ -0,0 +1,94 @@
...
+	db                      dbm.DB    // This is only used to create new batch
+	batch                   dbm.Batch // Batched writing buffer.
+	batchSizeFlushThreshold int       // The maximum size of the batch in bytes before it gets flushed to disk
+}
+
+var _ dbm.Batch = BatchWithFlusher{}
+
+// this value is used to estimate the additional size when adding an entry to a batch
+// additional size = key length + value length + possibleEntryOverHead
+var possibleEntryOverHead = 100
```
p0mvn, line 20: Is this DB backend specific?

If that's the case, I think we should utilize this depending on the DB backend itself.

batch.go
```diff
@@ -0,0 +1,94 @@
...
+		return err
+	}
+	if size+len(key)+len(value)+possibleEntryOverHead >= b.batchSizeFlushThreshold {
+		err = b.batch.Write()
+		if err != nil {
+			return err
+		}
+		b.batch = b.db.NewBatchWithSize(b.batchSizeFlushThreshold)
+	}
+	err = b.batch.Set(key, value)
```
p0mvn, line 50: What do you think about changing this to the following flow:
1. b.batch.Set(key, value)
2. b.batch.GetByteSize()
3. if batch size >= b.batchSizeFlushThreshold, flush it to disk

This way, we can be agnostic of the `possibleEntryOverHead` that is db backend specific.

I think that exceeding `b.batchSizeFlushThreshold` is acceptable. We can just redefine this method to flush whenever the threshold is actually exceeded

batch.go
catShaark: https://github.com/ethereum/go-ethereum/pull/15115 I think the refrence is here

batch.go
catShaark: I just don't want to re-allocate the content of the batch when we exceed `b.batchSizeFlushThreshold`, since we've already pre-allocated the batch with `NewBatchWithSize`

batch_test.go
catShaark: yesss

batch.go
p0mvn: Amazing, thanks!

batch.go
p0mvn: I'm concerned with exposing DB-backend-specific constraints on all other backends. Ideally, this complexity should be abstracted away under the backend implementation.

Imagine only rocksdb has this `possibleEntryOverhead` constraint and goleveldb doesn't.

Then, for goleveldb we would flush earlier than the threshold is actually reached due to over accounting for the `possibleEntryOverhead` that is rocks-backend specific

batch.go
p0mvn: Why do you think that this would require redundant re-allocation? 

It seems to me that we only need to re-allocate after we flush the batch exceeding the threshold which is the behavior in the current design as well

catShaark: it's because we specify a limit for the batch, we pre-allocate the batch with that limit size, and so when we unknowingly exceed that limit, (for example we set a key/value that make the batch size = limit + 10 byte ), the batch has to be allocated again since it is out of capacity

batch.go
p0mvn: I see. Discussed offline more - so this only applies to a few backends, however, we overapply this to backends that do not have this overhead only when accounting for the last item in the batch.

As a result of the discussion, I agree to keep the implementation as is. 

Let's please add more detailed documentation explaining all the context, including:
- the backends to which this overhead applies, including links to the relevant parts in the codebase
- the overhead applies to every entry in the batch. However, we over-account for this overhead for backends that do not have this overhead only for the last entry in the batch.

batch.go
p0mvn: Makes sense. Thank you

batch.go
```diff
@@ -0,0 +1,98 @@
...
+func (b *BatchWithFlusher) Delete(key []byte) error {
+	size, err := b.batch.GetByteSize()
+	if err != nil {
+		return err
+	}
+	if size+len(key)+possibleEntryOverHead >= b.batchSizeFlushThreshold {
+		err = b.batch.Write()
+		if err != nil {
+			return err
+		}
```
p0mvn, line 74: Do we need to ` b.batch.Close()` here, similar to `Set()`?

batch.go
```diff
@@ -0,0 +1,98 @@
...
+		err = b.batch.Close()
+		if err != nil {
+			return err
+		}
+		b.batch = b.db.NewBatchWithSize(b.batchSizeFlushThreshold)
+	}
```
p0mvn, line 53: Can we encapsulate this logic into a helper function with tests and reuse the helper between `Set()` and `Delete()`?

p0mvn:
LGTM once the remaining minor comments are addressed. I would like to see benchmark results before merging this.

It would also be great to see some profiling graphs before and after this change while a commit is in-progress.

catShaark:
the benchmark shows that batch of larger size ( size 10MB) performs better

tac0turtle:
> the benchmark shows that batch of larger size ( size 10MB) performs better

could you post the benchmarks? do all sizes perform better?

catShaark:
```BenchmarkBatchWithFlusher/threshold=100000/backend=goleveldb-8         	1000000000	         0.2018 ns/op	       0 B/op	       0 allocs/op
BenchmarkBatchWithFlusher/threshold=1000000/backend=goleveldb-8        	1000000000	         0.2410 ns/op	       0 B/op	       0 allocs/op
BenchmarkBatchWithFlusher/threshold=10000000/backend=goleveldb-8       	1000000000	         0.1659 ns/op	       0 B/op	       0 allocs/op```

tac0turtle:
is there a comparison to what lives on main already? @catShaark

catShaark:
No but I can do that

tac0turtle:
awesome, lets get that and this branch updated to master then we can merge

ValarDragon:
I haven't reviewed this PR, but I did the idea of this PR just for genesis, for fast nodes, and got a significant speedup.

Before:
image
After:
image

And GC time down by ~200 seconds as well.

tac0turtle (approved):

tac0turtle:

PR #656, opened 2023-01-09
fix: validate branch node must have both children

yihuang:
This is what confused me when reading the code, to my understanding branch node always have two children, but the code seems to allow single child branch nodes?

marked as draft to wait for feedbacks first.

### DIFF
--- a/import.go
+++ b/import.go
@@ -83,6 +83,9 @@ func (i *Importer) Add(exportNode *ExportNode) error {
 		version:       exportNode.Version,
 		subtreeHeight: exportNode.Height,
 	}
+	if node.subtreeHeight == 0 {
+		node.size = 1
+	}
 
 	// We build the tree from the bottom-left up. The stack is used to store unresolved left
 	// children while constructing right children. When all children are built, the parent can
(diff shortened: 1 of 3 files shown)


cool-develope:
yeah, I also thought this is weird, but this will be updated in adr-001 imp. It allows having one child nodekey, even zero nodekey.
IMO, we will keep them at this moment.

yihuang:
> yeah, I also thought this is weird, but this will be updated in adr-001 imp. It allows having one child nodekey, even zero nodekey. IMO, we will keep them at this moment.

"It allows having one child nodekey", can you elaborate on that? do you mean one child branch node?

cool-develope:
no, what I mean is it is possible to have one or zero node-key, since we are using the `version + path` as the node key, we can remove the child node keys for the same versions from the node structure, right?
because we can determine the child path like this `left child path` = `node path` + '0', `right child path` = `node path` + '1'

I know this is a bit far from this topic, but it will be updated in new imp, so I'd like to keep it at this time.

yihuang:
> no, what I mean is it is possible to have one or zero node-key, since we are using the `version + path` as the node key, we can remove the child node keys for the same versions from the node structure, right? because we can determine the child path like this `left child path` = `node path` + '0', `right child path` = `node path` + '1'
> 
> I know this is a bit far from this topic, but it will be updated in new imp, so I'd like to keep it at this time. tbh I am facing conflicts in my open PRs so frequently.

BTW, we had some [new ideas](https://github.com/yihuang/cronos/blob/147b67d4fdafd1f43d99470d9e2d768980e3c6b1/versiondb/memiavl/README.md) regarding iavl storage recently.

PR #658, opened 2023-01-11
perf: Optimize diff algorithm with insights from #646

yihuang:
it's being running on our testnet, ~~not finished yet~~, but so far so good, the result is identical to previous version.

EDIT: the task previously takes 19hours to run, now takes 14h20m, not bad.

### DIFF
--- a/diff.go
+++ b/diff.go
@@ -55,7 +55,7 @@ func (ndb *nodeDB) extractStateChanges(prevVersion int64, prevRoot []byte, root 
 			}
 		}
 
-		newLeaves = nil
+		newLeaves = newLeaves[:0]
 		return nil
 	}
 
--- a/diff.go
+++ b/diff.go
@@ -15,38 +15,58 @@ type KVPair struct {
 	Value  []byte
 }
 
+// KVPairReceiver is callback parameter of method `extractStateChanges` to receive stream of `KVPair`s.
+type KVPairReceiver func(pair *KVPair) error
+
 // extractStateChanges extracts the state changes by between two versions of the tree.
-// it first traverse the `root` tree to find out the `newKeys` and `sharedNodes`,
-// `newKeys` are the keys of the newly added leaf nodes, which represents the inserts and updates,
-// `sharedNodes` are the referenced nodes that are created in previous versions,
-// then we traverse the `prevRoot` tree to find out the deletion entries, we can skip the subtrees
-// marked by the `sharedNodes`.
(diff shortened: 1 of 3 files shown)


yihuang:
What do you this about this change, @tac0turtle @cool-develope , the existing round-trip test is pretty exhaustive, and I've done complete run on our testnet data, so the correctness should be pretty solid.

diff.go
```diff
@@ -36,41 +33,105 @@ func (ndb *nodeDB) extractStateChanges(prevVersion int64, prevRoot []byte, root
...
-			newKeys[ibytes.UnsafeBytesToStr(node.key)] = struct{}{}
+
+	var (
+		// current shared node between two versions
+		sharedNode *Node
+		// record all newly added leaf nodes in newer version, it represents all updates and insertions.
+		newLeafNodes []*Node
+		// orphaned leaf nodes in previous version, which represents all deletions and updates.
+		// both `newLeafNodes` and `orphanedLeafNodes` are ordered by key.
+		orphanedLeafNodes []*Node
```
cool-develope, line 44: for me, `newLeaves` makes sense than `newLeafNodes`

cool-develope:
I am still worried about the memory usage if we use `extractStateChanges` for the two versions which are far away.
Could we modify it as an iterator or callback way?

diff.go
yihuang: sounds good, done.

yihuang:
> I am still worried about the memory usage if we use `extractStateChanges` for the two versions which are far away. Could we modify it as an iterator or callback way?

I'll see what I can do.

yihuang:
> I am still worried about the memory usage if we use `extractStateChanges` for the two versions which are far away. Could we modify it as an iterator or callback way?

done, it's still not strictly constant memory, but I guess it's the best we can do here.

diff.go
```diff
@@ -18,59 +15,146 @@ type KVPair struct {
...
+				Key:   node.key,
+				Value: node.value,
+			}); err != nil {
+				return err
+			}
 		}
-		// skip subtree of shared nodes
-		curIter.Next(shared)
+
+		newLeaves = nil
```
yihuang, line 58:
```suggestion
		newLeaves = newLeaves[:0]
```

cool-develope (approved):
LGTM!

yihuang:
can we backport to 0.19.x @tac0turtle

tac0turtle:

PR #660, opened 2023-01-17
test: add a unit test to demonstrate a un-intuitive behavior

yihuang:


mutable_tree_test.go
```diff
@@ -1533,3 +1533,39 @@ func TestNoFastStorageUpgrade_Integration_SaveVersion_Load_Iterate_Success(t *te
...
+	require.NoError(t, err)
+
+	rootHash, version, err = tree.SaveVersion()
+	require.NoError(t, err)
+	require.Equal(t, initialVersion+1, version)
+
+	// the following versions behaviors normally
+	node, err = tree.ndb.GetNode(rootHash)
+	require.NoError(t, err)
+	require.Equal(t, int64(initialVersion+1), node.version)
```
yihuang, line 1570:
```suggestion
	require.Equal(t, initialVersion+1, node.version)
```

mutable_tree_test.go
```diff
@@ -1533,3 +1533,39 @@ func TestNoFastStorageUpgrade_Integration_SaveVersion_Load_Iterate_Success(t *te
...
+	require.Equal(t, int64(1), node.version)
+
+	_, err = tree.Set([]byte("hello"), []byte("world1"))
+	require.NoError(t, err)
+
+	rootHash, version, err = tree.SaveVersion()
+	require.NoError(t, err)
+	require.Equal(t, initialVersion+1, version)
+
+	// the following versions behaviors normally
```
yihuang, line 1567:
```suggestion
	// the following versions behaves normally
```

cool-develope:
It is a kind of `should be fixed`, isn't it?

yihuang:
> It is a kind of `should be fixed`, isn't it?

I think it's hard to fix because of legacy data, maybe we can fix in the 2.0

tac0turtle:
merging but agree we should fix, want to open an issue?

yihuang:
> merging but agree we should fix, want to open an issue?

My concern of updating this is this is a root hash breaking change.
So far I'm able to replay our change sets since genesis to reproduce same root hash with a single algorithm, I find it surprisingly lucky, it means we haven't done any breaking change at this level since our genesis, this'll be the first one.

cool-develope:
> > merging but agree we should fix, want to open an issue?
> 
> My concern of updating this rule is this is a root hash breaking change. So far I'm able to replay our change sets since genesis to reproduce same root hash with the same algorithm, I find it surprisingly lucky, it means we haven't done any breaking change at this level since our genesis, this'll be the first one.

yeah, you are right, the `version` participate in the hash and proof calc

PR #662, opened 2023-01-17
feat: refactor the export traversal order as pre-order

cool-develope:
ref: #608, #656

We are using post-order traversal in the `Export`, but it violates [adr-001](#608), not being able to make a path from `ExportNode`.
It refactored `Export/Import` to provide both traversal orders `pre-order` and `post-order`

### DIFF
--- a/export.go
+++ b/export.go
@@ -6,6 +6,14 @@ import (
 	"fmt"
 )
 
+type OrderType int
+
+// OrderTraverse is the type of traversal order to use when exporting and importing.
+const (
+	PreOrderTraverse OrderType = iota
+	PostOrderTraverse
+)
+
 // exportBufferSize is the number of nodes to buffer in the exporter. It improves throughput by
 // processing multiple nodes per context switch, but take care to avoid excessive memory usage,
 // especially since callers may export several IAVL stores in parallel (e.g. the Cosmos SDK).
@@ -27,17 +35,17 @@ type ExportNode struct {
 
 // Exporter exports nodes from an ImmutableTree. It is created by ImmutableTree.Export().
 //
-// Exported nodes can be imported into an empty tree with MutableTree.Import(). Nodes are exported
-// depth-first pre-order (NLR), this order must be preserved when importing in order to recreate
-// the same tree structure.
+// Exported nodes can be imported into an empty tree with MutableTree.Import(). Nodes are exported in traverseOrder,
+// this order must be preserved when importing in order to recreate the same tree structure.
 type Exporter struct {
-	tree   *ImmutableTree
-	ch     chan *ExportNode
-	cancel context.CancelFunc
+	tree          *ImmutableTree
+	traverseOrder OrderType
+	ch            chan *ExportNode
+	cancel        context.CancelFunc
 }
 
 // NewExporter creates a new Exporter. Callers must call Close() when done.
-func newExporter(tree *ImmutableTree) (*Exporter, error) {
+func newExporter(tree *ImmutableTree, traverseOrder OrderType) (*Exporter, error) {
 	if tree == nil {
 		return nil, fmt.Errorf("tree is nil: %w", ErrNotInitalizedTree)
 	}
@@ -48,9 +56,10 @@ func newExporter(tree *ImmutableTree) (*Exporter, error) {
 
 	ctx, cancel := context.WithCancel(context.Background())
 	exporter := &Exporter{
-		tree:   tree,
-		ch:     make(chan *ExportNode, exportBufferSize),
-		cancel: cancel,
+		tree:          tree,
+		traverseOrder: traverseOrder,
+		ch:            make(chan *ExportNode, exportBufferSize),
+		cancel:        cancel,
 	}
 
 	tree.ndb.incrVersionReaders(tree.version)
@@ -61,22 +70,41 @@ func newExporter(tree *ImmutableTree) (*Exporter, error) {
 
 // export exports nodes
 func (e *Exporter) export(ctx context.Context) {
-	e.tree.root.traverse(e.tree, true, func(node *Node) bool {
-		exportNode := &ExportNode{
-			Key:     node.key,
-			Value:   node.value,
-			Version: node.version,
-			Height:  node.subtreeHeight,
-		}
+	defer close(e.ch)
+	switch e.traverseOrder {
+	case PreOrderTraverse:
+		e.tree.root.traverse(e.tree, true, func(node *Node) bool {
+			exportNode := &ExportNode{
+				Key:     node.key,
+				Value:   node.value,
+				Version: node.version,
+				Height:  node.subtreeHeight,
+			}
 
-		select {
-		case e.ch <- exportNode:
-			return false
-		case <-ctx.Done():
-			return true
-		}
-	})
-	close(e.ch)
+			select {
+			case e.ch <- exportNode:
+				return false
+			case <-ctx.Done():
+				return true
+			}
+		})
+	case PostOrderTraverse:
+		e.tree.root.traversePost(e.tree, true, func(node *Node) bool {
+			exportNode := &ExportNode{
+				Key:     node.key,
+				Value:   node.value,
+				Version: node.version,
+				Height:  node.subtreeHeight,
+			}
+
+			select {
+			case e.ch <- exportNode:
+				return false
+			case <-ctx.Done():
+				return true
+			}
+		})
+	}
 }
 
 // Next fetches the next exported node, or returns ExportDone when done.
--- a/export.go
+++ b/export.go
@@ -9,6 +9,7 @@ import (
 type OrderType int
 
 // OrderTraverse is the type of traversal order to use when exporting and importing.
+// PreOrder is needed for the new node-key refactoring. The default is PostOrder.
 const (
 	PreOrderTraverse OrderType = iota
 	PostOrderTraverse
--- a/export.go
+++ b/export.go
@@ -28,7 +28,7 @@ type ExportNode struct {
 // Exporter exports nodes from an ImmutableTree. It is created by ImmutableTree.Export().
 //
 // Exported nodes can be imported into an empty tree with MutableTree.Import(). Nodes are exported
-// depth-first post-order (LRN), this order must be preserved when importing in order to recreate
+// depth-first pre-order (NLR), this order must be preserved when importing in order to recreate
 // the same tree structure.
 type Exporter struct {
 	tree   *ImmutableTree
@@ -61,7 +61,7 @@ func newExporter(tree *ImmutableTree) (*Exporter, error) {
 
 // export exports nodes
 func (e *Exporter) export(ctx context.Context) {
-	e.tree.root.traversePost(e.tree, true, func(node *Node) bool {
+	e.tree.root.traverse(e.tree, true, func(node *Node) bool {
 		exportNode := &ExportNode{
 			Key:     node.key,
 			Value:   node.value,
--- a/immutable_tree.go
+++ b/immutable_tree.go
@@ -155,8 +155,8 @@ func (t *ImmutableTree) Hash() ([]byte, error) {
 
 // Export returns an iterator that exports tree nodes as ExportNodes. These nodes can be
 // imported with MutableTree.Import() to recreate an identical tree.
-func (t *ImmutableTree) Export() (*Exporter, error) {
-	return newExporter(t)
+func (t *ImmutableTree) Export(traverseOrder OrderType) (*Exporter, error) {
+	return newExporter(t, traverseOrder)
 }
 
 // GetWithIndex returns the index and value of the specified key if it exists, or nil and the next index
--- a/immutable_tree.go
+++ b/immutable_tree.go
@@ -155,8 +155,16 @@ func (t *ImmutableTree) Hash() ([]byte, error) {
 
 // Export returns an iterator that exports tree nodes as ExportNodes. These nodes can be
 // imported with MutableTree.Import() to recreate an identical tree.
-func (t *ImmutableTree) Export(traverseOrder OrderType) (*Exporter, error) {
-	return newExporter(t, traverseOrder)
+// The default export order is PostOrderTraverse.
+func (t *ImmutableTree) Export() (*Exporter, error) {
+	return newExporter(t, PostOrderTraverse)
+}
+
+// Export returns an iterator that exports tree nodes as ExportNodes with pre-order traversal.
+// This api is for new node-key refactoring since we need to build a new tree with pre-order traversal
+// to construct the new node-key (path)
+func (t *ImmutableTree) ExportPreOrder() (*Exporter, error) {
+	return newExporter(t, PreOrderTraverse)
 }
 
 // GetWithIndex returns the index and value of the specified key if it exists, or nil and the next index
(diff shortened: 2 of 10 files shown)


cool-develope:
@yihuang , sorry I referred your PR (#656), there is no way to contribute to your PR. 
Please review it.

yihuang:
Oh, I don't like this choice, post-order is perfect for snapshot.
I don't understand why using path as nonce in the first place, but since it don't affect snapshot format, I can ignore that.

yihuang:
> > That's why I don't like the choice of using path as nonce, I don't see the benefits yet, but I see the troubles 😂
> 
> I can see some significant advantages of the `path`, one thing is parallel restricting of the tree because assigning path of left child tree and right child tree would be independent. 

can you elaborate on this?

> The other thing is we can skip child node keys for the same version. I believe it will reduce the entire storage. Those are the advantages of pre-order iterating.

I think that'd some insignificant save if there's any at all:
- the version field can always be saved if you like, no matter what nonce strategy used.
- using path as nonce make the nonce bigger than a simply sequential one, which means more bytes with variable-length integer encoding.
- for most of the nodes only one of the children's key is saved.
- the saved bytes if there's any is insignificant compared to the hash field(32bytes) and key field(usually dozens of bytes).

yihuang:
> > can you elaborate on this?
> 
> OK, for example when commit the branch, we are trying to assign the path for new nodes with iterating the tree. we can do this iterating and assigning parallelly, there would be more use cases

again, iteration and assigning nonce is trivial when committing branch, computing hash is the heavy one, if it can help with hash computation, that'd be very useful.

> 
> > the version field can always be saved if you like, no matter what nonce strategy used.
> 
> it's not true, now we have `leftNodeKey` and `rightNodeKey` instead of `leftHash` and `rightHash`, we will save the whole child node key

node key is just a tuple `(version, nonce)`, right? I mean that version field, compared with other nonce assignment strategy.

> 
> > using path as nonce make the nonce bigger than a simply sequential one, which means more bytes under variable-length integer encoding.
> 
> `version` is uint64 and it requires 8bytes, I think 8bytes is enough for path, 8bytes = 64bit (64 height in tree)

I assume we do variable length integer encoding here, so larger integers can take more space on average.

cool-develope:
> again, iteration and assigning nonce is a trivial part during committing branch, computing hash is the heavy one, if it can parallelize hash computation, that'd be very useful.

right, I think we can parallelize hash calc with assigning the path (sequence id is impossible)

regarding (version, nonce), I think there would be a problem in encoding/decoding of nodes if we use empty methods for version

anyhow, if we use the sequence id as a nonce, post-order would be a problem in export/import. and I think it is not a good place for this topic.

cool-develope:
I am just thinking pre-order is better than post-order at least in our iavl, I can't find the reason why use post-order

cool-develope:
> > I am just thinking pre-order is better than post-order at least in our iavl, I can't find the reason why use post-order
> 
> my biggest concern is actually consensus breaking, without using path as nonce, the new node key format is not a consensus breaking change, that'd make it much more easier to rollout to the network node to node asynchronously.

what is your plant to migrate to new version?

yihuang:
> > > I am just thinking pre-order is better than post-order at least in our iavl, I can't find the reason why use post-order
> 
> > 
> 
> > my biggest concern is actually consensus breaking, without using path as nonce, the new node key format is not a consensus breaking change, that'd make it much more easier to rollout to the network node to node asynchronously.
> 
> 
> 
> what is your plan to migrate to new version?
> 
> 
> 
> my idea is we can restrict the storage using export/import, there is a no way of soft landing

I'm striving for a non-consensus breaking version that just do storage optimization, the first step is versiondb running alongside with existing iavl tree, the second step should be optimize iavl tree itself, I think there are lots of potential already before introducing breaking stuff.

yihuang:
> right, I think we can parallelize hash calc with assigning the path (sequence id is impossible)

I think parallel hash computation is a very interesting topic, the difficult part is most of the time we create branches rather than full sub-trees. A naïve implementation could easily end up slower than a sequential one, because the potential for parallel is low. But I don't see why we can't do that with sequential nonce, we can do sequential iteration while distributing the task of hash.
I'll research parallel hash computation more, it'll be useful in our change set verification step.

cool-develope:
@yihuang 
OK, I will close this PR after reflecting the idea in the node-key refactoring branch, let's discuss further in the next storage meeting

cool-develope:
@yihuang , I just remembered why the `version + nonce` is not working in the `Import`.
There is no way to get the current sequence integer of the given version, you can't this status in memory (like map[int64]int32) because there could be massive versions (int64)

yihuang:
> @yihuang , I just remembered why the `version + nonce` is not working in the `Import`. There is no way to get the current sequence integer of the given version, you can't keep this status in memory (like map[int64]int32) because there could be massive versions (int64)

May I know what's the plan to rebuild the path even with pre-order export? the path need to be the path in the version node created in, not the exported version, right?

yihuang:
> > May I know what's the plan to rebuild the path even with pre-order export? the path need to be the path in the version node created in, not the exported version, right?
> 
> 
> 
> I plan to build the current node path based on the parent one even it is inherited from the different version.

Isn't that breaks assumption of path design?

cool-develope:
@yihuang , how about adding a flag to denote if this export is post-order or pre-order? then it would not be a consensus breaking, right?

yihuang:
> @yihuang , how about adding a flag to denote if this export is post-order or pre-order? then it would not be a consensus breaking, right?

there's format version field in state sync snapshot, will the new node key format support both format?

cool-develope:
> > @yihuang , how about adding a flag to denote if this export is post-order or pre-order? then it would not be a consensus breaking, right?
> 
> there's format version field in state sync snapshot, will the new node key format support both format? if that's the case, then it don't breaks anything.

I have no exact idea how to interact within cosmos-sdk (with snapshot format), but that's true, the iavl will provide both post-order, pre-order, how about this @tac0turtle ?

cool-develope:
@yihuang , done please review again.
@tac0turtle , do we need to update the cosmos/store?

tac0turtle:
We will update store, when this version is released in alpha/beta.

export.go
```diff
@@ -6,6 +6,14 @@ import (
 	"fmt"
 )
 
+type OrderType int
+
+// OrderTraverse is the type of traversal order to use when exporting and importing.
```
tac0turtle, line 11: Could we provide short reason in the Godoc which is preferred and which is backwards compatible?

tac0turtle (commented):
My understanding of this pr is the new node key refactor works with both post and pre order import, if so do we need both? I could be lacking the understanding the need for both if the node key refactor will be merged

tac0turtle (approved):
LGTM left one request for a godoc, after that lets merge this

cool-develope:
@kocubinski @yihuang 
I updated it to provide both `pre-order` and `post-order`, please review it!

export.go
yihuang: agree, what's the advantage of preorder type and why people should use it?

immutable_tree.go
```diff
@@ -155,8 +155,8 @@ func (t *ImmutableTree) Hash() ([]byte, error) {
 
 // Export returns an iterator that exports tree nodes as ExportNodes. These nodes can be
 // imported with MutableTree.Import() to recreate an identical tree.
-func (t *ImmutableTree) Export() (*Exporter, error) {
-	return newExporter(t)
+func (t *ImmutableTree) Export(traverseOrder OrderType) (*Exporter, error) {
```
kocubinski, line 158: Do we really need an API breaking change here? Maybe adding a new method `ExportPreOrder` would be better.

mutable_tree.go
```diff
@@ -156,8 +156,8 @@ func (tree *MutableTree) Get(key []byte) ([]byte, error) {
 //
 // Import can only be called on an empty tree. It is the callers responsibility that no other
 // modifications are made to the tree while importing.
-func (tree *MutableTree) Import(version int64) (*Importer, error) {
-	return newImporter(tree, version)
+func (tree *MutableTree) Import(version int64, traverseOrder OrderType) (*Importer, error) {
```
kocubinski, line 159: Same as https://github.com/cosmos/iavl/pull/662#discussion_r1092063909, I'm not sure we should break the Import API here.  `ImportPreOrder` might be better.

immutable_tree.go
cool-develope: that sounds great

import.go
```diff
@@ -83,82 +122,57 @@ func (i *Importer) Add(exportNode *ExportNode) error {
 		version:       exportNode.Version,
 		subtreeHeight: exportNode.Height,
 	}
-
+	if node.subtreeHeight == 0 {
```
kocubinski, line 125:
```suggestion
    // set leaf nodes subtree size = 1
	if node.subtreeHeight == 0 {
```

node.go
```diff
@@ -321,7 +321,7 @@ func (node *Node) validate() error {
 		if node.value != nil {
 			return errors.New("value must be nil for non-leaf node")
 		}
-		if node.leftHash == nil && node.rightHash == nil {
+		if node.leftHash == nil || node.rightHash == nil {
```
kocubinski, line 324: Could you explain this change, it's more restrictive than previous, right?

import.go
```diff
@@ -83,82 +122,57 @@ func (i *Importer) Add(exportNode *ExportNode) error {
...
-	// We don't modify the stack until we've verified the built node, to avoid leaving the
-	// importer in an inconsistent state when we return an error.
-	stackSize := len(i.stack)
-	switch {
-	case stackSize >= 2 && i.stack[stackSize-1].subtreeHeight < node.subtreeHeight && i.stack[stackSize-2].subtreeHeight < node.subtreeHeight:
-		node.leftNode = i.stack[stackSize-2]
-		node.leftHash = node.leftNode.hash
-		node.rightNode = i.stack[stackSize-1]
-		node.rightHash = node.rightNode.hash
-	case stackSize >= 1 && i.stack[stackSize-1].subtreeHeight < node.subtreeHeight:
```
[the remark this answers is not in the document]
cool-develope: it is related to #656, the inner nodes always have two children

node.go
cool-develope: #656

export.go
```diff
@@ -6,6 +6,15 @@ import (
...
+// PreOrder is needed for the new node-key refactoring. The default is PostOrder.
```
yihuang, line 12: I thought new node-key refactoring works with both orderings?

cool-develope: no, the pre-order is needed for node-key refactoring

tac0turtle: slightly confused now. This pr adds support for both pre and post order, but node key refactor require pre-order. If a node exports using post-order then we cant import it into the new version correct?

cool-develope: yes, that's why provides both orders

cool-develope: even it is the original version, we can request a pre-order snapshot, then import it into the new version

tac0turtle: I don't think this flow works with the sdk. Users will snapshot what they know, post-order, then when a node using this new version requests chunks, im not sure it can ask for pre or post order, it just gets what the other node has. I believe this flow has a negotiation between peers on which snapshot?

cool-develope: If we can add the param which denote pre-order or post-order in Snapshot of cosmos-sdk, we don't need negotiation.

- Firstly, we create the node with original version and sync the original snapshot (post-order) and save (pre-order) snapshot (it is possible if we merge this PR, since we provide both-way import/export)
- Every nodes which want new version start sync from the created node, and get (pre-order) snapshot and they can rebuild the new version iavl tree

yihuang: Currently, in the state-sync protocol, node can fetch the snapshot chunks from multiple nodes, they are expected to be consistent and deterministic, there's no way for client node to filter different snapshot formats.

cool-develope: but we can point the source nodes in the config, right? In the second step, we can set the source nodes as the new created nodes which provide pre-order snapshot

tac0turtle: so I think this makes sense, cause on start of the node, it will export using pre-order and ingest the same snapshot to recreate state and move the chain ahead. The question becomes how to support historical queries to the older state or if that is a needed feature? 

This would mean that a chain that upgraded to this version would share updated snapshots. The only question comes from a node using an older version of iavl which I don't think we will support in realtime, but even the old way can be done pre-order correct?

yihuang: I think the reason post-order is chosen in the first place is the node hash is updated in a post-order way, you need to update the children first to update the parent node, will pre-order import need more temporary memory?

cool-develope: yeah, literally post-order more makes sense, both way requires stack to keep the current path, you are right pre-order will require 2 times memory, but the stack length is at most the height of the tree, it is so trivial

cool-develope: @tac0turtle , the old version can use any order, the new version should use pre-order.

yihuang: > but the stack length is at most the height of the tree, it is so trivial

yeah, that should be trivial then, if this is indeed necessary, I think the chains can do a coordinated upgrade in advance to switch to pre-order snapshots, before doing node key format migration.

cool-develope: > yeah, that should be trivial then, if this is indeed necessary, I think the chains can do a coordinated upgrade in advance to switch to pre-order snapshots, before doing node key format migration.

I think this is what we should discuss, tbh I have no clear idea which way is more efficient

cool-develope:
@yihuang https://github.com/cosmos/iavl/pull/662#issuecomment-1400749296

cool-develope:
> > @yihuang , I just remembered why the `version + nonce` is not working in the `Import`. There is no way to get the current sequence integer of the given version, you can't keep this status in memory (like map[int64]int32) because there could be massive versions (int64)
> 
> In this case, we can at least use a global increasing unique nonce, similar to how you have to use the path in current version (instead of node creation version).

It might lead to several problems in the implementation, `int32` is enough in the tree operation, but it will be int64 or `big.Int` in export/import

yihuang:
> > > @yihuang , I just remembered why the `version + nonce` is not working in the `Import`. There is no way to get the current sequence integer of the given version, you can't keep this status in memory (like map[int64]int32) because there could be massive versions (int64)
> > 
> > 
> > In this case, we can at least use a global increasing unique nonce, similar to how you have to use the path in current version (instead of node creation version).
> 
> It might lead to several problems in the implementation, `int32` is enough in the tree operation, but it will be int64 or `big.Int` in export/import

A continuously increasing number is at least smaller than the path which is also a unique integer between the nodes, right?

yihuang:
> There is no way to get the current sequence integer of the given version, you can't keep this status in memory (like map[int64]int32) because there could be massive versions (int64)

But in practice, there are only millions or tens of millions of versions, a continuous array of 10millions of `int32`s is only: `(10000000 * 4) / 1024/1024 = 38.15` megabytes, not something unimaginable I think.

cool-develope:
> But in practice, there are only millions or tens of millions of versions, a continuous array of 10millions of `int32`s is only: `(10000000 * 4) / 1024/1024 = 38.15` megabytes, should be practical I think.

That makes sense, it will be at most hundreds MG, I will implement version + local nonce, let's do some benchmarks

tac0turtle:
closing this as we dont need pre-order reconstruction. We discussed this yesterday on the storage working group call

PR #664, opened 2023-01-23
perf: Improve IAVL Import speed for state sync by 2x~3x

yzang2019:
**Problem Statement**
Currently the cosmos state sync bottleneck when restoring the offered snapshot chunks is on the IAVL Store side. 

This is because in order to restore all the chunks from the snapshot, we need to load all the items from the chunk data into the IAVL Store to rebuild the tree from bottom up in order and do this one chunk at a time. Since rebuilding the tree requires strict ordering guarantee, all the steps to import a new node into tree is now processed by a single goroutine. 

While single threaded model guarantees correctness, this is extremely inefficient and will not be able to fully utilize multicores or faster disk IOPs to speed things up. In certain scenarios when we want to bootstrap a new pruned node as fast as possible (e.g. aws autoscaling group), the slowness of state sync will become a bottleneck and blocker. 

**Context**
In the currently architecture, whenever we import a new node to the IAVL tree, it needs to go through the following process:

1. Build the tree structure using a stack (fast)
2. Compute the hash of the node (fast)
3. Validate the node contents (fast)
4. Serialize the node data into byte array (medium)
5. Append node data to the current batch (heavy ops)
6. If batch size is full, flush the whole batch and rotate a new batch (heavy ops)
7. Update the stack (fast)

One major problem of using a single thread for all these step is that step 5 and step 6 could become really slow and when a batch write takes dozens of seconds or up till a minute to complete, the whole thread becomes idle, waiting for the previous batch write IO to complete.

**Proposal**

- Optimize threading model
- Optimize batch write efficiency with larger configurable batch size

The proposal is to try our best to parallelize the node import steps while still keep the minimum ordering guarantee needed to rebuild the IAVL Tree. Upon deep investigation, we found out that the logic of building the tree from stack , and the remaining logic to persist node data can happen in parallel. That means we can use a few goroutines with channels to parallel processing the nodes as well as maintaining the necessary order for each step. 

Another biggest gain is by using a larger batch write size. We find out different batch size would impact total time by a lot.  With a proper tuned batch size, we will be able to speed up the import by a lot.

IAVL drawio (3)

Here's what the threading model would look like:
**Add**: [Step1 -> Step2 -> Step3 -> Step4 -> channelA -> Step7 
**goroutineA**: [channelA -> Step5 -> channelB] (Act as the buffer to reduce pre processing time)
**goroutineC**: [channelC -> Step6] (Do the batch write in the background)

With this new threading model, when we are busy doing disk IO write, our CPU will not become idle and can continuously doing its job to build the tree structure, compute hash and perform data serializations, as well as append node data to the next batch in parallel. This would save us huge amount of time compare to the past where all the remaining work has to stop and wait for the batch to complete.

Of course, there's still a certain limitation we will hit. We have limited memory and we can not buffer everything into the channel. When the batch write becomes really really slow, the channel buffer might still get filled up, the side effect would be that we will see memory usage increased and the next batch size becomes really large. But with the assumption of bigger batches is more efficient than small batches, we think this will still benefit us in some way to improve the speed of the whole process. 

**Benchmark**
We've done a few benchmark for sei atlantic-1 chain. Here's the benchmark result:

Testing Setup:

- Total compressed snapshot size: 650MB
- Total state key-value entries: 28,000,000
- Total decompressed state size: 7GB

Testing Result:
<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-5922dc77-7fff-f50f-2842-d90c37bf06b7"><div dir="ltr" style="margin-left:0pt;" align="left">

Instance Type | CPU | RAM | Instance Cost | Boot Drive | 1.2.2 Beta | 1.2.2Beta With IAVL Patch
-- | -- | -- | -- | -- | -- | --
c6a.4xlarge | 16 | 32G | 0.76/h | No | 15 min | 5 min
c6a.2xlarge | 8 | 16G | 0.38/h | No | 16 min | 5 min 30s
m5.4xlarge | 16 | 64G | 0.89/h | No | 21 min | 8 min
m5.4xlarge | 16 | 64G | 0.89/h | Yes | 26 min | 14 min
m4.2xlarge | 8 | 32G | 0.47/h | No | 23 min | 8 min 30s
m4.2xlarge | 8 | 32G | 0.47/h | Yes | 27 min | 15 min

</div></b>

**Key Takeaway:**

- With this IAVL patch, we are able to reduce total state sync time by up to 60% - 67%
- CPU per core speed matters more than num of cores and ram size for state sync
- Whether storing data on snapshot boot drive or on a newly created EBS volume also matters a lot


import.go
```diff
@@ -44,17 +59,107 @@ func newImporter(tree *MutableTree, version int64) (*Importer, error) {
...
+
+// serializeNode get the next validated and hashed node from the channel
+// and dump the node to byte buf, push the serialized node data to another channel
+func serializeNode(i *Importer) {
+	for i.batch != nil {
+		if currNode, open := <-i.chNode; open {
+			buf := bufPool.Get().(*bytes.Buffer)
+			buf.Reset()
+			if err := currNode.writeBytes(buf); err != nil {
+				panic(err)
```
kocubinski, line 97: How does this change impact the application's error handling holistically? Previously we'd get an error back from `Importer.Add` if there was a problem, now we'll have a panic in a goroutine.

import.go
```diff
@@ -44,17 +59,107 @@ func newImporter(tree *MutableTree, version int64) (*Importer, error) {
...
+			if i.batch != nil {
+				err := i.batch.Set(i.tree.ndb.nodeKey(nodeData.node.hash), nodeData.data)
+				if err != nil {
+					panic(err)
+				}
+			}
+			i.batchMtx.RUnlock()
+			i.batchSize++
+			// Only commit a new batch if size meet desiredBatchSize and there's no pending batch write
+			if i.batchSize >= desiredBatchSize && len(i.chBatch) < 1 {
```
kocubinski, line 128: If I/O is blocked for any reason then the batch could inflate (in memory) to the size of the remaining dataset to import (on disk) right?

kocubinski (changes requested):
Thank for the contribution! Leveraging concurrency is a good direction, and impressive improvements and benchmarks.

Just some questions about error handling and how we're buffering batches if write is blocked.  I think we should limit that somehow.

Also tests now appear to breaking with
```
WARNING: DATA RACE
Read at 0x00c01b1f2650 by goroutine 399:
```

yihuang:
Looks great. But I guess step 4 is pretty fast, maybe not justified a separate goroutine for it.

import.go
yzang2019: Good point! I think a better solution would be to send the error into an error channel and return that through the Importer.Add function. What do you think?

import.go
yzang2019: That is a valid concern, we can add back max batch size to prevent unbounded memory usage, so batch size would be controlled between desired size and max size, what do you think?

yzang2019:
> Thank for the contribution! Leveraging concurrency is a good direction, and impressive improvements and benchmarks.
> 
> Just some questions about error handling and how we're buffering batches if write is blocked. I think we should limit that somehow.
> 
> Also tests now appear to breaking with
> 
> ```
> WARNING: DATA RACE
> Read at 0x00c01b1f2650 by goroutine 399:
> ```

Yeah, I'll look into the test failures, thanks for taking a look and providing the feedbacks!

yzang2019:
> Looks great. But I guess step 4 is pretty fast, maybe not justified a separate goroutine for it. And I guess the real bottleneck is the bulk loading the data into db itself, current solution runs it in parallel, it definitely helps. there are also some solutions to speed up the data ingestion itself, for example there are several options for rocksdb[1](#user-content-fn-1-51625c4474984fc9c355a35db2cf1544), IMO the best solution for rocksdb is writing out sst files separately using `SSTFileWriter` (you can even write multiple files concurrently if necessary), then ingest them into db in one step.
> 
> ## Footnotes
> 1. https://rockset.com/blog/optimizing-bulk-load-in-rocksdb/ [↩](#user-content-fnref-1-51625c4474984fc9c355a35db2cf1544)

I'm not that familiar with RocksDB, but if we can improve the ingestion itself, it's definitely going to help the most. AFAIK, we provide users to choose their own underline database implementation, so optimizing for one type of DB might not help for other DB engine. However, if RocksDB outperforms other DB a lot, we can switch the default DB engine to RocksDB and recommend users to migrate?

yzang2019:
Addressed most of the comments, could you please take another look? Thanks!

kocubinski (approved):

import.go
```diff
@@ -125,28 +224,25 @@ func (i *Importer) Add(exportNode *ExportNode) error {
 
 	buf := bufPool.Get().(*bytes.Buffer)
 	buf.Reset()
-	defer bufPool.Put(buf)
-
-	if err = node.writeBytes(buf); err != nil {
-		return err
+	if err := node.writeBytes(buf); err != nil {
+		panic(err)
```
cool-develope, line 228: why `panic` here?

yzang2019: ah good catch, let me fix it

yzang2019: @cool-develope Should be fixed now! Please take another look, thanks!

tac0turtle:
sorry for delay, we will merge this work after merging the node key refactor. The rebase should not be large if at all. Thank you for the PR

tac0turtle:
@yzang2019 could you rebase and then we can take care of merging this

import.go
```diff
@@ -44,19 +65,97 @@ func newImporter(tree *MutableTree, version int64) (*Importer, error) {
...
+
+	importer.chBatchWg.Add(1)
+	go batchWrite(importer)
+
+	return importer, nil
+}
+
+// WithDesiredBatchSize set the desired batch size for write
+func (i *Importer) WithDesiredBatchSize(batchSize uint32) *Importer {
+	i.desiredBatchSize = batchSize
```
elias-orijtech, line 95: The mutability of `desiredBatchSize` and `maxBatchSize` during an ongoing import seems like too much complexity for little gain.

import.go
```diff
@@ -44,19 +65,97 @@ func newImporter(tree *MutableTree, version int64) (*Importer, error) {
...
+	return i
+}
+
+// WithMaxBatchSize set the maximum allowed batch size for write, should be greater than desired batch size.
+// Consider increase max batch size to reduce overall import time.
+func (i *Importer) WithMaxBatchSize(batchSize uint32) *Importer {
+	i.maxBatchSize = batchSize
```
elias-orijtech, line 102: Same.

import.go
```diff
@@ -44,19 +65,97 @@ func newImporter(tree *MutableTree, version int64) (*Importer, error) {
...
+		stack:            make([]*Node, 0, 8),
+		batchMtx:         sync.RWMutex{},
+		desiredBatchSize: defaultDesiredBatchSize,
+		maxBatchSize:     defaultMaxBatchSize,
+		chNodeData:       make(chan NodeData, 2*defaultDesiredBatchSize),
+		chNodeDataWg:     sync.WaitGroup{},
+		chBatch:          make(chan db.Batch, 1),
+		chBatchWg:        sync.WaitGroup{},
+		chError:          make(chan error, 1),
+		allChannelClosed: false,
```
elias-orijtech, line 81: Without digging into details, this seems like a lot of state to keep 2 goroutines in check. See also my PR comment.

elias-orijtech (commented):
The speedup and general approach seems good. Two issues:

1. A test would be nice. From a glance at `import_test.go` I see only tests for cases for small tree and edge cases, which doesn't exercise this new code much.
2.  `Import` is (partially) mutable during an ongoing import. That seems like unwanted complexity to me, and would explain the (some of) the rather larger amount of new state needed for managing the new goroutines.

I suggest a design where you set up parameters, and *then* start the import, which can from then on only be `Cancel`ed or `Commit`ed. That may involve an `ImportConfig` type, or you could even make `desiredBatchSize` and `desiredMaxBatchSize` non-configurable by the YAGNI principle.

tac0turtle:
@yzang2019 can you help update and address comments please

yzang2019:
> @yzang2019 can you help update and address comments please

Yes, let me do a rebase to merge the latest changes and address the comments

fedekunze:
@yihuang @yzang2019 would be great to include this in the next release

tac0turtle:
yup we will include this in the 1.0 release. abci++ work in the sdk is concluding so there is more time for these things

import.go
```diff
@@ -44,19 +65,97 @@ func newImporter(tree *MutableTree, version int64) (*Importer, error) {
...
+				i.batchSize = 0
+			}
+		} else {
+			break
+		}
+	}
+	i.chNodeDataWg.Done()
+}
+
+// batchWrite get a new batch from the channel and execute the batch write to the underline DB.
```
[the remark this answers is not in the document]
yzang2019: Fixed

import.go
yzang2019: Addressed

import.go
yzang2019: Moved them to another struct called ChannelConfig

cool-develope (approved):
LGTM!
Could you please add the benchmark codes? Also, it is preferred to add some tests of the batch channel with small custom config, since the default config is too large

yzang2019:
> LGTM! Could you please add the benchmark codes? Also, it is preferred to add some tests of the batch channel with small custom config, since the default config is too large

what benchmark code shall I add? I can see there's already a BenchmarkImport test?

elias-orijtech (commented):
I still think this PR is more complicated than need be. From your PR description, step (5) and (6) are expensive,

> (5) Append node data to the current batch (heavy ops)
> (6) If batch size is full, flush the whole batch and rotate a new batch (heavy ops)

and the goal of the PR is to start a batch (5) while the previous batch is in (6). Correct?

If so, I don't see the need for 2 extra goroutines. I think (5) can be done as before this PR (synchronously from `Importer.Add`) and only the commit phase (6) needs a separate goroutine.

In the detail, there still seem to be too much synchronization state. For example, why have `batchMtx` when you communicate with the worker goroutines through `chNodeData` and `chBatch`? Similarly, why have the `WaitGroup`s `chNodeDataWg` and `chbatchWg` when you can close, say, `chBatch` and wait for a value on `chError`?

If one goroutine is enough, I expect the only state you need is a `chan error`. A sketch:

```go
   type Importer struct {
      ...
      commitBatch <-chan error
      ...
   }

   func (i *Importer) Add(exportNode *ExportNode) error {
      ...
      i.writeNode(...)
      if batchSize >= commitBatchSize {
          // Wait for previous batch.
          if i.commitBatch != nil {
             if err := <-i.commitBatch; err != nil {
                return err
             }
          }
          result := make(chan error)
          i.commitBatch = result
          go i.commitBatch(result, i.batch)
          // Start new batch.
          i.batch = ...
      }
   }
```

tac0turtle:
hey we recently merged a fix that helps with performance, would you want to test that and see how it compares

zsystm:
@yzang2019 
How can I setup testing environment for benchmark? 

```
Testing Setup:

Total compressed snapshot size: 650MB
Total state key-value entries: 28,000,000
Total decompressed state size: 7GB
```

import.go
```diff
@@ -45,57 +82,121 @@ func newImporter(tree *MutableTree, version int64) (*Importer, error) {
...
+	return importer, nil
+}
+
+// setBatchData get the next serialized node data from channel, and write the data to the current batch
+func setBatchData(i *Importer) {
+	for i.batch != nil {
+		if nodeData, open := <-i.chanConfig.chNodeData; open {
+			i.batchMtx.RLock()
+			if i.batch != nil {
+				err := i.batch.Set(i.tree.ndb.nodeKey(nodeData.node.GetKey()), nodeData.data)
```
odeke-em, line 110: This is a mutating operation being invoked under a read lock. One do the reasons why it is important to just use a plain mutex, and also please see https://cyber.orijtech.com/scsec/cosmos-go-coding-guide#avoid-syncrwmutex

import.go
```diff
@@ -45,57 +82,121 @@ func newImporter(tree *MutableTree, version int64) (*Importer, error) {
...
+	importer.chanConfig.chBatchWg.Add(1)
+	go batchWrite(importer)
+
+	return importer, nil
+}
+
+// setBatchData get the next serialized node data from channel, and write the data to the current batch
+func setBatchData(i *Importer) {
+	for i.batch != nil {
+		if nodeData, open := <-i.chanConfig.chNodeData; open {
```
odeke-em, line 107: You can reduce nesting and ease reading by inverting this condition per

```go
nodeData, open := ...
if !open {
   break
}

i.batchMtx.Lock()
...
```

import.go
```diff
@@ -45,57 +82,121 @@ func newImporter(tree *MutableTree, version int64) (*Importer, error) {
...
+			i.batchMtx.RLock()
+			if i.batch != nil {
+				err := i.batch.Set(i.tree.ndb.nodeKey(nodeData.node.GetKey()), nodeData.data)
+				if err != nil {
+					i.batchMtx.RUnlock()
+					i.chanConfig.chError <- err
+					break
+				}
+			}
+			i.batchSize++
```
odeke-em, line 117: Another mutation while a read lock is being held

import.go
```diff
@@ -45,57 +82,121 @@ func newImporter(tree *MutableTree, version int64) (*Importer, error) {
...
+			if (i.batchSize >= i.chanConfig.desiredBatchSize && len(i.chanConfig.chBatch) < 1) || i.batchSize >= i.chanConfig.maxBatchSize {
+				i.chanConfig.chBatch <- i.batch
+				i.batch = i.tree.ndb.db.NewBatch()
+				i.batchSize = 0
+			}
+		} else {
+			break
+		}
+	}
+	i.chanConfig.chNodeDataWg.Done()
```
odeke-em, line 129: Please make this a defer and move it to the top too.

import.go
```diff
@@ -45,57 +82,121 @@ func newImporter(tree *MutableTree, version int64) (*Importer, error) {
...
+}
+
+// batchWrite get a new batch from the channel and execute the batch write to the underlying DB.
+func batchWrite(i *Importer) {
+	for i.batch != nil {
+		if nextBatch, open := <-i.chanConfig.chBatch; open {
```
odeke-em, line 135: Please invert this condition and break early for !open instead of that else after many lines of code.

import.go
```diff
@@ -45,57 +82,121 @@ func newImporter(tree *MutableTree, version int64) (*Importer, error) {
...
+				break
+			}
+			i.batchMtx.Lock()
+			nextBatch.Close()
+			i.batchMtx.Unlock()
+		} else {
+			break
+		}
+	}
+	i.chanConfig.chBatchWg.Done()
```
odeke-em, line 148: Please use defer

import.go
```diff
@@ -45,57 +82,121 @@ func newImporter(tree *MutableTree, version int64) (*Importer, error) {
...
 // writeNode writes the node content to the storage.
 func (i *Importer) writeNode(node *Node) error {
+
+	if err := i.validate(node); err != nil {
+		return err
+	}
+
+	if nodeData, err := i.getDataBytes(node); err != nil {
+		return err
+	} else {
```
odeke-em, line 160: No need for this else after the prior return err, for idiomatic Go

import.go
```diff
@@ -45,57 +82,121 @@ func newImporter(tree *MutableTree, version int64) (*Importer, error) {
...
+	}
+
+	return nil
+}
+
+func (i *Importer) validate(node *Node) error {
 	if _, err := node._hash(node.nodeKey.version); err != nil {
 		return err
 	}
 	if err := node.validate(); err != nil {
```
odeke-em, line 175: Or you can simply invoke: return node.validate()
Then delete the following 4 lines

import.go
```diff
@@ -197,7 +319,7 @@ func (i *Importer) Commit() error {
 			len(i.stack))
 	}
 
-	err := i.batch.WriteSync()
```
odeke-em, line 200: You can transform this into: if err := ...; err != nil

import.go
```diff
@@ -211,3 +333,24 @@ func (i *Importer) Commit() error {
 	i.Close()
 	return nil
 }
+
+// waitAndCloseChannels will try to close all the channels for importer and wait for remaining work to be done.
+// This function should only be called in the Commit or Close action. If any error happens when draining the remaining data in the channel,
+// The error will be popped out and returned.
+func (i *Importer) waitAndCloseChannels() error {
+	// Make sure all pending works are drained and close the channels in order
+	if !i.chanConfig.allChannelClosed {
```
odeke-em, line 342: Please invert then return early

```go
if i.chanConfig.allChannelsClosed { // Already closed
   return nil
}

```

import.go
```diff
@@ -211,3 +333,24 @@ func (i *Importer) Commit() error {
...
+		i.chanConfig.allChannelClosed = true
+		close(i.chanConfig.chNodeData)
+		i.chanConfig.chNodeDataWg.Wait()
+		close(i.chanConfig.chBatch)
+		i.chanConfig.chBatchWg.Wait()
+		// Check errors
+		select {
+		case err := <-i.chanConfig.chError:
+			return err
+		default:
```
odeke-em, line 352: Did you intend to add this default case? I ask because if an error wasn't already in the channel under multiplexing, this select default clause immediately executes and doesn't wait for anything on the channel.

odeke-em (changes requested):
Thanks for this update @yzang2019, just some initial review comments.

yzang2019:
Seems like this patch is already cherry picked in https://github.com/cosmos/iavl/pull/793, going to close this original PR now

PR #676, opened 2023-02-02
feat: refactor the node key as `version + local nonce(seq id)`

cool-develope:


cool-develope:
benchmark results between `path` and `local nonce`

```
name                                                                 old time/op    new time/op    delta
Large/memdb-1000000-100-16-40/query-no-in-tree-guarantee-fast-6        3.49µs ± 3%    3.27µs ± 3%   -6.51%  (p=0.000 n=16+18)
Large/memdb-1000000-100-16-40/query-no-in-tree-guarantee-slow-6        17.4µs ±11%    12.9µs ± 7%  -26.14%  (p=0.000 n=17+18)
Large/memdb-1000000-100-16-40/query-hits-fast-6                        4.35µs ± 4%    4.03µs ± 9%   -7.45%  (p=0.000 n=18+20)
Large/memdb-1000000-100-16-40/query-hits-slow-6                        21.8µs ± 8%    16.1µs ± 7%  -26.00%  (p=0.000 n=18+19)
Large/memdb-1000000-100-16-40/iteration-fast-6                          591ms ± 5%     544ms ± 4%   -7.94%  (p=0.000 n=20+20)
Large/memdb-1000000-100-16-40/iteration-slow-6                          9.07s ± 2%     6.67s ± 2%  -26.43%  (p=0.000 n=16+19)
Large/memdb-1000000-100-16-40/update-6                                  218µs ±15%     174µs ± 6%  -19.85%  (p=0.000 n=19+19)
Large/memdb-1000000-100-16-40/block-6                                  23.9ms ± 4%    19.7ms ± 5%  -17.58%  (p=0.000 n=18+19)
Large/goleveldb-1000000-100-16-40/query-no-in-tree-guarantee-fast-6    9.49µs ± 5%    9.38µs ± 3%     ~     (p=0.086 n=19+19)
Large/goleveldb-1000000-100-16-40/query-no-in-tree-guarantee-slow-6    25.8µs ±35%    21.8µs ±23%  -15.53%  (p=0.000 n=18+19)
Large/goleveldb-1000000-100-16-40/query-hits-fast-6                    9.67µs ± 3%   10.05µs ± 4%   +3.94%  (p=0.000 n=20+20)
Large/goleveldb-1000000-100-16-40/query-hits-slow-6                    32.0µs ± 4%    27.3µs ±10%  -14.54%  (p=0.000 n=20+20)
Large/goleveldb-1000000-100-16-40/iteration-fast-6                      1.28s ± 2%     1.30s ± 6%     ~     (p=0.429 n=20+20)
Large/goleveldb-1000000-100-16-40/iteration-slow-6                      24.2s ± 2%     15.9s ± 4%  -34.35%  (p=0.000 n=18+19)
Large/goleveldb-1000000-100-16-40/update-6                              343µs ±21%     298µs ±17%  -13.16%  (p=0.000 n=19+19)
Large/goleveldb-1000000-100-16-40/block-6                              38.1ms ±11%    34.2ms ±17%  -10.21%  (p=0.000 n=19+20)

name                                                                 old alloc/op   new alloc/op   delta
Large/memdb-1000000-100-16-40/query-no-in-tree-guarantee-fast-6         88.0B ± 0%     88.0B ± 0%     ~     (all equal)
Large/memdb-1000000-100-16-40/query-no-in-tree-guarantee-slow-6        1.07kB ±10%    0.85kB ± 0%  -20.61%  (p=0.000 n=16+16)
Large/memdb-1000000-100-16-40/query-hits-fast-6                          210B ± 2%      210B ± 2%     ~     (p=0.978 n=16+16)
Large/memdb-1000000-100-16-40/query-hits-slow-6                        1.31kB ± 0%    1.08kB ± 0%  -17.75%  (p=0.000 n=17+20)
Large/memdb-1000000-100-16-40/iteration-fast-6                          232MB ± 0%     232MB ± 0%   -0.00%  (p=0.037 n=20+17)
Large/memdb-1000000-100-16-40/iteration-slow-6                         1.22GB ± 0%    1.10GB ± 0%  -10.50%  (p=0.000 n=18+18)
Large/memdb-1000000-100-16-40/update-6                                 26.4kB ± 1%    24.1kB ± 1%   -8.71%  (p=0.000 n=19+18)
Large/memdb-1000000-100-16-40/block-6                                  2.78MB ± 0%    2.55MB ± 1%   -8.37%  (p=0.000 n=19+20)
Large/goleveldb-1000000-100-16-40/query-no-in-tree-guarantee-fast-6      993B ± 0%      996B ± 1%   +0.27%  (p=0.000 n=19+20)
Large/goleveldb-1000000-100-16-40/query-no-in-tree-guarantee-slow-6    2.63kB ±61%    2.23kB ±14%  -15.15%  (p=0.000 n=18+17)
Large/goleveldb-1000000-100-16-40/query-hits-fast-6                    1.03kB ± 1%    1.03kB ± 1%     ~     (p=0.770 n=18+18)
Large/goleveldb-1000000-100-16-40/query-hits-slow-6                    3.12kB ± 1%    3.00kB ± 1%   -3.97%  (p=0.000 n=19+20)
Large/goleveldb-1000000-100-16-40/iteration-fast-6                      298MB ± 0%     298MB ± 0%   +0.07%  (p=0.000 n=19+18)
Large/goleveldb-1000000-100-16-40/iteration-slow-6                     3.44GB ± 0%    2.53GB ± 0%  -26.46%  (p=0.000 n=18+18)
Large/goleveldb-1000000-100-16-40/update-6                             51.9kB ±10%    47.4kB ±10%   -8.63%  (p=0.000 n=19+19)
Large/goleveldb-1000000-100-16-40/block-6                              6.42MB ± 7%    5.50MB ±12%  -14.33%  (p=0.000 n=20+20)

name                                                                 old allocs/op  new allocs/op  delta
Large/memdb-1000000-100-16-40/query-no-in-tree-guarantee-fast-6          3.00 ± 0%      3.00 ± 0%     ~     (all equal)
Large/memdb-1000000-100-16-40/query-no-in-tree-guarantee-slow-6          58.4 ± 6%      36.0 ± 0%  -38.33%  (p=0.000 n=16+16)
Large/memdb-1000000-100-16-40/query-hits-fast-6                          4.00 ± 0%      4.00 ± 0%     ~     (all equal)
Large/memdb-1000000-100-16-40/query-hits-slow-6                          65.5 ± 1%      43.0 ± 0%  -34.35%  (p=0.000 n=20+20)
Large/memdb-1000000-100-16-40/iteration-fast-6                          4.00M ± 0%     4.00M ± 0%     ~     (p=0.307 n=19+20)
Large/memdb-1000000-100-16-40/iteration-slow-6                          39.1M ± 0%     33.0M ± 0%  -15.53%  (p=0.000 n=18+18)
Large/memdb-1000000-100-16-40/update-6                                    711 ± 1%       554 ± 1%  -22.05%  (p=0.000 n=19+15)
Large/memdb-1000000-100-16-40/block-6                                   73.6k ± 1%     57.4k ± 1%  -22.03%  (p=0.000 n=20+20)
Large/goleveldb-1000000-100-16-40/query-no-in-tree-guarantee-fast-6      19.0 ± 0%      19.0 ± 0%     ~     (all equal)
Large/goleveldb-1000000-100-16-40/query-no-in-tree-guarantee-slow-6      81.7 ±37%      51.9 ±12%  -36.47%  (p=0.000 n=18+17)
Large/goleveldb-1000000-100-16-40/query-hits-fast-6                      19.0 ± 0%      19.0 ± 0%     ~     (all equal)
Large/goleveldb-1000000-100-16-40/query-hits-slow-6                      90.5 ± 1%      64.0 ± 0%  -29.28%  (p=0.000 n=20+16)
Large/goleveldb-1000000-100-16-40/iteration-fast-6                      5.31M ± 0%     5.31M ± 0%   +0.00%  (p=0.000 n=19+19)
Large/goleveldb-1000000-100-16-40/iteration-slow-6                      70.5M ± 0%     57.2M ± 0%  -18.89%  (p=0.000 n=18+18)
Large/goleveldb-1000000-100-16-40/update-6                                939 ± 4%       775 ± 5%  -17.52%  (p=0.000 n=18+19)
Large/goleveldb-1000000-100-16-40/block-6                                100k ± 4%       83k ± 8%  -16.54%  (p=0.000 n=19+20)
```

yihuang:
👍 I like it, the only issue is how to efficiently assign nonce by version during import.
I guess int16 array with a map to handle overflows would be more memory efficient.

ValarDragon:
Cool that `local_version + nonce` is faster on all metrics!

cool-develope:
> 👍 I like it, the only issue is how to efficiently assign nonce by version during import. I guess int16 array with another map to handle overflows would be more memory efficient. Another idea to reduce memory usage is, we don't have to track nonce for each version strictly, we can do that for a version range, that'll greatly reduce the array size.

the idea of version range is interesting, is there a way to know about the approx count of leaves in one block?

immutable_tree.go
```diff
@@ -324,12 +323,7 @@ func (t *ImmutableTree) clone() *ImmutableTree {
 //
 
 func (t *ImmutableTree) nodeSize() int {
```
tac0turtle, line 325: nit: out of scope of this pr but returning 64 or 32 should be set here.

tac0turtle:
can you update the docs associated with this change, also the orphan docs need updating

immutable_tree.go
cool-develope: it is only used in the test, if specify 32 or 64, we would have an inconvenience.

keyformat/key_format.go
```diff
@@ -105,6 +106,13 @@ func (kf *KeyFormat) Key(args ...interface{}) []byte {
 	return kf.KeyBytes(segments...)
 }
 
+func (kf *KeyFormat) NodeKey(nodeKey int64) []byte {
```
kocubinski, line 109: I can't find usages of this function anywhere, am I missing something?

cool-develope: good catch

catShaark:
This maybe out of scope for this PR but I think we should change the key format for leaf nodes to be `version | l | key` where key is the value's key. Doing so, we can fetch key/value fair from iavl.Store and iterate without tree traversal.

node.go
```diff
@@ -16,40 +17,51 @@ import (
 	"github.com/cosmos/iavl/internal/encoding"
 )
 
+// NodeKey represents a key of node in the DB.
+type NodeKey struct {
+	version int64
```
catShaark, line 22: shouldn't version be int32 and nonce be int64 since there're more nonce than version

ValarDragon:
Shouldn't hardware prefetching mostly eliminate that as a time overhead?

yihuang:
> This maybe out of scope for this PR but I think we should change the key format for leaf nodes to be `version | l | key` where key is the value's key. Doing so, we can fetch key/value fair from iavl.Store and iterate iavl.Store without tree traversal.

This may not serve what you want because you don't know what exact version the key is modified, unless also maintain some other indexes.

yihuang:
> > This maybe out of scope for this PR but I think we should change the key format for leaf nodes to be version | l | key where key is the value's key. Doing so, we can fetch key/value fair from iavl.Store and iterate iavl.Store without tree traversal.
> 
> 
> 
> Shouldn't hardware prefetching mostly eliminate the tree traversal as a time overhead? 

Tree nodes are scattered at different versions, prefetching don't help much I think.

yihuang:
> > > This maybe out of scope for this PR but I think we should change the key format for leaf nodes to be `version | l | key` where key is the value's key. Doing so, we can fetch key/value fair from iavl.Store and iterate iavl.Store without tree traversal.
> 
> > 
> 
> > Sounds like a very clever idea, but this may not serve what you want because you don't know what exact version the key is modified, unless also maintain some other indexes.
> 
> 
> 
> But we always know the version since when we load iavl.Store, we always specify the version

That is the version you want to query in, not the version the key is last modified, the version contained in the node is the version the node is created in, aka. when the key is modified or inserted.

cool-develope:
> > > This maybe out of scope for this PR but I think we should change the key format for leaf nodes to be `version | l | key` where key is the value's key. Doing so, we can fetch key/value fair from iavl.Store and iterate iavl.Store without tree traversal.
> > 
> > 
> > Sounds like a very clever idea, but this may not serve what you want because you don't know what exact version the key is modified, unless also maintain some other indexes.
> 
> But we always know the version since when we load iavl.Store, we always specify the version

@yihuang is right, we have different versions, how is key-iteration query possible? since the version prefix is different

alexanderbez (approved):

Makefile
```diff
@@ -53,18 +53,18 @@ lint-fix:
...
 .PHONY: bench
 
 # fullbench is extra tests needing lots of memory and to run locally
 fullbench:
 	cd benchmarks && \
 		go test $(LDFLAGS) -run=NOTEST -bench=RandomBytes . && \
-		go test $(LDFLAGS) -tags cleveldb,rocksdb,pebbledb -run=NOTEST -bench=Small . && \
-		go test $(LDFLAGS) -tags cleveldb,rocksdb,pebbledb -run=NOTEST -bench=Medium . && \
-		go test $(LDFLAGS) -tags cleveldb,rocksdb,pebbledb -run=NOTEST -timeout=30m -bench=Large . && \
+		go test $(LDFLAGS) -tags pebbledb -run=NOTEST -bench=Small . && \
```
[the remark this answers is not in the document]
cool-develope: The test failed on RocksDB previously for large data, I will test again

node.go
```diff
@@ -16,40 +17,51 @@ import (
...
+	nonce   int32
+}
```
kocubinski, line 24: I'm not really seeing the benefit of `NodeKey` struct and abstraction, it seems to have created a big diff for no benefit that I can see.  Did you consider a type alias like `type NodeKey []byte`?   I think this could remove some code.

node.go
```diff
@@ -16,40 +17,51 @@ import (
...
+func (nk *NodeKey) GetKey() []byte {
+	b := make([]byte, 12)
+	binary.BigEndian.PutUint64(b, uint64(nk.version))
+	binary.BigEndian.PutUint32(b[8:], uint32(nk.nonce))
+	return b
+}
+
 // Node represents a node in a Tree.
 type Node struct {
 	key           []byte
```
kocubinski, line 35: If `nodeKey NodeKey` is now representing this node's key, what is the purpose of the `key` field?

kocubinski (commented):
Logically things look OK to me.  Maybe it's a big nit, but I wonder if this diff and implementation would be cleaner and more readable without the `NodeKey`  struct and the key just represented as []byte (same as it is now as hash).

To that I end I did a little refactoring PoC to see what it looks like.  Let me know what you think, since you've spent time on this problem here than me.

https://github.com/cosmos/iavl/compare/592/refactor-nonce-new...kocubinski/592?expand=1

node.go
```diff
@@ -16,40 +17,51 @@ import (
 	"github.com/cosmos/iavl/internal/encoding"
 )
 
+// NodeKey represents a key of node in the DB.
+type NodeKey struct {
```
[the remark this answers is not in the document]
cool-develope: I am not sure, we have been using the `node key` term so far, and actually `NodeKey` struct is used as an identifier, it is a little different with `dbKey`, right?

node.go
```diff
@@ -16,40 +17,51 @@ import (
...
 	}
 }
 
 // MakeNode constructs an *Node from an encoded byte slice.
 //
 // The new node doesn't have its hash saved or set. The caller must set it
 // afterwards.
-func MakeNode(buf []byte) (*Node, error) {
-	// Read node header (height, size, version, key).
+func MakeNode(nodeKey *NodeKey, buf []byte) (*Node, error) {
```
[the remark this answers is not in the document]
cool-develope: I think this API is not exposed

node.go
cool-develope: it will increase the node size in the storage (writeBytes).
the real db key of the node is 12 bytes, but the leftNodeKey can be 2 bytes.
`BigEndian` vs `binaary.PutUvarint`

cool-develope: I agree with you, it will reduce some encodeing/decoding operations, but the storage size is more significant IMO

proof.go
```diff
@@ -166,23 +166,27 @@ func (pln ProofLeafNode) Hash() ([]byte, error) {
...
-func (node *Node) pathToLeaf(t *ImmutableTree, key []byte, path *PathToLeaf) (*Node, error) {
+func (node *Node) pathToLeaf(t *ImmutableTree, key []byte, version int64, path *PathToLeaf) (*Node, error) {
 	if node.subtreeHeight == 0 {
 		if bytes.Equal(node.key, key) {
 			return node, nil
 		}
 		return node, errors.New("key does not exist")
 	}
 
+	nodeVersion := version
```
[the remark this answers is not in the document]
cool-develope: it is related to new design, we have new added nodes while uncommitting stage, and these nodes will have no nodekey and versioni, in that case we should indicate the current version for those nodes.

node.go
```diff
@@ -440,20 +492,32 @@ func (node *Node) writeBytes(w io.Writer) error {
...
+		}
+		if node.leftNodeKey == nil {
+			return ErrLeftNodeKeyEmpty
+		}
+		cause = encoding.EncodeVarint(w, node.leftNodeKey.version)
+		if cause != nil {
+			return fmt.Errorf("writing the version of left node key, %w", cause)
 		}
-		cause = encoding.EncodeBytes(w, node.leftHash)
+		cause = encoding.EncodeVarint(w, int64(node.leftNodeKey.nonce))
```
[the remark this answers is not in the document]
cool-develope: it only cares about how much is the number, not about the integer variable type in `EncodeVarint`.
the result is same for 2(int64) and 2(int32), we will reduce the memory usage with int32 nonce

cool-develope:
Here is the rocksDB benchmarking result, there is not much improvement in RocksDB

`iavl: 0.20.0-alpha1-23-gde1532f
goos: linux
goarch: amd64
pkg: github.com/cosmos/iavl/benchmarks
cpu: DO-Premium-Intel
                                                                     │   old.txt          │   new.txt    │    
                                                                     │    sec/op          │    sec/op    │ 
Larger/memdb-1000000-100-16-40/query-no-in-tree-guarantee-fast-4       3.818µ ±  3%         4.010µ ±  3% 
Larger/memdb-1000000-100-16-40/query-no-in-tree-guarantee-slow-4       20.66µ ±  6%         18.34µ ±  5% 
Larger/memdb-1000000-100-16-40/query-hits-fast-4                       5.650µ ±  7%         5.913µ ±  4% 
Larger/memdb-1000000-100-16-40/query-hits-slow-4                       25.17µ ±  8%         22.22µ ±  2% 
Larger/memdb-1000000-100-16-40/iteration-fast-4                        592.3m ±  3%         553.8m ±  5% 
Larger/memdb-1000000-100-16-40/iteration-slow-4                         12.78 ±  4%          6.023 ±  2%
Larger/memdb-1000000-100-16-40/update-4                                317.9µ ±  5%         212.5µ ± 10%
Larger/memdb-1000000-100-16-40/block-4                                 34.30m ±  6%         24.11m ±  3%
Larger/goleveldb-1000000-100-16-40/query-no-in-tree-guarantee-fast-4   4.123µ ±  7%         4.169µ ±  2%
Larger/goleveldb-1000000-100-16-40/query-no-in-tree-guarantee-slow-4   32.94µ ± 12%         26.92µ ±  6%
Larger/goleveldb-1000000-100-16-40/query-hits-fast-4                   12.36µ ±  3%         12.22µ ±  4%
Larger/goleveldb-1000000-100-16-40/query-hits-slow-4                   41.00µ ±  5%         33.70µ ±  4%
Larger/goleveldb-1000000-100-16-40/iteration-fast-4                    780.4m ±  3%         819.9m ± 10%
Larger/goleveldb-1000000-100-16-40/iteration-slow-4                     30.43 ±  3%          13.92 ±  2%
Larger/goleveldb-1000000-100-16-40/update-4                            373.6µ ±  9%         276.1µ ± 10%
Larger/goleveldb-1000000-100-16-40/block-4                             50.60m ±  3%         35.58m ±  5%
Larger/rocksdb-1000000-100-16-40/query-no-in-tree-guarantee-fast-4     6.655µ ±  6%         6.695µ ±  8%
Larger/rocksdb-1000000-100-16-40/query-no-in-tree-guarantee-slow-4     22.06µ ±  9%         25.07µ ±  7%
Larger/rocksdb-1000000-100-16-40/query-hits-fast-4                     8.097µ ±  4%         8.287µ ±  3%
Larger/rocksdb-1000000-100-16-40/query-hits-slow-4                     29.97µ ±  4%         28.99µ ±  4%
Larger/rocksdb-1000000-100-16-40/iteration-fast-4                       4.352 ±  2%          4.359 ±  3%
Larger/rocksdb-1000000-100-16-40/iteration-slow-4                       16.81 ±  2%          15.28 ±  2%
Larger/rocksdb-1000000-100-16-40/update-4                              436.1µ ± 10%         281.0µ ± 10%
Larger/rocksdb-1000000-100-16-40/block-4                               45.70m ±  6%         32.38m ±  4%
geomean                                                                1.356m               1.135m
`

node.go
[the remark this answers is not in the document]
kocubinski: It is exported, see: https://go.dev/tour/basics/3

node.go
kocubinski: I'm not following how the use of a struct to organize code relates to encoding?  All the same information is present.

node.go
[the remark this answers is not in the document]
cool-develope: absolutely, it is exported. I believe there is no usecase of this api outside.

node.go
cool-develope: because the encoding style is different in the key and value of database, we are using `binary.PutUvarint` in value writeBytes (to reduct the storage size) and BigEndian in the nodekey (to keep sorted).
If we go with []byte in leftNodeKey, we can't use the `node.leftNodeKey` directly as the db key of left node, it will require extra encoding/decoding.

kocubinski: That seems like something we could code around, but it's also just a style choice in the first place, so I'll resolve this conversation. :+1:

kocubinski (approved):

ValarDragon:
:tada:

Issue #682, opened 2023-02-17
Remove Tendermint dependency from v0.19.x


julienrbrt:
For migrating Cosmos SDK v0.47.x to CometBFT, we need to remove `tendermint/tendermint` dependency from the SDK dependency graph. ref https://github.com/cosmos/cosmos-sdk/pull/15060#issuecomment-1434886385.

We should keep tm-db (and not migrate at the same time to cosmos-db / cometbft-db), as we need it for v0.45 and v0.46 (and v0.47).

This has been done on main https://github.com/cosmos/iavl/commit/3da70633364adb442a57c2ef12f2559a20a01c09, would it be possible to do the same for the v0.19.x line?

cc @tac0turtle @cool-develope

julienrbrt:
Closing this then, as discussed (on Slack), I have a PR reverting the migration to Cosmos-DB.

PR #683, opened 2023-02-18
refactor: revert IAVL v0.20.x to CometBFT-DB

julienrbrt:
As mentioned in https://github.com/cosmos/iavl/issues/682, we unfortunately cannot remove the tendermint deps in the 0.19.x line.

This means, we need to use the 0.20.x line for v0.47. Because cosmos-db has not been audited, we are reverting to cometbft-db (ex tm-db) in the v0.20.x line.


tac0turtle (approved):

tac0turtle:
we should branch off of release/v0.20.x before this commit so main in the sdk can still work with cosmos-db

Issue #688, opened 2023-02-22
export: don't export branch node's key in snapshot


yihuang:
in IAVL, a branch node's key is always the smallest leaf key in it's right branch, the key can be infered during the import process, so no need to include it in the snapshot at all.

## Migration process

- Step 1, change the import logic to derive the branch node's key on the fly, ignore the one included in the snapshot, this is not a consensus breaking change.
- Step 2, change the export logic to skip branch node's key, this should be considered as a breaking change and need a coordinated upgrade, so the snapshots created after the upgrade are updated consistently across the network.

Do we need to update the snapshot format version for this?

### Positives

- Reduce snapshot size, say `L` is the number of leave nodes, previously we export `2 * L - 1` keys and `L` values, now we export `L` keys and `L` values, if key and value are the same length on average, it reduces 30% size of snapshot.
- In the future, can apply delta encoding to the leaf keys to further reduce snapshot size, since the leaf keys are exported in order, apply delta encoding to it should produce very good compression rate.

yihuang:
fixed by #703

PR #689, opened 2023-02-22
skip branch node's key in import/export

yihuang:
Closes: #688

since it can be derived on the fly.

I think should be ok to backport to 0.19, since it's not breaking when the new option is `false`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - [#646](https://github.com/cosmos/iavl/pull/646) Remove the `orphans` from the storage
 - [#622](https://github.com/cosmos/iavl/pull/622) `export/newExporter()` and `ImmutableTree.Export()` returns error for nil arguements
+- [#]() Skip branch node's key in import/export, the export behavior change is guarded by new option `DisableExportBranchNodeKey` which defaults to `false`, it's only breaking if user switch that option on.
 
 ### API Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - [#646](https://github.com/cosmos/iavl/pull/646) Remove the `orphans` from the storage
 - [#622](https://github.com/cosmos/iavl/pull/622) `export/newExporter()` and `ImmutableTree.Export()` returns error for nil arguements
-- [#]() Skip branch node's key in import/export, the export behavior change is guarded by new option `DisableExportBranchNodeKey` which defaults to `false`, it's only breaking if user switch that option on.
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - [#646](https://github.com/cosmos/iavl/pull/646) Remove the `orphans` from the storage
 - [#622](https://github.com/cosmos/iavl/pull/622) `export/newExporter()` and `ImmutableTree.Export()` returns error for nil arguements
+- [#]() Skip branch node's key in import/export, the export behavior change is guarded by new option `DisableExportBranchNodeKey` which defaults to `false`, it's only breaking if user switch that option on.
```
yihuang, line 7:
```suggestion
- [#689](https://github.com/cosmos/iavl/pull/689) Skip branch node's key in import/export, the export behavior change is guarded by new option `DisableExportBranchNodeKey` which defaults to `false`, it's only breaking if user switch that option on.
```

import.go
```diff
@@ -27,6 +27,9 @@ type Importer struct {
 	batch     db.Batch
 	batchSize uint32
 	stack     []*Node
+	// minKeyStack has the same length as stack, it maintains the smallest key in each node's subtree,
+	// branch node always use the smallest key in it's right branch, so we don't have to export the branch node's key.
```
yihuang, line 31:
```suggestion
	// we use it to derive the smallest key of the right branch for branch nodes, so we don't have to export the branch node's key in snapshot.
```

options.go
```diff
@@ -81,6 +81,10 @@ type Options struct {
 
 	// When Stat is not nil, statistical logic needs to be executed
 	Stat *Statistics
+
+	// don't export branch node's key in snapshot, default to false.
+	// see: https://github.com/cosmos/iavl/issues/688
+	DisableExportBranchNodeKey bool
```
tac0turtle, line 87: is there any reason to expose this to an application

yihuang: I guess it's possible to control the migration process without doing a coordinated upgrade, for example release the binary first but disable the breaking behavior, after most of people updated binary, then enable this.

alexanderbez (approved):

yihuang:
https://github.com/cosmos/iavl/issues/690#issuecomment-1441167234

Just had some thoughts about format migration, I think better to introduce a new format number for this change, and have some support in sdk side to smooth the migration process.

tac0turtle:
agree, bumping the version number works and sdk can deny incompatible versions. Only thing is what version a node gets is non deterministic cause comet doesnt have topics. it could be okay cause the ux around state sync is already so poor that many users dont use it

yihuang:
closed in favor of https://github.com/cosmos/iavl/pull/703

Issue #690, opened 2023-02-22
export: apply delta encoding for better compression rate


yihuang:
Snapshot is always created and processed in order, taking advantage of that, we can apply delta encodings to some fields:
- version
  IAVL tree has the nice property that parent node's version is always bigger than or equal to children's version, and in snapshot, nodes are added in post-order, where parent node is put after children nodes, so we can only record the delta value, to be specific `parent version - max(left version, right version)`, smaller value results in smaller encoded size with varint encoding.
- key
  After the #688 improvement, we only have leaf keys in snapshot, they are ordered, so delta encoding can also potentially have some good compression rate there.
- height
  since IAVL tree has the invariant that `parent height = max(left height, right height) + 1`, I think we don't need to encode the height field at all in the snapshot.

The version and height compression rely on the fact that nodes are exported in post-order. the key compression only need the leaf nodes are exported in order.

yihuang:
Some thoughts about snapshot format migration:
- the IAVL library always support all the legacy versions, add a `format` parameter to `import/export` methods, so it's always a non-breaking change to introduce a new format in IAVL library.
- add a new api to return all the supported formats in current version.
- let cosmos-sdk to handle the versioning of the snapshot, it can check the snapshot format against supporting list, and can have a new config option to specify which format to export.

PR #703, opened 2023-03-13
feat: more compact exported snapshot

yihuang:
Closes: #690 #688

- Purely additive changes for iavl library, add new APIs to support a more compact snapshot format.
- Still use the existing `ExportNode` type as the exchange format with sdk, sdk can still use the protobuf encoding as before, which skips empty fields, which works well with the new optimizations.
- Leave the issue of format negotiation and migration to sdk.

### Where to put the code

With the approach taken in this PR, it's totally fine to move the logic to the sdk side as well, if it makes more sense.

### Test Result

Dumped from cronos mainnet data, evm module, version 2000000:

| Snapshot Type             | Size   |
| ------------------------- | ------ |
| Normal snapshot  | 469.4M |
| Compact snapshot | 402M   |
| memiavl snapshot          | 2.3G   |

* The new format further reduce the zlib compressed size by `14%`, not as big as expected, I guess mainly because zlib compression already done a good job.
* Both normal/compact snapshots are dumped with sdk's snapshot manager, with and without the compact format introduced in this PR, sum the size of all the chunks.
* memiavl snapshot is just put for reference, it's a different snapshot format that contains all the node hashes, with zero compression applied.

EDIT:

Just to confirm my suspicion, I recreate the snapshot with zlib compression disabled, it goes like this:

| Snapshot Type    | Size  |
| ---------------- | ----- |
| Normal snapshot  | 1.82G |
| Compact snapshot | 856M  |

The compression rate is `54%`.


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+### Improvements
+
+- [#]() New APIs `NewCompressExporter`/`NewCompressImporter` to support more compact snapshot format.
```
yihuang, line 5:
```suggestion
- [#703](https://github.com/cosmos/iavl/pull/703) New APIs `NewCompressExporter`/`NewCompressImporter` to support more compact snapshot format.
```

tac0turtle (commented):
lets leave this here for now, in store v2 this could be moved higher, into the sdk

compress.go
```diff
@@ -0,0 +1,144 @@
...
+	} else {
+		l = len(b)
+	}
+	for ; off < l; off++ {
+		if a[off] != b[off] {
+			break
+		}
+	}
+	return off
+}
```
cool-develope, line 137:
```
func diffOffset(a, b []byte) int {
	var off int
	for ; off < len(a) && l < len(b)l; off++ {
		if a[off] != b[off] {
			break
		}
	}
	return off
}
```

compress.go
```diff
@@ -0,0 +1,144 @@
...
+		// apply delta encoding to leaf keys
+		n.Key, e.lastKey = deltaEncode(n.Key, e.lastKey), n.Key
+
+		e.versionStack = append(e.versionStack, n.Version)
+	} else {
+		// branch keys can be derived on the fly when import, safe to skip
+		n.Key = nil
+
+		// delta encode the version
+		maxVersion := maxInt64(e.versionStack[len(e.versionStack)-1], e.versionStack[len(e.versionStack)-2])
```
cool-develope, line 48: just curious about the efficiency of version delta encode?

compress.go
cool-develope, line 137:
```suggestion
func diffOffset(a, b []byte) int {
	var off int
	for ; off < len(a) && off < len(b); off++ {
		if a[off] != b[off] {
			break
		}
	}
	return off
}
```

yihuang:
> just curious about the efficiency of version delta encode?

Just did some more detailed comparisons:

## With Zlib Compression(level 7)

|                                             | Size (bytes) | Compression Ratio |
| ------------------------------------------- | ------------ | ----------------- |
| Original                                    | 492170475    | -                 |
| Skip branch key                             | 455839686    | 7%                |
| Skip branch key + key delta                 | 444661628    | 10%               |
| Skip branch key + key delta + version delta | 421542633    | 14%               |

## Without Zlib Compression

|                                             | Size (bytes) | Compression Ratio |
| ------------------------------------------- | ------------ | ----------------- |
| Original                                    | 1955535429   | -                 |
| Skip branch key                             | 1308448614   | 33%               |
| Skip branch key + key delta                 | 936722517    | 52%               |
| Skip branch key + key delta + version delta | 897597115    | 54%               |

compress.go
yihuang: This change will do two comparisons instead of one in the for loop

compress.go
yihuang: https://github.com/cosmos/iavl/pull/703#issuecomment-1472958077

cool-develope:
txs for detail report

cool-develope (approved):

tac0turtle:

PR #717, opened 2023-03-23
fix: import forward version

cool-develope:
Since we have no root store, we should create the link node for the forward version when importing


tac0turtle:
> Since we have no root store

could you touch on what you mean by this?

cool-develope:
> could you touch on what you mean by this?

Originally, we store a root hash with the version as the key.  For now, we assume the root node is always with `version` + 1 (nonce) as the key. When we have no updates, we are trying to store the previous version root node link. This is figured out in tree operations but not in import, so it is

import_test.go
```diff
@@ -206,6 +206,22 @@ func TestImporter_Commit(t *testing.T) {
 	require.True(t, has)
 }
 
+func TestImporter_Commit_ForwardVersion(t *testing.T) {
+	tree, err := NewMutableTree(db.NewMemDB(), 0, false)
+	require.NoError(t, err)
+	importer, err := tree.Import(2)
```
kocubinski, line 212: I don't really understand how we'd arrive here during a production import (import version ahead of a node in the stack). can you explain a bit more or provide an example?

cool-develope: for example, if there are no tree operations (like set, delete) for the recent 2 versions, we have a root of version `n`, and need to import version `n+2`. In that case, we are only storing the root of version `n` at import previously and not storing `n+2`th root which is the previous version root link, so when commit the tree, it says there is no version `n+2` in the tree.

tac0turtle: did this change from the previous code? or was it handled by orphans?

import.go
```diff
@@ -177,6 +177,11 @@ func (i *Importer) Commit() error {
 		if err := i.writeNode(i.stack[0]); err != nil {
 			return err
 		}
+		if i.stack[0].nodeKey.version < i.version { // it means there is no update in the given version
+			if err := i.batch.Set(i.tree.ndb.nodeKey(&NodeKey{version: i.version, nonce: 1}), i.tree.ndb.nodeKey(i.stack[0].nodeKey)); err != nil {
```
yihuang, line 181: it seems `GetNode` don't handle this redirection node format, or is it handled in another place?

batch.go
```diff
@@ -16,7 +16,7 @@ type BatchWithFlusher struct {
 var _ dbm.Batch = &BatchWithFlusher{}
 
 // Ethereum has found that commit of 100KB is optimal, ref ethereum/go-ethereum#15115
-var defaultFlushThreshold = 100000
+// var defaultFlushThreshold = 100000
```
yihuang, line 19: maybe delete the dead code?

yihuang (approved):
LGTM

batch.go
cool-develope: it came from another PR, seems like using it as a reference description. just comment to avoid lint failures

PR #718, opened 2023-03-23
fix: `traverseNodes` unexpected behaviour

catShaark:
- If there's no update in the latest version of the tree, we [set the root node's value to be the previous root node's key](https://github.com/cosmos/iavl/blob/842d58b63d0cfd82c67987b4bcaac3a9edffecc3/nodedb.go#L545) instead of its marshalled bytes. When we call `traverseNodes`, we iterate through every nodes (including root nodes) and unmarshal the marshalled node's value via `MakeNode`. `MakeNode` is made only for unmarshaling a marshalled node's value, so if a marshalled node's value is actually a node's key like described above, it'll handle incorrectly.
- Suggest that in `traverseNodes` we check if the marshalled node's value is indeed a root node's key so that we can ignore it


cool-develope:
@catShaark , thanks for catching it, it seems like we need to refactor the `traverseNode` function using `prefixIterate`
FYI, it is a test utility function (not critical part).

catShaark:
should I close this now

cool-develope:
> should I close this now

or you can modify it

catShaark:
ohh, I'll modify this

cool-develope (approved):

tac0turtle:
@catShaark can we get a changelog here then we can merge

catShaark:
Yess

catShaark:
@tac0turtle, changelog added

tac0turtle (approved):

PR #724, opened 2023-04-03
fix: consider the referenced nodes in `traverseNodes`

cool-develope:
Closes: #718


cool-develope:
@catShaark
I cleaned the code based on your PR, thanks for contributing again

tac0turtle:
which pr do we keep open? this or catsharks?

cool-develope:
this PR include some extra cleaning ups, let's move with @catShaark firstly

nodedb.go
```diff
@@ -511,6 +511,15 @@ func (ndb *nodeDB) HasVersion(version int64) (bool, error) {
 	return ndb.db.Has(nodeKeyFormat.Key(version, []byte{1}))
 }
 
+func isReferenceToRoot(bz []byte) bool {
+	if bz[0] == nodeKeyFormat.Prefix()[0] {
+		if len(bz) == 13 {
```
yihuang, line 516: It seems the condition for a referenced node is hacky, maybe we should add a prefix to normal node values as well, so we can distinguish with prefix alone.

PR #726, opened 2023-04-08
feat: make ChangeSet and KVPair protobuf serializable

yihuang:
Implement protobuf messages, and setup buf.build

```
$ buf generate proto
```

## Context

we are reusing the `ChangeSet` and `KVPair` type from the iavl library, and we need it to be serializable, so instead of rolling our own types for it, it's be better if the iavl type support protobuf serialization directly.


tac0turtle (commented):
dam i guess we need the proto type, @cool-develope does this help you as well?

tac0turtle (approved):

tac0turtle:
@mergifyio backport release/v0.20.x

tac0turtle:

PR #727, opened 2023-04-09
perf: Make debugging a constant, to allow deadcode elimination

ValarDragon:
Changes this private var to a const, so that the relevant code can get deadcode eliminated by the compiler.

I have no benchmark proving efficiency win, but seems like a clear / costless improvement.


internal/logger/logger.go
```diff
@@ -4,7 +4,7 @@ import (
 	"fmt"
 )
 
-var debugging = false
+const debugging = false
```
tac0turtle, line 7: if its a const then a user cant modify this via flags.

ValarDragon: The user already can't modify it via flags? Its a private variable.

As built right now, only a developer can change this by editing the variable.

tac0turtle (approved):
Opened another pr removing this. We can merge this for now

PR #728, opened 2023-04-09
perf: Remove SaveNode cache update

ValarDragon:
Remove NodeDB.SaveNode from updating the LRU cache. (FastNodes still update the LRU cache on write)

SaveNode touching the cache adds many extra nodes into this, that are likely unused given the fast nodes. They would only come in handy for IBC proof queries. 

But caching every write is an ineffective strategy for that, instead you would rather employ strategies that "watch" for applicable key ranges.

SaveNodes adds some overhead for InitGenesis logic with the extra heap allocations, which this then speeds up.

### DIFF
--- a/nodedb.go
+++ b/nodedb.go
@@ -199,7 +199,6 @@ func (ndb *nodeDB) SaveNode(node *Node) error {
 	}
 
 	logger.Debug("BATCH SAVE %+v\n", node)
-	ndb.nodeCache.Add(node)
 	return nil
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Improvements
 
 - [#703](https://github.com/cosmos/iavl/pull/703) New APIs `NewCompressExporter`/`NewCompressImporter` to support more compact snapshot format.
+- [#728](https://github.com/cosmos/iavl/pull/728) Remove automatically adding internal IAVL nodes into the LRU cache.
 
 ### Breaking Changes
 


nodedb.go
```diff
@@ -199,7 +199,6 @@ func (ndb *nodeDB) SaveNode(node *Node) error {
 	}
 
 	logger.Debug("BATCH SAVE %+v\n", node)
-	ndb.nodeCache.Add(node)
```
tac0turtle, line 202: can we get a changelog for this

tac0turtle (commented):
does this effect performance? could we run the benchmarks

ValarDragon:
This does affect performance, for genesis writes: 
image

Removes that 30 seconds of time spent in replacing the latest LRU objects (that was mostly wasted work)

But more importantly, I claim that this improves the cache hit rate for realistic query workloads, independent of genesis work.

ValarDragon:
This could have some disadvantage for repeated writes to the same state, in a subsequent block. (As it would have to traverse the same nodes)

Its unclear to me that this is going to be a problem, but not super obvious to make that claim though :/ 

So I understand if we close this, and just do something to bound writing save node to the LRU if theres too many writes within a single block.

nodedb.go
ValarDragon: added!

yihuang:
One workaround for genesis writing is maybe support setting empty cache size temporarily?

tac0turtle:
closing this with iavlv2 being around the corner

PR #730, opened 2023-04-09
perf: Add resetBatch optimization to fastnode genesis

ValarDragon:
Add reset batch optimization to writing fast nodes at genesis, mirroring whats done for normal node saving.

This is really the same idea as #652 , but at the genesis-only level right now, as thats whats already done for normal nodes, and deemed normal at genesis. (Whereas #652 may have to reason about different guarantees, e.g. signalling when the full batch is completed)

This brings a significant speedup for InitGenesis logic.

Before (370 s + GC):
image

After (41.3 s):
image

This is also with massive GC reductions (due to far less leveldb.grow operations), and lower peak RAM usage. (One less full-module write copy)


tac0turtle (approved):

cool-develope (approved):

ValarDragon:
Marked this and #729 as backport v0.20.x, if thats alright. Would like to get Osmosis updated to have these, to reduce our InitGenesis overheads

PR #735, opened 2023-04-10
refactor: add logger

tac0turtle:
replace local logger with sdk one, this help sin the future when we add logs to the repo. We should avoid adding info logs here and mainly to error and debug if possible


julienrbrt (approved):
lgtm! We probably need a changelog that mentions the api break.

cmd/iaviewer/main.go
```diff
@@ -126,7 +127,7 @@ func ReadTree(dir string, version int, prefix []byte) (*iavl.MutableTree, error)
 		db = dbm.NewPrefixDB(db, prefix)
 	}
 
-	tree, err := iavl.NewMutableTree(db, DefaultCacheSize, false)
+	tree, err := iavl.NewMutableTree(db, DefaultCacheSize, false, log.NewNopLogger())
```
julienrbrt, line 130: I wouldn't use a Nop logger here actually.

tac0turtle: good call

PR #740, opened 2023-04-11
fix: run benchmark only with two dbs 

tac0turtle:
run benchmarks on two dbs for now in ci

this fixes broken ci


Makefile
```diff
@@ -52,18 +52,18 @@ lint-fix:
...
 .PHONY: bench
 
 # fullbench is extra tests needing lots of memory and to run locally
 fullbench:
 	cd benchmarks && \
 		go test $(LDFLAGS) -run=NOTEST -bench=RandomBytes . && \
-		go test $(LDFLAGS) -tags cleveldb,rocksdb,pebbledb -run=NOTEST -bench=Small . && \
-		go test $(LDFLAGS) -tags cleveldb,rocksdb,pebbledb -run=NOTEST -bench=Medium . && \
-		go test $(LDFLAGS) -tags cleveldb,rocksdb,pebbledb -run=NOTEST -timeout=30m -bench=Large . && \
+		go test $(LDFLAGS) -tags rocksdb,pebbledb -run=NOTEST -bench=Small . && \
```
julienrbrt, line 64: Does this means cleveldb is not supported anymore or only that benchmark are not relevant?

tac0turtle: cosmos-db removed it, it wasnt removed here

julienrbrt (approved):

PR #749, opened 2023-04-20
feat: implement lazy set and delete of legacy nodes

cool-develope:
- create the different prefix for the new format node key to avoid the complex when removing nodes
- delete all legacy versions at once in `DeleteVersionsTo` to avoid duplicated codes
- this PR will not include the updating of legacy node to new format


nodedb.go
```diff
@@ -41,11 +41,11 @@ const (
 
 var (
 	// All node keys are prefixed with the byte 'n'. This ensures no collision is
```
yihuang, line 43:
```suggestion
	// All new node keys are prefixed with the byte 's'. This ensures no collision is
```

nodedb.go
```diff
@@ -122,17 +133,32 @@ func (ndb *nodeDB) GetNode(nk []byte) (*Node, error) {
 	ndb.opts.Stat.IncCacheMissCnt()
 
 	// Doesn't exist, load.
-	buf, err := ndb.db.Get(ndb.nodeKey(nk))
```
yihuang, line 125: `nk` is the node key but without prefix?

node.go
```diff
@@ -171,6 +168,69 @@ func MakeNode(nk []byte, buf []byte) (*Node, error) {
...
+		if cause != nil {
+			return nil, fmt.Errorf("deocding node.leftHash, %w", cause)
+		}
+		buf = buf[n:]
+
+		rightHash, _, cause := encoding.DecodeBytes(buf)
+		if cause != nil {
+			return nil, fmt.Errorf("decoding node.rightHash, %w", cause)
+		}
+		node.leftNodeKey = leftHash
```
yihuang, line 228: it should be the node key with prefix here?

yihuang (changes requested):
- should we store the prefixed node key in the `leftChildKey` and `rightChildKey`?
- can you add some unit tests for the mixed nodes scenario.

nodedb.go
cool-develope: yes

cool-develope:
> * should we store the prefixed node key in the `leftChildKey` and `rightChildKey`?

Here, the concern is it will increase the node encoding size

> * can you add some unit tests for the mixed nodes scenario.

I added the legacy db generator commands and unit-tests

cool-develope:
the lazy update is a little tricky, the same problem with `Import/Export`, how to track the current nonce with the specific version when updating the legacy nodes?
@yihuang have you any ideas?

yihuang:
> the lazy update is a little tricky, the same problem with `Import/Export`, how to track the current nonce with the specific version when updating the legacy nodes? @yihuang have you any ideas?

WDYM? we just need to use a increasing nonce?

Issue #747, opened 2023-04-14
Implement a `lazy set` for the migration to the new node key format


cool-develope:
## Context
The migration to the new node key format is tricky. To avoid stopping the network, it introduces a `lazy set`. We will accept the existence of both new and old format nodes in the tree.

## Proposal

- Refactor the `Node` structure to accept both old and new nodes, like this

```go
// Node represents a node in a Tree.
type Node struct {
	key           []byte
	value         []byte
	hash          []byte
	nodeKey       *NodeKey
	leftNodeKey   []byte  // it can be a node hash for the old node and be a node key serialization for the new node.
	rightNodeKey  []byte
	size          int64
	leftNode      *Node
	rightNode     *Node
	subtreeHeight int8
}
```

- Setting new nodes will store the node as a new node key format. It will refactor them to the new format for old nodes that face while updating the tree.

yihuang:
Sounds great, so there are just different versions of node key, new version could reference old version, but not the other way.

cool-develope:
> Sounds great, so there are just multiple versions of node key, new version could reference old version, but not the other way.

exactly

PR #750, opened 2023-04-26
feat: implement the `lazy_set` feature

cool-develope:
Closes: #747

- It will allow loading the tree from the legacy version (old format tree).
- all new nodes will be written in a new format (node key format) after loading it.


cmd/dbgenerator/main.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+const (
+	DefaultCacheSize = 1000
+)
+
+func main() {
+	args := os.Args[1:]
+	if len(args) < 4 {
+		fmt.Fprintln(os.Stderr, "Usage: dbgenerator <dbtype> <dbdir> <random|sequential> <version>")
```
ValarDragon, line 21: Can we get a description for what these are / a readme? (Maybe chatGPT can help?)

Quite cool that you made a tool to help testing / generate sample trees :)

nodedb.go
```diff
@@ -87,22 +100,23 @@ func newNodeDB(db dbm.DB, cacheSize int, opts *Options, lg log.Logger) *nodeDB {
...
+		fastNodeCache:       cache.New(fastNodeCacheSize),
+		versionReaders:      make(map[int64]uint32, 8),
+		storageVersion:      string(storeVersion),
 	}
 }
 
 // GetNode gets a node from memory or disk. If it is an inner node, it does not
 // load its children.
-func (ndb *nodeDB) GetNode(nk *NodeKey) (*Node, error) {
+func (ndb *nodeDB) GetNode(nk []byte) (*Node, error) {
```
ValarDragon, line 119: This feels slightly odd to me. Shouldn't we have at least some type safety here?

Is the idea that theres two different formats of `[]byte` here? Would it be helpful to have a type alias here, or not really?

migrate_test.go
```diff
@@ -0,0 +1,152 @@
...
+	t.Log("Result: " + out.String())
+
+	return err
+}
+
+func TestLazySet(t *testing.T) {
+	legacyVersion := 1000
+	dbType := "goleveldb"
+	dbDir := fmt.Sprintf("legacy-%s-%d", dbType, legacyVersion)
+	relateDir := fmt.Sprintf("./cmd/dbgenerator/%s", dbDir)
```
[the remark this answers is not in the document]
cool-develope: for the above case, the path is visible within the project. It makes it easy to track them, sometimes I can see defer removing is not working properly, especially when panic in debug

nodedb.go
cool-develope: we are using `NodeKey` type for the new format, I will add some descriptions, how about this?

nodedb.go
```diff
@@ -391,13 +527,23 @@ func (ndb *nodeDB) DeleteVersionsFrom(fromVersion int64) error {
...
 }
 
 // DeleteVersionsTo deletes the oldest versions up to the given version from disk.
 func (ndb *nodeDB) DeleteVersionsTo(toVersion int64) error {
+	legacyLatestVersion, err := ndb.getLegacyLatestVersion()
+	if err != nil {
+		return err
+	}
+	// If the legacy version is greater than the toVersion, we don't need to delete anything.
+	// It will delete the legacy versions at onece.
```
tac0turtle, line 542:
```suggestion
	// It will delete the legacy versions at once.
```

tac0turtle (approved):
overall changes look good to me, would like a few more approvals before merging

mutable_tree.go
```diff
@@ -73,6 +73,14 @@ func (tree *MutableTree) IsEmpty() bool {
 
 // VersionExists returns whether or not a version exists.
 func (tree *MutableTree) VersionExists(version int64) bool {
+	legacyLatestVersion, err := tree.ndb.getLegacyLatestVersion()
+	if err != nil {
+		return false
+	}
+	if version <= legacyLatestVersion {
+		has, _ := tree.ndb.hasLegacyVersion(version)
+		return has
```
elias-orijtech, line 82: If ignoring the error is ok, it's clearer to do something like:

```
		has, err := tree.ndb.hasLegacyVersion(version)
		return err == nil && has
```

mutable_tree.go
```diff
@@ -973,10 +999,13 @@ func (tree *MutableTree) balance(node *Node) (newSelf *Node, err error) {
 func (tree *MutableTree) saveNewNodes(version int64) error {
 	nonce := int32(0)
 	newNodes := make([]*Node, 0)
-	var recursiveAssignKey func(*Node) (*NodeKey, error)
-	recursiveAssignKey = func(node *Node) (*NodeKey, error) {
+	var recursiveAssignKey func(*Node) ([]byte, error)
+	recursiveAssignKey = func(node *Node) ([]byte, error) {
 		if node.nodeKey != nil {
-			return node.nodeKey, nil
+			if node.nodeKey.nonce > 0 {
```
elias-orijtech, line 1005: `nonce` is a signed `int32` so I expected `if node.nodeKey.nonce != 0`. Raised https://github.com/cosmos/iavl/issues/763 to avoid similar issues.

node.go
```diff
@@ -110,49 +136,134 @@ func MakeNode(nodeKey *NodeKey, buf []byte) (*Node, error) {
...
+				nonce       int64
+			)
+			leftNodeKey.version, n, cause = encoding.DecodeVarint(buf)
+			if cause != nil {
+				return nil, fmt.Errorf("decoding node.leftNodeKey.version, %w", cause)
+			}
+			buf = buf[n:]
+			nonce, n, cause = encoding.DecodeVarint(buf)
+			if cause != nil {
+				return nil, fmt.Errorf("deocding node.leftNodeKey.nonce, %w", cause)
```
elias-orijtech, line 166: Typo: "decoding".

node.go
```diff
@@ -110,49 +136,134 @@ func MakeNode(nodeKey *NodeKey, buf []byte) (*Node, error) {
...
+			}
+			buf = buf[n:]
+			if nonce < int64(math.MinInt32) || nonce > int64(math.MaxInt32) {
+				return nil, errors.New("invalid nonce, must be int32")
+			}
+			leftNodeKey.nonce = int32(nonce)
```
elias-orijtech, line 172: A better test that avoids depending on the particular type of `nonce` is to convert back to the larger format:

```
leftNodeKey.nonce = int32(nonce)
if nonce != int64(leftNodeKey.nonce) {
    return nil, errors.New("nonce out of range")
}
```

node.go
```diff
@@ -110,49 +136,134 @@ func MakeNode(nodeKey *NodeKey, buf []byte) (*Node, error) {
...
+			node.leftNodeKey = leftNodeKey.GetKey()
+		}
+		if mode&ModeLegacyRightNode != 0 { // legacy rightNodeKey
+			node.rightNodeKey, _, cause = encoding.DecodeBytes(buf)
```
elias-orijtech, line 176: Rename `cause` to `err` while we're here?

node.go
```diff
@@ -110,49 +136,134 @@ func MakeNode(nodeKey *NodeKey, buf []byte) (*Node, error) {
...
+			}
+			buf = buf[n:]
+			nonce, _, cause = encoding.DecodeVarint(buf)
+			if cause != nil {
+				return nil, fmt.Errorf("decoding node.rightNodeKey.nonce, %w", cause)
+			}
+			if nonce < int64(math.MinInt32) || nonce > int64(math.MaxInt32) {
+				return nil, errors.New("invalid nonce, must be int32")
+			}
+			rightNodeKey.nonce = int32(nonce)
```
elias-orijtech, line 197: Same comment as above.

node.go
```diff
@@ -110,49 +136,134 @@ func MakeNode(nodeKey *NodeKey, buf []byte) (*Node, error) {
...
+func MakeLegacyNode(hash, buf []byte) (*Node, error) {
+	// Read node header (height, size, version, key).
+	height, n, cause := encoding.DecodeVarint(buf)
+	if cause != nil {
+		return nil, fmt.Errorf("decoding node.height, %w", cause)
+	}
+	buf = buf[n:]
+	if height < int64(math.MinInt8) || height > int64(math.MaxInt8) {
+		return nil, errors.New("invalid height, must be int8")
+	}
```
elias-orijtech, line 214: I suggest the same test as above:

```
height8 := int8(height)
if height != int64(height8) {
    return nil, errors.New("height out of range")
}
```

and use `height8` verbatim below.

nodedb.go
```diff
@@ -438,24 +592,44 @@ func (ndb *nodeDB) DeleteFastNode(key []byte) error {
...
+func (ndb *nodeDB) legacyNodeKey(nk []byte) []byte {
+	return legacyNodeKeyFormat.Key(nk)
+}
+
+func (ndb *nodeDB) legacyRootKey(version int64) []byte {
+	return legacyRootKeyFormat.Key(version)
+}
+
 func (ndb *nodeDB) getFirstVersion() (int64, error) {
 	if ndb.firstVersion == 0 {
```
elias-orijtech, line 612: Nit: early return when `firstVersion` is set:

```
if ndb.firstVersion != 0 {
    return ndb.firstVersion, nil
}
```

tac0turtle:
@cool-develope seems some tests are still failing

cmd/dbgenerator/main.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	return t, err
+}
+
+func randBytes(length int) []byte {
+	key := make([]byte, length)
+	// math.rand.Read always returns err=nil
+	// we do not need cryptographic randomness for this test:
+	rand.Read(key)
```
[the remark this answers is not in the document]
cool-develope: we are using math/rand in generating random integers

migrate_test.go
```diff
@@ -0,0 +1,152 @@
...
+	"os/exec"
+	"testing"
+
+	"cosmossdk.io/log"
+	dbm "github.com/cosmos/cosmos-db"
+	"github.com/stretchr/testify/require"
+)
+
+func createLegacyTree(t *testing.T, dbType, dbDir string, version int) error {
+	cmd := exec.Command("sh", "-c", fmt.Sprintf("cd cmd/dbgenerator && go run main.go %s %s random %d", dbType, dbDir, version)) //nolint:gosec
```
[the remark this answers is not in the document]
cool-develope: it means, the legacy iavl version is imported itself, it is not possible in the current versions since we are in v1.0

cool-develope:
> @cool-develope seems some tests are still failing

it seems like  the dbdir removing is not working properly in GA

cmd/dbgenerator/main.go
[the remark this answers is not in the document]
elias-orijtech: Yes, what I meant was that `math.rand.Read` is deprecated and will generate linter warnings.

diff.go
```diff
@@ -21,7 +21,7 @@ type KVPairReceiver func(pair *KVPair) error
 //
 // The algorithm don't run in constant memory strictly, but it tried the best the only
 // keep minimal intermediate states in memory.
-func (ndb *nodeDB) extractStateChanges(prevVersion int64, prevRoot *NodeKey, root *NodeKey, receiver KVPairReceiver) error {
+func (ndb *nodeDB) extractStateChanges(prevVersion int64, prevRoot, root []byte, receiver KVPairReceiver) error {
```
kocubinski, line 24: I'm a big fan of this change in particular. Actually it's what I originally recommended. https://github.com/cosmos/iavl/pull/676#pullrequestreview-1329237749

node.go
```diff
@@ -18,27 +18,51 @@ import (
...
 // Node represents a node in a Tree.
 type Node struct {
 	key           []byte
 	value         []byte
 	hash          []byte
 	nodeKey       *NodeKey
-	leftNodeKey   *NodeKey
-	rightNodeKey  *NodeKey
+	leftNodeKey   []byte
+	rightNodeKey  []byte
```
kocubinski, line 65: 🤩

kocubinski: I propose taking this further.  Why not defined `nodeKey as []byte` and store version and nonce as private, top-level fields on `Node`?

migrate_test.go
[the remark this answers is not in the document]
kocubinski: I agree that `t.TmpDir()`is idiomatic.  What do you mean by visible within the project? Generally a devs workflow should not take precedence over established idioms.

nodedb.go
```diff
@@ -363,14 +394,94 @@ func (ndb *nodeDB) deleteVersion(version int64) error {
...
-	return ndb.traverseOrphans(version, func(orphan *Node) error {
-		return ndb.batch.Delete(ndb.nodeKey(orphan.nodeKey))
+	return ndb.traverseOrphans(version, version+1, func(orphan *Node) error {
+		return ndb.batch.Delete(ndb.nodeKey(orphan.GetKey()))
+	})
+}
+
+// deleteLegacyNodes deletes all legacy nodes with the given version from disk.
+// NOTE: This is only used for DeleteVersionsFrom.
+func (ndb *nodeDB) deleteLegacyNodes(version int64, nk []byte) error {
```
kocubinski, line 410: Is there test coverage for these new `deleteLegacy*` methods?

kocubinski:
> > @cool-develope seems some tests are still failing
> 
> it seems like the dbdir removing is not working properly in GA

we should probably use `t.TmpDir()` https://github.com/cosmos/iavl/pull/750#discussion_r1178626555

node.go
cool-develope: `[]byte nodeKey` is meaningless, `leftNodeKey` and `rightNodeKey` is used in getting children
The only thing we need is `nonce` and `version`

nodedb.go
cool-develope: yeah, it will be ussed in `DeleteVersionsTo` and it is covered in `migrate_test.go`

migrate_test.go
```diff
@@ -0,0 +1,160 @@
...
+func createLegacyTree(t *testing.T, dbType, dbDir string, version int) error {
+	relateDir := fmt.Sprintf("./cmd/dbgenerator/%s", dbDir)
+	if _, err := os.Stat(relateDir); err == nil {
+		err := os.RemoveAll(relateDir)
+		if err != nil {
+			t.Errorf("%+v\n", err)
+		}
+	}
+
+	cmd := exec.Command("sh", "-c", fmt.Sprintf("cd cmd/dbgenerator && go run . %s %s random %d %d", dbType, dbDir, version, version/2)) //nolint:gosec
```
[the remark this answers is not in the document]
cool-develope: it will cause circular dependency since we are in v1, `go.mod` doesn't analyze the `iavl` imports

node.go
kocubinski: `NodeKey` seems also without meaning, just makes the code harder to read.  if we need `nonce` and `version` why not put it on `Node` directly?

cool-develope: good question, the nonce and version are used together in several encodings, so I created the specific struct `NodeKey`
and also it ensures `version+nonce` as a node key

cool-develope:
@tac0turtle , it seems like 5mins is not enough for test

PR #758, opened 2023-05-03
perf: reduce use of unsafe functions, and make them safer

elias-orijtech:
The safer functions require Go 1.20. Let me know if that's not acceptable.

CC @odeke-em

### DIFF
--- a/mutable_tree.go
+++ b/mutable_tree.go
@@ -138,7 +138,7 @@ func (tree *MutableTree) Get(key []byte) ([]byte, error) {
 	}
 
 	if !tree.skipFastStorageUpgrade {
-		if fastNode, ok := tree.unsavedFastNodeAdditions[ibytes.UnsafeBytesToStr(key)]; ok {
+		if fastNode, ok := tree.unsavedFastNodeAdditions[string(key)]; ok {
 			return fastNode.GetValue(), nil
 		}
 		// check if node was deleted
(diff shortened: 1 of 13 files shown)


odeke-em:
Thank you @elias-orijtech! Kindly cc-ing @tac0turtle @JeancarloBarrios for a question about when Cosmos will switch to Go1.20. I think let's move this PR forward once the Cosmos-SDK is also using Go1.20.*

mutable_tree.go
```diff
@@ -138,7 +138,7 @@ func (tree *MutableTree) Get(key []byte) ([]byte, error) {
 	}
 
 	if !tree.skipFastStorageUpgrade {
-		if fastNode, ok := tree.unsavedFastNodeAdditions[ibytes.UnsafeBytesToStr(key)]; ok {
+		if fastNode, ok := tree.unsavedFastNodeAdditions[string(key)]; ok {
```
yihuang, line 141: is it because the string conversion is the same as the unsafe version in go 1.20? or because the performance is impact illegible.

elias-orijtech: map[string([]byte)] is a no-copy, no garbage operation in Go since some releases ago.

It's the same performance, but safe.

elias-orijtech:
Btw, I only see "squash and merge" as option. I'd prefer "rebase and merge" or another option that preserves the individual commits.

Also, let me know if you expect me or someone else to press the merge button :)

elias-orijtech:
Maybe "perf:" is the wrong prefix, because there are no speedups in this PR, only safeness. Is "security:" better, or something else?

elias-orijtech:
@tac0turtle did you intend to take only the latest commit in the series? I think https://github.com/cosmos/iavl/pull/758/commits/36a73f22f5314b2656a0047dfa747b6d8eda716d and https://github.com/cosmos/iavl/pull/758/commits/99f583e77d81f32d686db05d3171d34a1c11e45c should go in too, and they don't depend on Go 1.20.

PR #768, opened 2023-05-11
feat: Change the v0.20.x line to use cosmos DB

ValarDragon:
This is done so we can merge #765


ValarDragon:
This CI lint seems wrong? I tried changing both of the nolint's in the codebase, but they do warn for a gocritic problem. (Also these are unchanged by this PR)

ValarDragon:
cc @tac0turtle or anyone who worked on the initial cosmos DB move, any idea why benchmarks is getting build errors for rocksDB and how to fix it?

(Also is this something thats acceptable to merge into v0.20.x line?)

tac0turtle:
I can look into this. We have different versions for things due to the comet migration so it's confusing.

julienrbrt:
> I can look into this. We have different versions for things due to the comet migration so it's confusing.

I was looking as well lol -> https://github.com/cosmos/iavl/pull/769
I think we cannot because this will break v0.47.x users.

ValarDragon:
oh yikes thats super confusing. OK :/

Guess I have to revive the IAVL fork until Osmosis updates its SDK

tac0turtle:
@ValarDragon which version of the sdk do you need this for?

ValarDragon:
Its for SDK v0.45! But I think it shouldn't be a big deal to make a more complicated IAVL change here to track the batch size in IAVL instead of the DB layer for this release line.

PR #769, opened 2023-05-12
docs: add cosmos sdk compatibility table

julienrbrt:
I have every time to lookup which versions are compatible with which IAVL version and tm-db/cometbft-db/cosmos-db version.
This adds a table, so we do not need to look it up all the time.

Feel free to close it if we do not need to add this to the repo, at least now I still have this PR to refer myself to.

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -1,6 +1,6 @@
 # IAVL x Cosmos SDK
 
-| IAVL                                                           | DB Interfaces                                            | Cosmos SDK       |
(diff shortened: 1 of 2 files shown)


tac0turtle (approved):

VERSIONS.md
```diff
@@ -0,0 +1,8 @@
+# IAVL x Cosmos SDK
```
tac0turtle, line 1: we should add this to the bottom of the readme

julienrbrt: done

PR #771, opened 2023-05-18
Add BatchWithFlusher to iavl tree

### LINKED ISSUE #619: Refactor One Large Commit Into Smaller Batches
Proposed in-collaboration with: @ValarDragon @catShaark

## Introduction

Our commit logic does not take into account the constraints of the underlying LSM key-value store backends. As a result, a lot of performance is wasted.

This issue is a proposal to refactor the commit logic to be more aligned with the underlying LSM backends.

### Separate One Large Commit into Smaller Batches

We commit once per SDK store, at the end of every block, in ABCI's `Commit()`. Currently, this logical commit translates directly to a single atomic database commit per SDK store.

Every write incurs some "merkelization overhead" (all of the IAVL inner nodes which must be written). As a result, it is typical to have store writes of size 1-10MB, reaching gigabytes in certain operations. For example, importing blockchain genesis or processing Osmosis Epoch.

We claim here that separating one atomic database commit per store, into several smaller atomic commits, is of significant performance benefit. We sketch the reasons below:

* Committing 1GB atomically is much slower than doing 10,000 commits of size 100KB, potentially concurrently, and adding a 'wait group' to tell us when all commits have finished.
    * Ethereum has found that 100KB is optimal, and in general most databases are optimized for 100s of KBs to a few MBs [1].
* Most relevant DB implementations use contiguous segments of memory for handling changes related to the batch. So even describing these batches requires high amounts of RAM at once (which in Go translates to eventual stop-the-world garbage collecting) [7].
* A large batch composed of many small records can require twice as much memory when inserted into an LSM memtable than it is required in the batch itself. Note, that this causes a temporary increase in memory requirements because the batch memory is not freed until it is completely committed [2].
* There is a compactions overhead. Compactions happen periodically and merge the LSM SSTables. This helps to keep the number of SSTable files and read performance within acceptable bounds. When we commit a large batch at once, we have to pay this overhead immediately, causing large spikes in RAM and CPU usage [3]. Some LSM designs might even cause OOM due to space amplification [4].

The large commit problem is especially notable when importing genesis. Since it persists all genesis data in memory, Osmosis nodes require at least 64 GB RAM to survive the import [5].

To mitigate these problems, we suggest saving to disk in smaller atomic segments instead of one large.

### Batch Pre-Allocation

Additionally, LevelDB and, potentially, other backends benefit from batch pre-allocation [6]. By limiting the batch max size to a configurable number, we can pre-allocate it according to the limit.

### Step 1: Define `GetSizeBytes` on the `Batch` interface

Refactor `cosmos-db` [`Batch`](https://github.com/cosmos/cosmos-db/blob/9aa6a118b2ad73dfd798cf92e9908c52f100becf/types.go#L78) interface to have a method `GetSizeBytes` that returns the current size of the batch in bytes.

### Step 2: implement `FlushThresholdBatchDecorator`

Introduce a new component called `FlushThresholdBatchDecorator`:
```go
// ThresholdFlushBatchDecorator is a 
// decorator around batch that flushes
// to disk as soon as the configurable limit 
// is reached.
type FlushThresholdBatchDecorator struct {
    batch                   db.Batch        // Batched writing buffer.
    batchSizeFlushThreshold uint64 // The maximum size of the batch in bytes before it gets flushed to disk
}

var _ db.Batch = FlushThresholdBatchDecorator{} 

// WithFlushThreshold returns new batch decorated with flush threshold.
func WithFlushThreshold(batch db.Batch, flushThreshold uint64) db.Batch {
    // stub
    
    // pre-allocate the underlying batch buffer to flushThreshold if applicable to the db backend.
}

// Set sets value at the given key to the db.
// If the set grows the underlying batch until batchSizeFlushThreshold is reached,
// the batch is flushed to disk, cleared, and a new one is created with buffer
// pre-allocated to threshold.
func Set (key []byte, value []byte) {
    // stub
    
    // N.B.: use `GetSizeBytes` to
    // keep track of the current batch size.
}

// More db.Batch methods
```

### Step 3: Integrate `FlushThresholdBatchDecorator` into `nodedb`

- Introduce an option for for "maximum batch size" [here](https://github.com/cosmos/iavl/blob/58a0467e2b0ea55075b85b0513875592391ae073/options.go#L89).

- Propagate the option all the way to `nodeDb`.

- Replace the batch [here](https://github.com/cosmos/iavl/blob/58a0467e2b0ea55075b85b0513875592391ae073/nodedb.go#L99) with the new decorated `FlushThreshold` batch.

### Step 4: Benchmark and Profile

Compare the performance difference from the old results.

- The profiles taken on SDK with IAVL `v0.19.4`:
https://drive.google.com/drive/u/0/folders/1eMenapAeEjIJmiSEHrqQtoQ_g1KCN0YJ

- SDK branch from where profiles were taken: https://github.com/osmosis-labs/cosmos-sdk/tree/roman/profiles

We will need to implement a new benchmark that is isolating the `SaveVersion()` method and run it against trees of different sizes before and after the suggested design is implemented.

### Step 5: Investigate running commits in-parallel

There is no reason for blocking the execution on the overhead stemming from the initialization of multiple batches by the underlying LSM kv-store.

As a result, we can explore running the commits concurrently, or even in parallel.

[1]:https://github.com/ethereum/go-ethereum/pull/15115
[2]:https://github.com/cockroachdb/pebble/blob/master/docs/rocksdb.md#indexed-batches
[3]:http://www.benstopford.com/2015/02/14/log-structured-merge-trees/
[4]:https://dl.acm.org/doi/pdf/10.1145/3183713.3196927
[5]:https://github.com/osmosis-labs/osmosis/issues/2953
[6]:https://github.com/osmosis-labs/iavl/issues/32
[7]:https://github.com/syndtr/goleveldb/blob/master/leveldb/batch.go#L61-L74

catShaark:
Closes #619 

* This Pr allow iavl tree to use BatchWithFlusher to limit size of commit to db. By including flushThreshold to nodeDB, we can configure the size limit of each commit batch.
* This Pr changes `nodeDB` struct and `NewMutableTreeWithOpts` to replace uses of `dbm.Batch` to `BatchWithFlusher`
* We'll have to make changes to sdk accordingly to adapt this changes.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -16,10 +16,13 @@ type BatchWithFlusher struct {
 var _ dbm.Batch = &BatchWithFlusher{}
 
 // Ethereum has found that commit of 100KB is optimal, ref ethereum/go-ethereum#15115
-// var defaultFlushThreshold = 100000
+var defaultFlushThreshold = 100000
 
 // NewBatchWithFlusher returns new BatchWithFlusher wrapping the passed in batch
 func NewBatchWithFlusher(db dbm.DB, flushThreshold int) *BatchWithFlusher {
+	if flushThreshold <= 0 {
+		panic("flushThreshold can't be zero or negative")
+	}
 	return &BatchWithFlusher{
 		db:             db,
 		batch:          db.NewBatchWithSize(flushThreshold),
--- a/batch.go
+++ b/batch.go
@@ -16,7 +16,7 @@ type BatchWithFlusher struct {
 var _ dbm.Batch = &BatchWithFlusher{}
 
 // Ethereum has found that commit of 100KB is optimal, ref ethereum/go-ethereum#15115
-var defaultFlushThreshold = 100000
+var DefaultFlushThreshold = 100000
 
 // NewBatchWithFlusher returns new BatchWithFlusher wrapping the passed in batch
 func NewBatchWithFlusher(db dbm.DB, flushThreshold int) *BatchWithFlusher {
--- a/benchmarks/bench_test.go
+++ b/benchmarks/bench_test.go
@@ -27,7 +27,7 @@ func randBytes(length int) []byte {
 }
 
 func prepareTree(b *testing.B, db db.DB, size, keyLen, dataLen int) (*iavl.MutableTree, [][]byte) {
-	t, err := iavl.NewMutableTreeWithOpts(db, size, nil, false, log.NewNopLogger())
+	t, err := iavl.NewMutableTreeWithOpts(db, size, nil, false, log.NewNopLogger(), iavl.DefaultFlushThreshold)
 	require.NoError(b, err)
 	keys := make([][]byte, size)
 
--- a/immutable_tree.go
+++ b/immutable_tree.go
@@ -38,7 +38,8 @@ func NewImmutableTreeWithOpts(db dbm.DB, cacheSize int, opts *Options, skipFastS
 	return &ImmutableTree{
 		logger: lg,
 		// NodeDB-backed Tree.
-		ndb:                    newNodeDB(db, cacheSize, opts, lg),
+		// flushthreshold is 0 since there's no batch commit action.
+		ndb:                    newNodeDB(db, cacheSize, opts, lg, 0),
 		skipFastStorageUpgrade: skipFastStorageUpgrade,
 	}
 }
--- a/import.go
+++ b/import.go
@@ -48,7 +48,7 @@ func newImporter(tree *MutableTree, version int64) (*Importer, error) {
 	return &Importer{
 		tree:    tree,
 		version: version,
-		batch:   tree.ndb.db.NewBatch(),
+		batch:   NewBatch(tree.ndb.db, tree.ndb.flushThreshold),
 		stack:   make([]*Node, 0, 8),
 		nonces:  make([]int32, version+1),
 	}, nil
@@ -84,7 +84,7 @@ func (i *Importer) writeNode(node *Node) error {
 			return err
 		}
 		i.batch.Close()
-		i.batch = i.tree.ndb.db.NewBatch()
+		i.batch = NewBatch(i.tree.ndb.db, i.tree.ndb.flushThreshold)
 		i.batchSize = 0
 	}
 
--- a/iterator_test.go
+++ b/iterator_test.go
@@ -369,7 +369,7 @@ func TestNodeIterator_Success(t *testing.T) {
 }
 
 func TestNodeIterator_WithEmptyRoot(t *testing.T) {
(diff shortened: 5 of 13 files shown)


batch.go
```diff
@@ -16,10 +16,13 @@ type BatchWithFlusher struct {
 var _ dbm.Batch = &BatchWithFlusher{}
 
 // Ethereum has found that commit of 100KB is optimal, ref ethereum/go-ethereum#15115
-// var defaultFlushThreshold = 100000
+var DefaultFlushThreshold = 100000
 
 // NewBatchWithFlusher returns new BatchWithFlusher wrapping the passed in batch
 func NewBatchWithFlusher(db dbm.DB, flushThreshold int) *BatchWithFlusher {
+	if flushThreshold <= 0 {
```
tac0turtle, line 23: should fllushthreshold be a unsigned integer here? not sure when it would be less than 0

catShaark: it's because `batch.GetByteSize()` return `int`, we want `flushThreshold` to also be int so we don't have to do any conversion from `int` to `uint`

tac0turtle:
will this be able to be part of the next release or should we wait for the next release (unclear when that will be)

tac0turtle:
ping

catShaark:
I figure out the deadlock issue is due to the mem db iterator being channel based, the iterator will lock the db during iteration which inturn cause the deadlock

tac0turtle:
The sdk doesn't use memdb anymore. Maybe we have our own memdb here

catShaark:
So I have run benchmark on a live osmosis node. Comparing normal iavl batch vs batch with flusher (flush to state at 100KB), the result is almost the same for the case of normal blocks ( blocks that don't cause heavy commit to state ) but for epoch blocks that cause heavy commit, the commit time is reduced by 25%.

catShaark:
the benchmark also proves that iavl with batch is not state breaking. I've run 10000 blocks with no app hash error so far

tac0turtle:
amazing this is super useful!!

catShaark:
yes, we can have this on mainnets. Do we want the flush thresh hold to be configurable. If so then we have to change `iavl.NewMutableTreeWithOpts` which is used in sdk root multi logic, that means we also need to make changes to the sdk

catShaark:
here is the commit time of a osmosis epoch block :
|Flush threshold  | Commit time |
| ------------- | ------------- |
|25| 27.17s|
|50| 26.72s|
|100| 26.8s|
|250| 26.19s|
|500| 27.5s|
|750| 26.76s|
|1000| 41.83s|
|no-flush| 49.61s|

catShaark:
thanks for pinging, I will finish this by tomorow

catShaark:
I've made a new PR for this
https://github.com/cosmos/iavl/pull/807

Issue #772, opened 2023-05-19
Memory leak during State Sync


GAtom22:
## Description

Trying to perform the state sync on an Evmos node noticed that the process runs OOM.

Using `GODEBUG=gctrace=1` can be observed that the heap used increases while  applying the snapshot chunks.
Filtering only the gctrace logs when fetching and applying the snapshot chunks we can see the size of the heap continues to grow over time:

```
gc 34 @23.159s 0%: 0.046+1.5+0.004 ms clock, 0.37+1.4/2.6/5.7+0.037 ms cpu, 4650->4650->4650 MB, 4650 MB goal, 0 MB stacks, 1 MB globals, 8 P
gc 35 @23.172s 0%: 0.093+678+0.025 ms clock, 0.74+0/2.9/520+0.20 ms cpu, 5803->5803->3497 MB, 5803 MB goal, 0 MB stacks, 1 MB globals, 8 P
gc 36 @24.396s 0%: 0.047+1.7+0.008 ms clock, 0.38+0/3.2/7.2+0.067 ms cpu, 4201->4201->2706 MB, 4201 MB goal, 0 MB stacks, 1 MB globals, 8 P
gc 37 @24.841s 0%: 0.048+1.6+0.008 ms clock, 0.38+0/2.7/7.0+0.067 ms cpu, 3261->3261->2984 MB, 3261 MB goal, 0 MB stacks, 1 MB globals, 8 P
gc 38 @25.315s 0%: 0.045+1.7+0.008 ms clock, 0.36+0/3.1/6.8+0.066 ms cpu, 3581->3581->3282 MB, 3581 MB goal, 0 MB stacks, 1 MB globals, 8 P
gc 39 @25.874s 0%: 0.042+5.3+0.004 ms clock, 0.33+0/3.7/11+0.036 ms cpu, 3944->3944->3602 MB, 3944 MB goal, 0 MB stacks, 1 MB globals, 8 P
gc 40 @40.751s 0%: 0.10+2.3+0.009 ms clock, 0.87+0.21/4.4/10+0.075 ms cpu, 5347->5347->2349 MB, 5347 MB goal, 0 MB stacks, 1 MB globals, 8 P
gc 41 @40.912s 1%: 0.044+344+298 ms clock, 0.35+0.25/195/10+2390 ms cpu, 3502->4655->3502 MB, 3502 MB goal, 0 MB stacks, 1 MB globals, 8 P
[...]

PR #773, opened 2023-05-19
fix: memory leak in `Import`

yihuang:
Closes: #772 

don't need to reference the children, which recursively retain the whole subtree in memory.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
 - [#718](https://github.com/cosmos/iavl/pull/718) Fix `traverseNodes` unexpected behaviour
 - [#770](https://github.com/cosmos/iavl/pull/770) Add `WorkingVersion()int64` API.
 
+### Bug Fixes
+
+- [#]() Fix memory leak in `Import`.
+
 ### Breaking Changes
 
 - [#735](https://github.com/cosmos/iavl/pull/735) Pass logger to `NodeDB`, `MutableTree` and `ImmutableTree`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 
 ### Bug Fixes
 
-- [#]() Fix memory leak in `Import`.
+- [#773](https://github.com/cosmos/iavl/pull/773) Fix memory leak in `Import`.
 
 ### Breaking Changes
 
--- a/import.go
+++ b/import.go
@@ -136,12 +136,6 @@ func (i *Importer) Add(exportNode *ExportNode) error {
 		leftNode := i.stack[stackSize-2]
 		rightNode := i.stack[stackSize-1]
 
-		// remove the recursive references to avoid memory leak
-		leftNode.leftNode = nil
-		leftNode.rightNode = nil
-		rightNode.leftNode = nil
-		rightNode.rightNode = nil
-
 		node.leftNode = leftNode
 		node.rightNode = rightNode
 		node.leftNodeKey = leftNode.nodeKey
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,10 @@
 - [#718](https://github.com/cosmos/iavl/pull/718) Fix `traverseNodes` unexpected behaviour
 - [#770](https://github.com/cosmos/iavl/pull/770) Add `WorkingVersion()int64` API.
 
+### Bug Fixes
+
+- [#]() Fix memory leak in `Import`.
```
yihuang, line 15:
```suggestion
- [#773](https://github.com/cosmos/iavl/pull/773) Fix memory leak in `Import`.
```

nodedb.go
```diff
@@ -55,7 +55,7 @@ var (
 	metadataKeyFormat = keyformat.NewKeyFormat('m', 0) // m<keystring>
 )
 
-var errInvalidFastStorageVersion = fmt.Errorf("Fast storage version must be in the format <storage version>%s<latest fast cache version>", fastStorageVersionDelimiter)
+var errInvalidFastStorageVersion = fmt.Errorf("fast storage version must be in the format <storage version>%s<latest fast cache version>", fastStorageVersionDelimiter)
```
joshlopes, line 58: Just curious why this change

import.go
```diff
@@ -133,19 +133,28 @@ func (i *Importer) Add(exportNode *ExportNode) error {
...
-		node.rightNodeKey = node.rightNode.nodeKey
-		node.size = node.leftNode.size + node.rightNode.size
+		leftNode := i.stack[stackSize-2]
+		rightNode := i.stack[stackSize-1]
+
+		node.leftNode = leftNode
+		node.rightNode = rightNode
+		node.leftNodeKey = leftNode.nodeKey
+		node.rightNodeKey = rightNode.nodeKey
+		node.size = leftNode.size + rightNode.size
```
joshlopes, line 143: It seems to me that you moved leftNode and rightNode to vars and re-used, but isn't that the same has `node.leftNode` and `node.rightNode` or does GO have an issue with using the object?

import.go
```diff
@@ -133,19 +133,28 @@ func (i *Importer) Add(exportNode *ExportNode) error {
...
+		if err := i.writeNode(rightNode); err != nil {
 			return err
 		}
 		i.stack = i.stack[:stackSize-2]
+
+		// remove the recursive references to avoid memory leak
+		leftNode.leftNode = nil
+		leftNode.rightNode = nil
+		rightNode.leftNode = nil
+		rightNode.rightNode = nil
```
joshlopes, line 157: Guess this is the actual fix?

joshlopes:
@tac0turtle why are you removing comments ? 👎

tac0turtle:
removing what comments? i didnt touch anything other than merge

nodedb.go
yihuang: There a lint warns about the leading upper case in string

import.go
yihuang: Yes, it's equivalent, it's just I think it makes code shorter this way, because the node.leftNode will be repeated many times

import.go
yihuang: Yes

joshlopes:
@tac0turtle the comments are back, probably a hiccup within GH API, basically I left the comments above and you merged the PR - I saw the event merged and the comments were no were so I assumed you had deleted them to merge/don't stale. 

Apologies! 🙇

PR #774, opened 2023-05-19
fix: memory leak in `Import` (backport #773)

yihuang:
don't reference the children, which recursively retain the whole subtree in memory.


fedekunze (approved):

fedekunze:
@tac0turtle @alexanderbez can you review this one too?

tac0turtle (approved):
utACK

PR #777, opened 2023-05-24
refactor: Drop error results from functions that never fail

### LINKED ISSUE #756: Drop error result from NewMutableTree/NewMutableTreeWithOpts
Both constructors always return the `nil` error.

@odeke-em

### LINKED ISSUE #757: Drop error result from Node.hashWithCount
`Node.hashWithCount` can only return errors in two cases:

https://github.com/cosmos/iavl/blob/64ffc2749400f762d6279db0fd1ccc2b22e69492/node.go#L336-L343

The first case is its own error result, recursively. The second one is the error result from `hash.Hash.Write` which is documented to never return an error.

CC @odeke-em

elias-orijtech:
The hashing procedure can never fail, and not returning an error frees up callers from worrying about it.

Also includes a performance optimization.

Fixes #757.
Fixes #756.

CC @odeke-em


tac0turtle (approved):
UtAck

tac0turtle:
could we get a changelog for the api break

elias-orijtech:
> could we get a changelog for the api break also linting is failing

Done.

tac0turtle:
could we fix the linting issues as well please

PR #780, opened 2023-05-25
fix: the failure of load the legacy empty root

cool-develope:


tac0turtle (approved):

nodedb.go
```diff
@@ -763,13 +767,16 @@ func (ndb *nodeDB) GetRoot(version int64) ([]byte, error) {
...
+			return nil, nil
+		}
 		return val, nil
 	}
 	if len(val) == 0 { // empty root
 		return nil, nil
 	}
 	if isReferenceToRoot(val) { // point to the prev root
-		return val[1:], nil
+		return append(val[1:], 0, 0, 0, 1), nil
```
tac0turtle, line 779: what is this change for?

tac0turtle (commented):
one question but otherwise looks good

tac0turtle (commented):
one question but otherwise looks good

nodedb.go
cool-develope: previously, we were storing the node key as a reference node, now we are only storing the reference version without the nonce, since the root nonce is always 1

nodedb.go
```diff
@@ -781,8 +788,8 @@ func (ndb *nodeDB) SaveEmptyRoot(version int64) error {
 }
 
 // SaveRoot saves the root when no updates.
-func (ndb *nodeDB) SaveRoot(version int64, prevRootKey *NodeKey) error {
-	return ndb.batch.Set(nodeKeyFormat.Key(GetRootKey(version)), ndb.nodeKey(prevRootKey.GetKey()))
+func (ndb *nodeDB) SaveRoot(version, prevVersion int64) error {
```
cool-develope, line 791: @tac0turtle , ^|^

tac0turtle (approved):

PR #785, opened 2023-05-30
chore: remove TODO

elias-orijtech:
CC @odeke-em


elias-orijtech:
Someone got to this before me. Reduced PR to just removing the TODO.

Issue #737, opened 2023-04-10
Proposed new repo API


ValarDragon:
Currently the IAVL interfaces [here](https://pkg.go.dev/github.com/cosmos/iavl) make no sense, and as a huge part of that, touching IAVL is always incredibly overly complex.

This issue proposes some first-pass ways that I think we should be able to get out with confidence, and significantly reduce complexity of working with this code base.

Lets recap what the existing parts of our code are rn:
- MutableTree
  - This is the entry point for end-users, and is imo the bulk of why the code is so confusing. It serves the purpose of:
    - Managing versions. So the caller uses this to delete old versions, see what versions are availalbe
  - Managing WIP state being committed to and built up
    - (where your set API is)
  - Getting data at old versions, and latest versions
  - Getting Read-Only copies of old versions. 
- ImmutableTree
  - A read-only copy of a version. Most of the time old versions, but can also be taken out on latest version. Guarantees are at least unclear to me, for what happens when taking it for latest version while MutableTree has rights
- NodeDB
  - an internal only object, thats responsible for: Reading/Writing nodes / branches to disk, along the tree structure
  - Responsible for saving/deleting versions
  - The object passed into "everything"

Phased proposals:

## 1. Make a DB version manager API as the entrypoint

So make a new struct, that first acts as something intended to manage the various versions. 
TODO: Consider renaming version here to "Block Height"

TreeReader for now is just a type alias for ImmutableTree

```go
NewIavlDB(database, options) IavlDB

func (*IavlDB) VersionExists(version uint64) bool
func (*IavlDB) GetLatestVersion) TreeReader
func (*IavlDB) GetVersion(version uint64) TreeReader
// do whatever delete APIs we want to use.
func (*IavlDB) DeleteVersion(version uint64) error
func (*IavlDB) Rollback() error

// Only one WriteBatch can be open at once. SaveVersion deletes that write batch.
func (*IavlDB) StartWriteBatch() WriteBatch
func (*IavlDB) SaveVersion(batch WriteBatch) error

// TODO: Consider deleting?
func (*IavlDB) SetInitialHeight(version uint64)
```

TreeReader in the above is just a name alias for immutable tree.
In phase 1, the DB version manager can maintain a mutable tree instance, and have write batch just be that.

## 2. Refactor WriteBatch to be its own type

Basically move all of the logic in the MutableTree.Set to be done on the new WriteBatch struct.
Move the SaveVersion logic from MutableTree onto IavlDB

Delete MutableTree

[...]

PR #790, opened 2023-06-12
feat: ADR for api cleanup and async commit

cool-develope:


docs/architecture/adr-002-api-cleanup.md
```diff
@@ -0,0 +1,98 @@
...
+    Tree interface {
+        Has(key []byte) (bool, error)
+        Get(key []byte) ([]byte, error)
+        SaveVersion(batch struct{
+            keys [][]btye
+            values [][]byte
+        }) ([]byte, int64, error)
+        Version() int64
+        Hash() []byte
+        WorkingHash() []byte
```
yihuang, line 66: Is there a difference in `Hash` and `WorkingHash`, they both return the root hash of current tree?

docs/architecture/adr-002-api-cleanup.md
```diff
@@ -0,0 +1,98 @@
...
+The modified API will have the following structure:
+
+```go
+    Tree interface {
+        Has(key []byte) (bool, error)
+        Get(key []byte) ([]byte, error)
+        SaveVersion(batch struct{
+            keys [][]btye
+            values [][]byte
+        }) ([]byte, int64, error)
```
yihuang, line 63: Why not reuse the existing `ChangeSet` struct for the parameter.

docs/architecture/adr-002-api-cleanup.md
```diff
@@ -0,0 +1,98 @@
...
+        Hash() []byte
+        WorkingHash() []byte
+        VersionExists(version int64) bool
+        DeleteVersionsTo(version int64) error
+        GetVersioned(key []byte, version int64) ([]byte, *cmtprotocrypto.ProofOps, error)
+        GetTree(version int64) (*iavl.Tree, error)
+        SetInitialVersion(version uint64)
+        Iterator(start, end []byte, ascending bool) (types.Iterator, error)
+        LoadVersionForOverwriting(targetVersion int64) error
+    }
```
tac0turtle, line 74: How do you envision state import and export working, or do you think this wont be part of the tree interface instead it would be handled by the app?

docs/architecture/adr-002-api-cleanup.md
```diff
@@ -0,0 +1,98 @@
...
+### API Cleanup
+
+- Merge `ImmutableTree` and `MutableTree` into a single `Tree` structure.
+- Make `nodeDB` methods private.
+- Separate the cache from `nodeDB` and introduce a new `nodeCache` component.
+
+The modified API will have the following structure:
+
+```go
+    Tree interface {
```
ValarDragon, line 57: imo having read-only views as an interface makes sense

docs/architecture/adr-002-api-cleanup.md
ValarDragon: Why do we need the batch directly? Isn't using a generic interface, that can queue DB ops for the underlying DB immediately better?

ValarDragon:
I'd generally still suggest starting with separating out version management (#737 )

docs/architecture/adr-002-api-cleanup.md
cool-develope: Hash is used in LasCommitID like the hash of the last commit, and WorkingHash is the hash before the commit

docs/architecture/adr-002-api-cleanup.md
cool-develope: > Why not reuse the existing ChangeSet struct for the 

`ChangeSet` is great

> Why do we need the batch directly? Isn't using a generic interface, that can queue DB ops for the underlying DB immediately better?

This idea is proposed in Store V2 for atomic commit

docs/architecture/adr-002-api-cleanup.md
cool-develope: We are using the independent structs like `Exporter, Importer` in iavl

docs/architecture/adr-002-api-cleanup.md
cool-develope: I am just thinking `iavl` will be used as a SC layer and it may mislead if have a read-only interface.
Also, we have only one interface in `store` module.

docs/architecture/adr-002-api-cleanup.md
```diff
@@ -0,0 +1,98 @@
...
+        Version() int64
+        Hash() []byte
+        WorkingHash() []byte
+        VersionExists(version int64) bool
+        DeleteVersionsTo(version int64) error
+        GetVersioned(key []byte, version int64) ([]byte, *cmtprotocrypto.ProofOps, error)
+        GetTree(version int64) (*iavl.Tree, error)
+        SetInitialVersion(version uint64)
+        Iterator(start, end []byte, ascending bool) (types.Iterator, error)
+        LoadVersionForOverwriting(targetVersion int64) error
```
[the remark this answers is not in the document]
cool-develope: > I'd generally still suggest starting with separating out version management (#737 )

@tac0turtle @yihuang , how about separating out version manager?
After Saveversion with batch, we will have only version-related functions as a modifier. IMO, separated interfaces looks meaningless in the `store` module.

cool-develope:
> I'd generally still suggest starting with separating out version management (#737 )

@tac0turtle @yihuang , how about separating out version manager?
After Saveversion with batch, we will have only version-related functions as a modifier. IMO, separated interfaces looks meaningless in the `store` module.

yihuang:
> > I'd generally still suggest starting with separating out version management (#737 )
> 
> 
> 
> @tac0turtle @yihuang , how about separating out version manager?
> 
> After Saveversion with batch, we will have only version-related functions as a modifier. IMO, separated interfaces looks meaningless in the `store` module.

I'd suggest figure out the minimal changes needed to support store v2 first (async commit, atomic commit), I don't see a big problem in current API design.
The MutableTree is a singleton, so I think it's not too bad for it to also do the version manager job, I'm ok either way :)

tac0turtle:
lets separate concerns, first what is needed for store v2 then a new api. I think conflating the two could be clouding what is all needed

cool-develope:
Do we really require the `Immutable Tree` in Store V2, considering that iavl will solely be utilized for SC?
cc: @alexanderbez @tac0turtle

alexanderbez (approved):
Awesome work @cool-develope! I think this is a good jumping off point.

I can imagine, however, some of the finer points needing amendment as Store v2 work kicks off.

docs/architecture/adr-002-api-cleanup.md
```diff
@@ -0,0 +1,96 @@
...
+
+This ADR proposes a cleanup of the API to support [Store V2](https://github.com/cosmos/cosmos-sdk/blob/main/docs/architecture/adr-065-store-v2.md).
+
+There are some proposals for the speedup of the `Commit` by the async writes. See the [Discussion](https://github.com/cosmos/cosmos-sdk/issues/16173) for more details.
+
+There is a lot of legacy code in the SDK that is not used anymore and can be removed. See the [Discussion](https://github.com/cosmos/iavl/issues/737) for more details. 
+
+## Context
+
+The introduction of `Store V2` separates SS (State Storage) and SC (State Commitment), where `iavl` is used for SC and `versionDB` is already implemented for SS. The index of the `fast node` is not needed in `iavl` since `versionDB` allows key-value queries for a specific version.
```
kocubinski, line 21: Which versionDB? Do you mean the implementation in cronos on top of RocksDB using the (experimental) user defined timestamp feature?

Given that random access and iteration of keys from disk will perform best when data is stored in an indexed BTree I don’t know that we should commit to this rocks implementation.  Maybe it approximates a BTree, but either it seems beyond the scope of store interfaces to mandate what the SS backend will be.

tac0turtle: i think we should leave versiondb out of this ADR, it is orthogonal. The goal of this adr is to cleanup the api, but we are conflating two things in this adr, store v2 changes and api clean up. I think we should separate them. First api cleanup then we do the removal of fast node system

cool-develope: ok, I removed the `versionDB` from this adr scope, and made small the scope of API cleanup. I assume `nodeDB` API cleaning and creating one `Tree` interface doesn't require much work. Maybe we can open another ADR for API cleanup later.

cool-develope: The primary purpose of this ADR is to support store V2, but this will require much refactoring, I wanted to add some  API cleanup through this chance.

docs/architecture/adr-002-api-cleanup.md
tac0turtle: read only would be good as we still need the iavl tree for commitments, store v2 adr talks close to nothing about this. I wouldn't use the current adr as final its very very high level with a lack of experimentation and research

tac0turtle:
im still a bit lost with this adr, API cleanup has nothing to do with store v2. The commitment structure should be agnostic to what is going on at a higher level. I think we should remove any notion of store v2 and the plans there and work on creating a stable and sane API for this code base

cool-develope:
> im still a bit lost with this adr, API cleanup has nothing to do with store v2. The commitment structure should be agnostic to what is going on at a higher level. I think we should remove any notion of store v2 and the plans there and work on creating a stable and sane API for this code base

ok, I modified the ADR, and assumed `ImmutableTree` is still necessary.

docs/architecture/adr-002-api-cleanup.md
```diff
@@ -0,0 +1,94 @@
...
+
+### API Cleanup
+
+- Make `nodeDB` methods private and remove all unused methods.
+- Follow the naming convention of the `store` module.
+
+The modified API will have the following structure:
+
+```go
+    MutableTree interface {
```
tac0turtle, line 53: these are internal interfaces right? if these are public then i dont think this is fixing the current issue. The issue raised by dev is about the application only interacting with a single interface and making mutable and immutable tree internal workings of IAVL. 

what do you think of only having a reader and a writer interface.

cool-develope: yeah, it is an internal separation, we will have one interface externally.
do you mean rename the mutable and immutable to reader and writer? or have external interfaces?

docs/architecture/adr-002-api-cleanup.md
```diff
@@ -0,0 +1,94 @@
...
+The current implementation of `iavl` suffers from performance issues due to synchronous writes during the `Commit` process. To address this, the proposed changes aim to finalize the current version in memory, parallelize hash calculations in `WorkingHash`, and introduce asynchronous writes in the background.
+
+It is also necessary to refactor the `Set` function to accept a batch object for the atomic commitments. 
+
+Moreover, the existing architecture of `iavl` lacks modularity and code organization:
+
+- The boundary between `ImmutableTree` and `MutableTree` is not clear.
+- There are many public methods in `nodeDB`, making it less structured.
+
+## Decision
```
tac0turtle, line 30: what is the external api to users? the below seems to focus on internal changes.

cool-develope: do you think iavl needs the external API interface to clarify? I plan to keep the majority external API

cool-develope:
just removed the parallel WorkingHash part, I think it is unnecessary since we can parallelize on a module basis.
Regarding the naming convention (Reader/Writer), we can organize it on the consumer side, I'd like to keep the current structure.

docs/architecture/adr-002-api-cleanup.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Batch Set
+
+* Refactor the `Set` function to accept a batch object.
+* Eliminate the usage of `dbm.Batch` in `nodeDB`.
+
+### Async Commit
+
+* Finalize the current version in memory during `Commit`.
+* Perform async writes in the background.
```
tac0turtle, line 42: what are the side effects to this?

docs/architecture/adr-002-api-cleanup.md
```diff
@@ -0,0 +1,93 @@
...
+### API Cleanup
+
+* Make `nodeDB` methods private and remove all unused methods.
+* Follow the naming convention of the `store` module.
+
+The exposed API of `iavl` will be as follows:
+
+```go
+    MutableTree interface {
+        SaveChangeSet(cs *ChangeSet) ([]byte, error) // this is for batch set
```
tac0turtle, line 53: what is changeset here?

docs/architecture/adr-002-api-cleanup.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Positive
+
+* Atomicity of the `Commitments`.
+* Improved Commit performance through async writes.
+* Increased flexibility and ease of modification and refactoring for iavl.
+
+### Negative
+
+* Async Commit may result in increased memory usage and increased code complexity.
```
tac0turtle, line 88: i would add the added complexity in the code base. the asyncommit adds a layer of complexity

tac0turtle (commented):
if this will be iavl 2.0 we should remove fast node system.

tac0turtle (commented):
if this will be iavl 2.0 we should remove fast node system. 

After the storage call last week its unclear how much time should be spent on this repo seeing that memiavl is likely the future in the sdk. Do you have any thoughts for this?

docs/architecture/adr-002-api-cleanup.md
cool-develope: do you want to introduce some pseudo-code within this adr?

cool-develope:
> 

it is not iavl 2.0, just simple API cleanup
@kocubinski is currently working on iavl 2.0

docs/architecture/adr-002-api-cleanup.md
tac0turtle: it can be added to a con, no need for pseudo-code

tac0turtle (approved):
leaving an approval to move forward,

PR #793, opened 2023-06-21
perf: flush import batches in parallel

elias-orijtech:
As described in #664, both the construction and subsequent flushing of transaction batches are slow. This change is a minimal alternative to nothing else. This ensures simplicity and hopefully gains most of the performance improvement from #664.

On my macOS M1 machine:

goos: darwin
goarch: arm64
pkg: github.com/cosmos/iavl
BenchmarkImportBatch
BenchmarkImportBatch-8   	       4	 300864156 ns/op

running the same benchmark without this change:

BenchmarkImportBatch-8   	       3	 353229292 ns/op

CC @odeke-em @yzang2019 @tac0turtle

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,10 +7,10 @@ require (
 	github.com/cosmos/cosmos-db v1.0.0
 	github.com/cosmos/ics23/go v0.10.0
 	github.com/emicklei/dot v1.5.0
+	github.com/gogo/protobuf v1.3.2
 	github.com/golang/mock v1.6.0
 	github.com/stretchr/testify v1.8.4
-	golang.org/x/crypto v0.10.0
-	google.golang.org/protobuf v1.31.0
+	golang.org/x/crypto v0.11.0
 )
 
 require (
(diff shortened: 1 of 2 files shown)


tac0turtle:
any chance we can get a benchmark comparison of what is there now and this?

import.go
```diff
@@ -78,10 +81,21 @@ func (i *Importer) writeNode(node *Node) error {
...
+		var err error
+		if i.inflightCommit != nil {
+			err = <-i.inflightCommit
+			i.inflightCommit = nil
+		}
+		if err != nil {
 			return err
 		}
-		i.batch.Close()
+		result := make(chan error)
```
mark-rushakoff, line 93: A comment on why this is unbuffered would be helpful to clarify intent here. I would have expected it to be size 1 on my first read here, but there is probably a subtle reason why unbuffered is more appropriate.

mark-rushakoff (approved):
Marko asked me to take a look at this. It looks fine as far as I can see. Another comment or two about why the writes are happening in a separate goroutine would be helpful, but maybe those would quickly go away with work in #664.

import.go
elias-orijtech: I didn't intent anything subtle here: I use a buffered channel when I need one, otherwise not. What's the use of a buffered channel here?

elias-orijtech:
> any chance we can get a benchmark comparison of what is there now and this?

Done.

tac0turtle:
gentle ping on the benchmark for this work.

elias-orijtech:
> gentle ping on a benchmark for this work.

Added a benchmark as a larger version of the existing benchmark. See the commit https://github.com/cosmos/iavl/pull/793/commits/b9a01531e126d43d2b5920119961fe15c55af5e2 for details of before/after.

tac0turtle:
oh sorry missed that thank you

tac0turtle:

yzang2019:
Sei has done some benchmark and with a extremely large snapshot, we are able to reduce the import time by 2X with a large batch size and parallel flush

Issue #696, opened 2023-03-02
Racing conditions on `UnsavedFastIterator` & `MutableTree`


GAtom22:
Detected an issue with concurrent map operations to the `unsavedFastNodeAdditions` & `unsavedFastNodeRemovals` maps on the [UnsavedFastIterator](https://github.com/cosmos/iavl/blob/d2dd845c2b1d04517341acfe6e1546a51baf75a7/unsaved_fast_iterator.go#L19-L36) and [MutableTree](https://github.com/cosmos/iavl/blob/d2dd845c2b1d04517341acfe6e1546a51baf75a7/mutable_tree.go#L35-L44) structs.

I tried adding a mutex and use the lock when these operations happen and seems to solve the issue.

More info here https://github.com/cosmos/cosmos-sdk/issues/10568#issuecomment-1451875518

Error with the stack trace:

```
[90m12:54PM[0m [32mINF[0m executed block [36mheight=[0m9522889 [36mmodule=[0mstate [36mnum_invalid_txs=[0m0 [36mnum_valid_txs=[0m3 [36mserver=[0mnode
fatal error: concurrent map iteration and map write

goroutine 963386 [running]:
runtime.throw({0x27ef174?, 0x8?})
	runtime/panic.go:992 +0x71 fp=0xc0a2e49fb0 sp=0xc0a2e49f80 pc=0x43f0b1
runtime.mapiternext(0xc0a2e4a0a0?)
	runtime/map.go:871 +0x4eb fp=0xc0a2e4a020 sp=0xc0a2e49fb0 pc=0x41698b
github.com/cosmos/iavl.NewUnsavedFastIterator({0xc09c5e3520, 0x1e, 0x20}, {0xc09c5e3540, 0x1e, 0x20}, 0x1, 0xc0012f2880, 0xc07132da40, 0xc07132da70)
	github.com/cosmos/iavl@v0.19.4/unsaved_fast_iterator.go:55 +0x236 fp=0xc0a2e4a110 sp=0xc0a2e4a020 pc=0xe6f2d6
github.com/cosmos/iavl.(*MutableTree).Iterator(0xc0018040f0, {0xc09c5e3520, 0x1e, 0x20}, {0xc09c5e3540, 0x1e, 0x20}, 0xe0?)
	github.com/cosmos/iavl@v0.19.4/mutable_tree.go:216 +0x10c fp=0xc0a2e4a178 sp=0xc0a2e4a110 pc=0xe55f2c
[...]

PR #802, opened 2023-07-31
Support concurrency for IAVL and fix Racing conditions

mattverse:
cref: https://github.com/cosmos/iavl/issues/696

This PR fixes the problem stated above by replacing maps that causes racing conditions to sync map to be thread safe. 

Confirmed and tested on a node with the following procedure:
- Have node running 
- Spam it with tx bots and query bots
- Previously node would crash within 1, maximum 2 minutes with the racing condition logs mentioned above
- With fix applied node would not crash for 40+ minutes. 

After 40+ minutes, node does seem to crash, but suspecting this may be to different concurrency issues (still doing more investigation)

### DIFF
--- a/iterator_test.go
+++ b/iterator_test.go
@@ -3,6 +3,7 @@ package iavl
 import (
 	"math/rand"
 	"sort"
+	"sync"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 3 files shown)


julienrbrt:
I think this should target master instead of the release branch, this issue is probably present on v0.20 as well and v1.0.0.

mattverse:
@julienrbrt We can also backport this to older release versions as well though right?

julienrbrt:
> @julienrbrt We can also backport this to older release versions as well though right?

Right!

mattverse:
@julienrbrt Awesome, let's do this, I'll close this issue for now and create a PR pointing at main. After we have that reviewed we can probably backport this to all necessary branches

julienrbrt:
Awesome, let's do that!

PR #807, opened 2023-08-08
feat: Uses BatchWithFlusher in iavl tree

catShaark:
This PR replace the use of normal dbm.Batch with BatchWithFlusher in iavl tree
It also hard code the value of flushthreshold to 100KB

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -8,22 +8,20 @@ import (
 // around batch that flushes batch's data to disk
 // as soon as the configurable limit is reached.
 type BatchWithFlusher struct {
-	db             dbm.DB    // This is only used to create new batch
-	batch          dbm.Batch // Batched writing buffer.
-	flushThreshold int       // The maximum size of the batch in bytes before it gets flushed to disk
+	db    dbm.DB    // This is only used to create new batch
+	batch dbm.Batch // Batched writing buffer.
 }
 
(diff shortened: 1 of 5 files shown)


tac0turtle (approved):
LGTM thank you

cool-develope (approved):
overall looks good to me, except why hard code the flushThreshold?

catShaark:
That's because I discussed with marko and we decide to hard code the value to be the best option showed by benchmark result run on osmosis node

julienrbrt:
+1 on @cool-develope comment.

Given that `newNodeDB` takes `Options` we could add an option in here: https://github.com/notional-labs/iavl/blob/a945f5f903800a214d30e55f87ca96cc6bc78051/options.go#L71-L89 and use `100000` in the DefaultOptions. If going this way, we should make them variadic so you don't end up overwriting that value to 0 by defining new options.

tac0turtle:
@julienrbrt you fine with merging this and then i can make the changes.

julienrbrt:
> @julienrbrt you fine with merging this and then i can make the changes.

Works! We need changelogs as well to reflect some API breaks from here in the follow-up

julienrbrt:
@mergifyio backport release/1.x.x

julienrbrt:
@Mergifyio backport release/v1.x.x

tac0turtle:

Issue #819, opened 2023-08-17
Async Commit


cool-develope:
The proposal, detailed in [this document](https://github.com/cosmos/iavl/blob/master/docs/architecture/adr-002-api-cleanup.md), suggests writing new nodes asynchronously within the `SaveVersion` function. This approach aims to enhance the speed of finalization during the `Commit` stage.

PR #826, opened 2023-08-29
feat: async commit

cool-develope:
closes: #819

- Refactored writing nodes and fastnodes asynchronously
- Removed the `commitGap` since we have a `flusherBatch`

### DIFF
--- a/basic_test.go
+++ b/basic_test.go
@@ -448,6 +448,7 @@ func TestPersistence(t *testing.T) {
 	t1.SaveVersion()
 
 	// Load a tree
+	t1.ndb.waitAsyncWrite()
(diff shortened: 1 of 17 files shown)


tac0turtle:
Can we delete without waiting to finish the write? 
Do snapshots also work from the state within memory or there is a need to wait for the version to be written to disk?

cool-develope:
> Can we delete without waiting to finish the write? Do snapshots also work from the state within memory or there is a need to wait for the version to be written to disk?

yeah, it is possible, it will require more detailed management of the write routine.  I will try it after fixing the test failures.

tac0turtle:
closing this based off recent discussions on iavlv2

PR #843, opened 2023-10-13
fix(nodedb): prevent deadlock by releasing DeleteVersionsFrom mutex on error

### LINKED ISSUE #842: nodeDb.DeleteVersionsFrom can leave mutex locked indefinitely hence a deadlock
Noticed in an audit and while preparing a test for P #838 that this code exists https://github.com/cosmos/iavl/blob/3574b99d1cb0f06591c2bd02cba47bef7ebc9312/nodedb.go#L493-L499

but in case the erroring condition matches, it'll be locked forever. Let's fix that by 2 ways
- [ ] A dedicated closure with an error return
```go
err := func() error {
	ndb.mtx.Lock()
	defer ndb.mtx.Unlock()

	for v, r := range ndb.versionReaders {
		if v >= fromVersion && r != 0 {
			return fmt.Errorf("unable to delete version %v with %v active readers", v, r)
		}
	}
         return nil
}()
```

- [ ] Manually unlocking in the error path
```go
	ndb.mtx.Lock()
	for v, r := range ndb.versionReaders {
		if v >= fromVersion && r != 0 {
                         ndb.mtx.Unlock()
			return fmt.Errorf("unable to delete version %v with %v active readers", v, r)
		}
	}
	ndb.mtx.Unlock()
```

/cc @elias-orijtech

odeke-em:
Noticed during an audit that there can be a deadlock resulting from an unreleased mutex if there happens to be a newer version with one or more concurrent readers.

Fixes #842

### DIFF
--- a/nodedb.go
+++ b/nodedb.go
@@ -493,6 +493,7 @@ func (ndb *nodeDB) DeleteVersionsFrom(fromVersion int64) error {
 	ndb.mtx.Lock()
 	for v, r := range ndb.versionReaders {
 		if v >= fromVersion && r != 0 {
+			ndb.mtx.Unlock() // Unlock before exiting
 			return fmt.Errorf("unable to delete version %v with %v active readers", v, r)
 		}
 	}
--- a/nodedb_test.go
+++ b/nodedb_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strconv"
 	"testing"
+	"time"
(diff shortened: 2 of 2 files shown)


tac0turtle (approved):

testinginprod (approved):
Change is reasonable to me!

tac0turtle:

Issue #852, opened 2023-11-15
Post-order snapshot


cool-develope:
IAVL v1 and v0 export is based on post-order, implement post-order export for compatibility with IAVL v0 and v1

tac0turtle:
should we only do restore for post order that way its less code to maintain over the long term?

yihuang:
https://github.com/cosmos/iavl/pull/703

This compact format rely on post-order BTW.

tac0turtle:
we should evaluate which is overall better (compact and performant). I dont think we should support both in iavlv2, only one will get used in reality. This can come later

cool-develope:
with #853

PR #853, opened 2023-11-16
feat: add post-order export & restoring

cool-develope:
ref: #852

### DIFF
--- a/snapshot.go
+++ b/snapshot.go
@@ -315,12 +315,18 @@ type SnapshotNode struct {
 	Height  int8
 }
 
-func (sql *SqliteDb) ImportSnapshotFromTable(version int64, loadLeaves bool) (*Node, error) {
+func (sql *SqliteDb) ImportSnapshotFromTable(version int64, traverseOrder TraverseOrderType, loadLeaves bool) (*Node, error) {
 	read, err := sql.getReadConn()
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


tree_test.go
```diff
@@ -205,13 +204,14 @@ func TestTree_Build_Load(t *testing.T) {
 
 	// export the tree at version 12,000 and import it into a sql db
 	ctx := context.Background()
+	traverseOrder := PreOrder
```
yihuang, line 207: Can you add test for post-order traversal.

yihuang (commented):
Implementation looks good to me, but I think we need to add some tests for it.

tree_test.go
kocubinski: Yea, let's have this test run for both pre and post order.

snapshot.go
```diff
@@ -592,6 +514,165 @@ func (snap *sqliteSnapshot) prepareWrite() error {
...
+
+		return node, nil
+	}
+
+	node, err := step()
+
+	return node, uniqueVersions, err
+}
+
+func (snap *sqliteSnapshot) writeSnapNode(node *Node, version int64, count int) error {
```
kocubinski, line 644: I think the write logic may need to change a bit here now, if you pull `v2-alpha6` you'll see what I mean.  I needed to incorporate an additional flag to accommodate multiple write paths in snapshot.

https://github.com/cosmos/iavl/blob/640a21ec34b532d04886b6de2085c809dca6f343/snapshot.go#L30C1-L31

Issue #862, opened 2023-12-22
Refactor the pruning logic in V1


cool-develope:
## Context

V1 is pruning the orphan nodes one time and it takes much time. To figure out it, it will keep the orphan nodes in default and will expose the API to prune the orphan nodes manually.

yihuang:
Is it possible to do pruning through compaction filter, that'd be most efficient.

PR #863, opened 2023-12-22
feat: remove legacy orphans background

cool-develope:
Closes: #862 

## Context

- refactor removing the legacy orphans in background.
- add the mutex in `BatchWithFlusher` to avoid the race condition

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -1,6 +1,8 @@
 package iavl
 
 import (
+	"sync"
+
 	dbm "github.com/cosmos/cosmos-db"
 )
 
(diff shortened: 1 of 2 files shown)


ValarDragon:
Is this concurrency safe with main thread commits? (This can add to deleting while the main thread is committing right?)

cool-develope:
> Is this concurrency safe with main thread commits? (This can add to deleting while the main thread is committing right?)
> 
> Do we need a mutex to ensure the two don't try to add to batch at the same time?

yeah, the test is failed with data racing

ValarDragon:
This is the wrong spot to put it, we hacked something that worked in this branch https://github.com/cosmos/iavl/tree/dev/v1.0.0-orphan

(the first part of the loop is what was expensive)

I think the way you were doing the code is cleaner, just that was something that I got to work. Plz feel free to takeover/move anything that helps :pray:

cool-develope:
close in favor of #877

PR #864, opened 2023-12-23
Speedup key formatting

ValarDragon:
This speeds up key formatting, which is by no means the bottleneck, but appeared to be 0.5% of the time of some Osmosis mainnet sync pprof's on IAVL v0 and v1

### DIFF
--- a/node.go
+++ b/node.go
@@ -57,11 +57,15 @@ func GetRootKey(version int64) []byte {
 
 // Node represents a node in a Tree.
 type Node struct {
-	key           []byte
-	value         []byte
-	hash          []byte
-	nodeKey       *NodeKey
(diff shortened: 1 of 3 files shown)


node.go
```diff
@@ -57,11 +57,15 @@ func GetRootKey(version int64) []byte {
...
-	leftNodeKey   []byte
+	key     []byte
+	value   []byte
+	hash    []byte
+	nodeKey *NodeKey
+	// Legacy: LeftNodeHash
+	// v1: Left node ptr via Version/key
+	leftNodeKey []byte
+	// Legacy: RightNodeHash
+	// v1: Right node ptr via Version/key
```
ValarDragon, line 68: Driveby comment addition

PR #865, opened 2023-12-24
perf: Avoid making an extra heap copy in DecodeBytes

ValarDragon:
Deserializing bytes right now is making a new heap copy of the slice. In IAVL v1, with legacy nodes, over a 100 block period this appeared to be `0.3%` of the CPU time. (Plus heap / memory pressure)

This is only needed if the input bytes are mutated, but this is not the case in any usage of this, as each read is coming from a new `GetNode` or cache result. GetNode returns new heap-data for the value, and the cache entries do not mutate. Hence we can save on CPU time + heap pressure with this.

### DIFF
--- a/fastnode/fast_node.go
+++ b/fastnode/fast_node.go
@@ -30,6 +30,7 @@ func NewNode(key []byte, value []byte, version int64) *Node {
 }
 
 // DeserializeNode constructs an *FastNode from an encoded byte slice.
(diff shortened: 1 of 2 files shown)


tac0turtle (approved):

cool-develope:
@ValarDragon could we close it since it is included in #877?

ValarDragon:
Sure!

PR #872, opened 2024-01-18
feat: iavl v2

kocubinski:
See also: https://hackmd.io/rXq7coSyTFmhWIa4mifBYQ

For a complete end to end test, this version of code is running on an osmosis node syncing on mainnet, some metrics are viewable at [this Grafana](http://osmosis-iavl-v2.kocubinski.com:3000/d/fa4937cf-92ec-47b8-90c9-fc3d32957ebc/osmosis-bench?orgId=1&refresh=5s&from=now-30m&to=now).  Sync started from block `13946655` after an export of `application.db` to an iavl v2 SQLite database with the code in `./migrate` in this PR.

Sync rates as fast as ~10 blocks per second are observed, most of the spikiness in commit rate is spent waiting on blocks from consensus, but I don't have a way to show this in telemetry.  AFAIK this is about 2x from iavl v1 with osmosis,  I don't see evidence that the osmosis build is I/O bound on disk either as benchmarks put max throughput 10-20x iavl v1.

This configuration is using 45G of disk storage, down from 177G.
Checkpoints are taken every 5k blocks, with a WAL maintained on each commit.
Pruning occurs every 900 blocks with the last 5k blocks kept, therefore the last checkpoint always contains a fully materialized tree.


tac0turtle:
is this ready for review?

v2/export.go
```diff
@@ -0,0 +1,101 @@
...
+	PostOrder
+)
+
+type Exporter struct {
+	tree  *Tree
+	out   chan *Node
+	errCh chan error
+}
+
+func (tree *Tree) Export(order TraverseOrderType) *Exporter {
```
tac0turtle, line 19: go docs

v2/internal/encoding.go
```diff
@@ -0,0 +1,167 @@
...
+	},
+}
+
+var uvarintPool = &sync.Pool{
+	New: func() interface{} {
+		return &[binary.MaxVarintLen64]byte{}
+	},
+}
+
+// decodeBytes decodes a varint length-prefixed byte slice, returning it along with the number
```
tac0turtle, line 31:
```suggestion
// DecodeBytes decodes a varint length-prefixed byte slice, returning it along with the number
```

cool-develope:
Rollback feature is missed (DeleteVersionsFrom in v1). 
@kocubinski, I think MultiTree has the potential advantages in terms of deferred pruning and also would be easy to manage the flow.

cool-develope:
how do we test v2 ?

v2/cmd/gen/gen.go
```diff
@@ -0,0 +1,234 @@
...
+		limit int
+	)
+	cmd := &cobra.Command{
+		Use:   "emit",
+		Short: "emit generated changesets to disk",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			itr, err := getChangesetIterator(typ)
+			if err != nil {
+				return err
+			}
```
odeke-em, line 63: Missing defer iter.Close()

v2/cmd/gen/gen.go
```diff
@@ -0,0 +1,234 @@
...
+
+			stream := compact.StreamingContext{
+				In:          make(chan compact.Sequenced),
+				Context:     ctx,
+				OutDir:      out,
+				MaxFileSize: 100 * 1024 * 1024,
+			}
+
+			var wg sync.WaitGroup
+			wg.Add(1)
```
odeke-em, line 74: Please add a defer to close the channel as well as wait on the wait group per

```go
defer func() {
  close(ch)
  wg.Wait()
}()
```

v2/cmd/gen/gen.go
```diff
@@ -0,0 +1,234 @@
...
+			go func() {
+				stats, err := stream.Compact()
+				if err != nil {
+					log.Fatal().Err(err).Msg("failed to compact")
```
odeke-em, line 78: It's not recommended to invoke log.Fatal in a go routine, it doesn't give others time to cleanup.

v2/cmd/gen/gen.go
```diff
@@ -0,0 +1,234 @@
...
+			if err != nil {
+				return err
+			}
+
+			var i int64
+			var lastHash []byte
+			var lastVersion int64
+			start := time.Now()
+			for ; itr.Valid(); err = itr.Next() {
+				if err != nil {
```
odeke-em, line 170: We need to catch the end of the iteration, but this code just returns on any generic error

v2/cmd/gen/gen.go
```diff
@@ -0,0 +1,234 @@
...
+					return err
+				}
+				if limit > -1 && itr.Version() > limit {
+					break
+				}
+
+				changeset := itr.Nodes()
+				for ; changeset.Valid(); err = changeset.Next() {
```
odeke-em, line 178: How does this code differentiate between the end of the iterator and another error? I don't see any special casing to break out of the loop

v2/cmd/latest.go
```diff
@@ -0,0 +1,70 @@
...
+				errors = make(chan error)
+				cnt    = 0
+			)
+			for _, path := range paths {
+				cnt++
+				sqlOpts := iavl.SqliteDbOptions{Path: path}
+				sql, err := iavl.NewSqliteDb(pool, sqlOpts)
+				if err != nil {
+					return err
+				}
```
odeke-em, line 33: Missing sql.Close() on ending the function.

v2/cmd/scan/scan.go
```diff
@@ -0,0 +1,103 @@
...
+
+func probeCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "probe",
+		Short: "prob sqlite cgo configuration",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			f, err := os.CreateTemp("", "iavl-v2-probe.sqlite")
+			if err != nil {
+				return err
+			}
```
odeke-em, line 28: Missing defer f.Close()

v2/cmd/scan/scan.go
```diff
@@ -0,0 +1,103 @@
...
+			fmt.Println("mmap:", res)
+
+			if err = stmt.Close(); err != nil {
+				return err
+			}
+			if err = conn.Close(); err != nil {
+				return err
+			}
+			if err = os.Remove(f.Name()); err != nil {
```
odeke-em, line 68: Without f.Close() this code is undeterministic and might not have flushed data but even more on some systems, trying to delete an open file can be refused by the security policy.

odeke-em: Also you can simply invoke a return instead of special casing err != nil, vs nil error

v2/cmd/snapshot/snapshot.go
```diff
@@ -0,0 +1,79 @@
...
+			var (
+				pool   = iavl.NewNodePool()
+				done   = make(chan struct{})
+				errors = make(chan error)
+				cnt    = 0
+			)
+			for _, path := range paths {
+				cnt++
+				sqlOpts := iavl.SqliteDbOptions{Path: path}
+				sql, err := iavl.NewSqliteDb(pool, sqlOpts)
```
odeke-em, line 42: This handle is never closed on unrelated errors. Please close it after use perhaps with a defer, or in each error flow clause.

v2/export.go
```diff
@@ -0,0 +1,101 @@
...
+	exporter := &Exporter{
+		tree:  tree,
+		out:   make(chan *Node),
```
odeke-em, line 22: To speed this code up, please make these buffered channels with some capacity by means of a heuristic, so that traversal won't be constrained and blocked by the speed of reading out the nodes

odeke-em (commented):
Thank you @kocubinski, I just saw this change and here is a first pass review.

tac0turtle (approved):

tac0turtle:
gonna merge this as john is working iavlv2 cleann

PR #877, opened 2024-01-31
feat(master): async pruning of orphan nodes 

czarcas7ic:
This feature prevents the multiple hour long waiting period when chains upgrade from previous versions of IAVL to IAVL v1. The time comes from pruning orphan nodes. This synchronously prunes them. This branch has been NOT been tested against osmosis mainnet, however the v1.x.x version of this PR has.


cool-develope (approved):

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - [#726](https://github.com/cosmos/iavl/pull/726) Make `KVPair` and `ChangeSet` serializable with protobuf.
 - [#718](https://github.com/cosmos/iavl/pull/718) Fix `traverseNodes` unexpected behaviour
 - [#770](https://github.com/cosmos/iavl/pull/770) Add `WorkingVersion()int64` API.
+- [#877](https://github.com/cosmos/iavl/pull/877) Make pruning of orphan nodes synchronous.
```
cool-develope, line 13:
```suggestion
- [#877](https://github.com/cosmos/iavl/pull/877) Make pruning of legacy orphan nodes asynchronous.
```

batch.go
```diff
@@ -46,6 +49,9 @@ func (b *BatchWithFlusher) estimateSizeAfterSetting(key []byte, value []byte) (i
 // the batch is flushed to disk, cleared, and a new one is created with buffer pre-allocated to threshold.
 // The addition entry is then added to the batch.
 func (b *BatchWithFlusher) Set(key, value []byte) error {
+	b.mtx.Lock()
```
kocubinski, line 52: If each batch accumulates in memory before flushing to disk doesn't it make sense to lock at write instead of Set/Delete?

nodedb.go
```diff
@@ -406,60 +407,124 @@ func (ndb *nodeDB) deleteLegacyNodes(version int64, nk []byte) error {
...
+		counter := 0
+		for ; itr.Valid(); itr.Next() {
+			legacyRootKeyFormat.Scan(itr.Key(), &curVersion)
+			rootKeys = append(rootKeys, itr.Key())
+			if prevVersion > 0 {
+				if err := ndb.traverseOrphans(prevVersion, curVersion, func(orphan *Node) error {
+					counter++
+					if counter == 1000 {
+						counter = 0
+						time.Sleep(1000 * time.Millisecond)
```
kocubinski, line 452: Is this sleep to create a write gap so that `SaveVersion` doesn't block too frequently from the newly introduced mutex in `BatchWithFlusher` while a long prune operation is ongoing?  

If `deleteLegacyVersions` is only called once on migration I guess it's OK, but this sleep is hard to understand and could result in unexpected wait times from the main I/o thread.  I'm also curious about the overhead of mutex/set release on every `Set/Delete` call.

nodedb.go
```diff
@@ -406,60 +407,124 @@ func (ndb *nodeDB) deleteLegacyNodes(version int64, nk []byte) error {
...
+
+		// Delete orphans for all legacy versions
+		var prevVersion, curVersion int64
+		var rootKeys [][]byte
+		counter := 0
+		for ; itr.Valid(); itr.Next() {
+			legacyRootKeyFormat.Scan(itr.Key(), &curVersion)
+			rootKeys = append(rootKeys, itr.Key())
+			if prevVersion > 0 {
+				if err := ndb.traverseOrphans(prevVersion, curVersion, func(orphan *Node) error {
```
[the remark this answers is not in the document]
cool-develope: yes, it can simple iterate orphan values

cool-develope:
@czarcas7ic could you please wire up this PR? only encoding and fastKeyFormatter features

czarcas7ic:
@cool-develope Just verifying, you only want the FastPrefixFormatter change as well as encoding.go change, and remove everything else? My apologies, I am between multiple tasks right now so am losing context on this.

czarcas7ic:
Thanks, my apologies for the extra overhead that this may have caused!

PR #878, opened 2024-02-04
CV Optimistic Export and Import

chillyvee:
Optimistic Export and Import aims to reduce p2p statesync recovery times.

Validators attempting to recover nodes for cosmos-sdk chains with significant state size may be slashed for downtime prior to recovery.

Although all data is verified during tree build, rebuilding the entire IAVL tree using minimal data incurs high CPU overhead.

Evmos for example takes over 16 hours to rebuild IAVL on a "fast" CPU/DISK machine and can appear to "never finish".  The node must then fast-sync 16 hours of missed blocks which can take an additional 4-5 hours.  Slashing for downtime is within 30 hours meaning a validator must succeed on the first try or be slashed.

Since statesync appears to "hang", validators download a 244GB file from Polkachu to restore data directories.  This is possible because Polkachu is a trusted party.  Validators do not appear to validate the correctness of the IAVL tree so this method is less trusted but necessary. The problem is when Polkachu only makes 1 snapshot every 24 hours.  A validator may need to fast-sync 23 hours of lost block, wait for the next Polkachu snapshot, or try another random less known snapshot.

Optimistic statesync enables faster database import at the restoring node by directly inserting the final key/value pairs into the database.  As long as validators statesync from a trusted node, this can be a good alternative to downloading data directory zip files.

Cosmos-sdk should broadcast a different snapshot format so that nodes can select optimistic statesync node sources.

There are some code locations below where there may be more efficient methods to obtain the raw key/value storage than recalculation.

Comments welcome.

### DIFF
--- a/export.go
+++ b/export.go
@@ -78,7 +78,7 @@ func newExporterWithOptions(tree *ImmutableTree, optimistic bool) (*Exporter, er
 
 // export exports nodes
 func (e *Exporter) export(ctx context.Context) {
-	e.tree.root.traversePost(e.tree, true, func(node *Node) bool {
+	_ = e.tree.root.traversePost(e.tree, true, func(node *Node) bool {
 		exportNode := &ExportNode{
 			Key:     node.key,
 			Value:   node.value,
--- a/export.go
+++ b/export.go
@@ -1,6 +1,7 @@
 package iavl
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -31,13 +32,24 @@ type ExportNode struct {
 // depth-first post-order (LRN), this order must be preserved when importing in order to recreate
 // the same tree structure.
 type Exporter struct {
-	tree   *ImmutableTree
-	ch     chan *ExportNode
-	cancel context.CancelFunc
+	tree       *ImmutableTree
+	ch         chan *ExportNode
+	cancel     context.CancelFunc
+	optimistic bool // export raw key value pairs for optimistic import
 }
 
 // NewExporter creates a new Exporter. Callers must call Close() when done.
 func newExporter(tree *ImmutableTree) (*Exporter, error) {
+	return newExporterWithOptions(tree, false)
+}
+
+// NewOptimisticExporter creates a new Exporter with raw Key Values. Callers must call Close() when done.
+func newOptimisticExporter(tree *ImmutableTree) (*Exporter, error) {
+	return newExporterWithOptions(tree, true)
+}
+
+// NewExporterWithOptions creates a new Exporter and configures optimistic mode
+func newExporterWithOptions(tree *ImmutableTree, optimistic bool) (*Exporter, error) {
 	if tree == nil {
 		return nil, fmt.Errorf("tree is nil: %w", ErrNotInitalizedTree)
 	}
@@ -48,13 +60,18 @@ func newExporter(tree *ImmutableTree) (*Exporter, error) {
 
 	ctx, cancel := context.WithCancel(context.Background())
 	exporter := &Exporter{
-		tree:   tree,
-		ch:     make(chan *ExportNode, exportBufferSize),
-		cancel: cancel,
+		tree:       tree,
+		ch:         make(chan *ExportNode, exportBufferSize),
+		cancel:     cancel,
+		optimistic: optimistic,
 	}
 
 	tree.ndb.incrVersionReaders(tree.version)
-	go exporter.export(ctx)
+	if exporter.optimistic {
+		go exporter.optimisticExport(ctx)
+	} else {
+		go exporter.export(ctx)
+	}
 
 	return exporter, nil
 }
@@ -67,6 +84,40 @@ func (e *Exporter) export(ctx context.Context) {
 			Value:   node.value,
 			Version: node.nodeKey.version,
 			Height:  node.subtreeHeight,
+		}
+
+		select {
+		case e.ch <- exportNode:
+			return false
+		case <-ctx.Done():
+			return true
+		}
+	})
+	close(e.ch)
+}
+
+// optimisticExport exports raw key, value nodes
+// Cosmos-SDK should set different snapshot format so nodes can select between either "untrusted statesync" or "trusted-peer optimistic" import
+func (e *Exporter) optimisticExport(ctx context.Context) {
+	e.tree.root.traverse(e.tree, true, func(node *Node) bool {
+		// TODO: How to get the original db value bytes directly without writeBytes()?
+		buf := bufPool.Get().(*bytes.Buffer)
+		buf.Reset()
+		defer bufPool.Put(buf)
+
+		if err := node.writeBytes(buf); err != nil {
+			fmt.Printf("WARN: failed writeBytes")
+		}
+
+		bytesCopy := make([]byte, buf.Len())
+		copy(bytesCopy, buf.Bytes())
+
+		// Use Export Node Format.
+		exportNode := &ExportNode{
+			Key:     node.GetKey(), // TODO: How to get prefixed key so that import does not need to prefix?
+			Value:   bytesCopy,
+			Version: 0, // Version not used
+			Height:  0, // Height not used
 		}
 
 		select {
--- a/export_test.go
+++ b/export_test.go
@@ -298,6 +298,62 @@ func TestExporter_Import(t *testing.T) {
 	}
 }
 
+func TestOptimisticExporter_Import(t *testing.T) {
+	testcases := map[string]*ImmutableTree{
+		"empty tree": NewImmutableTree(dbm.NewMemDB(), 0, false, log.NewNopLogger()),
+		"basic tree": setupExportTreeBasic(t),
+	}
+	if !testing.Short() {
+		testcases["sized tree"] = setupExportTreeSized(t, 4096)
+		testcases["random tree"] = setupExportTreeRandom(t)
+	}
+
+	for desc, tree := range testcases {
+		tree := tree
+		t.Run(desc, func(t *testing.T) {
+			t.Parallel()
+
+			exporter, err := tree.OptimisticExport()
+			require.NoError(t, err)
+			defer exporter.Close()
+
+			newTree := NewMutableTree(dbm.NewMemDB(), 0, false, log.NewNopLogger())
+			importer, err := newTree.OptimisticImport(tree.Version())
+			require.NoError(t, err)
+			defer importer.Close()
+
+			for {
+				item, err := exporter.Next()
+				if err == ErrorExportDone {
+					err = importer.Commit()
+					require.NoError(t, err)
+					break
+				}
+				require.NoError(t, err)
+				err = importer.Add(item)
+				require.NoError(t, err)
+			}
+
+			treeHash := tree.Hash()
+			newTreeHash := newTree.Hash()
+
+			require.Equal(t, treeHash, newTreeHash, "Tree hash mismatch")
+			require.Equal(t, tree.Size(), newTree.Size(), "Tree size mismatch")
+			require.Equal(t, tree.Version(), newTree.Version(), "Tree version mismatch")
+
+			tree.Iterate(func(key, value []byte) bool { //nolint:errcheck
+				index, _, err := tree.GetWithIndex(key)
+				require.NoError(t, err)
+				newIndex, newValue, err := newTree.GetWithIndex(key)
+				require.NoError(t, err)
+				require.Equal(t, index, newIndex, "Index mismatch for key %v", key)
+				require.Equal(t, value, newValue, "Value mismatch for key %v", key)
+				return false
+			})
+		})
+	}
+}
+
 func TestExporter_Close(t *testing.T) {
 	tree := setupExportTreeSized(t, 4096)
 	exporter, err := tree.Export()
--- a/immutable_tree.go
+++ b/immutable_tree.go
@@ -160,6 +160,12 @@ func (t *ImmutableTree) Export() (*Exporter, error) {
 	return newExporter(t)
 }
 
+// OptimisiticExport returns an iterator that exports tree nodes as ExportNodes. These nodes can be
+// imported with MutableTree.Import() to recreate an identical tree.
+func (t *ImmutableTree) OptimisticExport() (*Exporter, error) {
+	return newOptimisticExporter(t)
+}
+
 // GetWithIndex returns the index and value of the specified key if it exists, or nil and the next index
 // otherwise. The returned value must not be modified, since it may point to data stored within
 // IAVL.
--- a/import.go
+++ b/import.go
@@ -178,7 +178,9 @@ func (i *Importer) OptimisticAdd(exportNode *ExportNode) error {
 	}
 	i.batchSize++
 
-	i.sendBatchIfFull()
+	if err := i.sendBatchIfFull(); err != nil {
+		return errors.New("failed sending db write batch")
+	}
 
 	return nil
 }
@@ -263,9 +265,8 @@ func (i *Importer) Commit() error {
 		return ErrNoImport
 	}
 
-	if i.optimistic {
-		// All keys should be already imported
-	} else {
+	// Optimistic: All keys should be already imported
+	if !i.optimistic {
 		switch len(i.stack) {
 		case 0:
 			if err := i.batch.Set(i.tree.ndb.nodeKey(GetRootKey(i.version)), []byte{}); err != nil {
--- a/import.go
+++ b/import.go
@@ -31,6 +31,9 @@ type Importer struct {
 
 	// inflightCommit tracks a batch commit, if any.
 	inflightCommit <-chan error
+
+	// Optimistic raw key value import
+	optimistic bool
 }
 
 // newImporter creates a new Importer for an empty MutableTree.
@@ -38,6 +41,17 @@ type Importer struct {
 // version should correspond to the version that was initially exported. It must be greater than
 // or equal to the highest ExportNode version number given.
 func newImporter(tree *MutableTree, version int64) (*Importer, error) {
+	return newImporterWithOptions(tree, version, false)
+}
+
+// newOptimisticImporter creates a new Importer for an empty MutableTree.
+//
+// expects optimistic raw key values for import
+func newOptimisticImporter(tree *MutableTree, version int64) (*Importer, error) {
+	return newImporterWithOptions(tree, version, true)
+}
+
+func newImporterWithOptions(tree *MutableTree, version int64, optimistic bool) (*Importer, error) {
 	if version < 0 {
 		return nil, errors.New("imported version cannot be negative")
 	}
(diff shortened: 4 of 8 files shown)


yihuang:
Do you use memiavl in evmos, that should only take dozens of minutes to restore.

chillyvee:
> Do you use memiavl in evmos, that should only take dozens of minutes to restore.

Will try it.  Thank you.

What do you think of the concept in this PR?

yihuang:
> Optimistic statesync enables faster database import at the restoring node by directly inserting the final key/value pairs into the database

what do you mean by "final key/value pairs", is that leaf key values only without branch nodes?

chillyvee:
> what do you mean by "final key/value pairs", do you mean only handle leaf key values skipping branch nodes?

Usually for the restoring node to write that to disk, the hash of all children must be computed before parents can be written to disk.

However it is unnecessary to rehash all data and rebuild that IAVL tree.  The receiver can write down the result of all the work.

Instead of
```
db.Set(recursive hashAllChildenAndBuildIAVLTree, value)
```

Assuming the source is trusted, It is faster to send the resulting key and value

Sender:
```
resultingKey = hash("k")
resultingValue = buffer.Write(....)
send(resultingKey, resultingValue)
```

Receiver:
```
db.Set(resultingKey, resultingValue) <- no tree build or hash operations
```

yihuang:
> > what do you mean by "final key/value pairs", do you mean only handle leaf key values skipping branch nodes?
> 
> Usually for the restoring node to write that to disk, the hash of all children must be computed before parents can be written to disk.
> 
> However it is unnecessary to rehash all data and rebuild that IAVL tree. The receiver can write down the result of all the work.
> 
> Instead of
> 
> ```
> db.Set(recursive hashAllChildenAndBuildIAVLTree, value)
> ```
> 
> Assuming the source is trusted, It is faster to send the resulting key and value
> 
> Sender:
> 
> ```
> resultingKey = hash("k")
> resultingValue = buffer.Write(....)
> send(resultingKey, resultingValue)
> ```
> 
> Receiver:
> 
> ```
> db.Set(resultingKey, resultingValue) <- no tree build or hash operations
> ```
> 
> Then the node will start as if all key/values were in the database for the current version/blockchain height.

I see, so the data written into `application.db` is the same right? but I guess the db write is the slowest part, for example memiavl is much faster despite it also rebuild the tree from current state sync snapshot format.

chillyvee:
> I see, so the data written into `application.db` is the same right? 

Yes same result to goleveldb (or other disk based iavl)

> but I guess the db write is the slowest part, for example memiavl is much faster despite it also rebuild the iavl tree from current state sync snapshot format.

Yes, memiavl as a solution when RAM is sufficient.

And this solution when disk write speeds are acceptable (with no other change in expected RAM) and utilize more disk based strategy.

yihuang:
> > I see, so the data written into `application.db` is the same right?
> 
> Yes same result to goleveldb (or other disk based iavl)
> 
> > but I guess the db write is the slowest part, for example memiavl is much faster despite it also rebuild the iavl tree from current state sync snapshot format.
> 
> Yes, memiavl as a solution when RAM is sufficient.

memiavl doesn't need much memory actually in our practice.

chillyvee:
> memiavl doesn't need much memory actually in our practice.

Will get back to you after some testing.  Having a hard time getting the right rpc/peers for evmos statesync.

chillyvee:
Tried to use memiavl to restore evmos.  Each time, the data directory is completely deleted and init is run with the chain-id evmos_9001-2.

After getting all statesync parts, the node crashes and upon start, this message appears
```
panic: invalid chain-id on BeginBlock; expected: , got: evmos_9001-2
```

Noticed that there is a fair amount of code changes and dependency on the cronos chain repository (github.com/crypto-org-chain/cronos/memiavl)  necessary to get enable memiavl for a new chain.  That might be more than a chain developer is able to do and maintain.  Another thought is to bring memiavl mainline into cosmos/iavl or cosmos/memiavl.  yihuang, you are most familiar so you can probably give a good comment on options or best way to implement.

This optimisitic export and import allows minimal code changes and faster restores.

yihuang:
> The faster iavl build time is confirmed (3+ hours. Same even with local snapshot files, no network dependency + restore on a fast machine with high ram + raid0 nvme)

what is the testing procedure, how large is the snapshot file, it's dozens of minutes for us, do you use the local state sync commands (`./simappd snapshots --help`)?

chillyvee:
Did not realize the chain-id was pulled from client.toml.  Fixed that and statesync with memiavl works.  Will let evmos run for a few more hours to generate snapshots for local restore testing.

Noticed a failed statesync snapshot.  Suspect that memiavl pruned the oldest height before the statesync snapshot could finish.  Future statesync snapshots were not generated.  Increased memiavl snapshots held to allow more time for statesync export.  

Do you think it would help to document the relation between these app.toml settings so that a configuration is practial?

```
pruning-keep-recent = "500000"

[memiavl]
  snapshot-interval = 1000
  snapshot-keep-recent = 1

[state-sync]
  snapshot-interval = 100000
  snapshot-keep-recent = 5
```

Statesync for evmos took about 1h40m to download and apply 2924 chunks over p2p.  Snapshot was old and took 4 hours to BlockSync to the latest block.

Locally disk reads are fast.  Reading all 2924 chunks takes only 1m30s.  

Will retry again when a local snapshot is successfully created for a more recent height.

yihuang:
> Noticed a failed statesync snapshot. Suspect that memiavl pruned the oldest height before the statesync snapshot could finish. Future statesync snapshots were not generated. Increased memiavl snapshots held to allow more time for statesync export.

do you have the error message?

> Do you think it would help to document the relation between these app.toml settings so that a configuration is practical?

we set `memiavl. snapshot-interval` to be `10000`, and `state-sync.snapshot-interval` to be the same as that when enabled.

yihuang:
> Noticed that there is a fair amount of code changes

I think most of the code changes are for the versiondb integration, memiavl itself is fairly simple.

> and dependency on the cronos chain repository (github.com/crypto-org-chain/cronos/memiavl) necessary to get enable memiavl for a new chain. That might be more than a chain developer is able to do and maintain. Another thought is to bring memiavl mainline into cosmos/iavl or cosmos/memiavl. yihuang, you are most familiar so you can probably give a good comment on options or best way to implement.

it's a good question on long term maintenance, we had a few discussions on storage workgroup calls, but the decision is to keep it as a third-party thing. personally I'm totally fine and supportive if we maintain it under cosmos org.

chillyvee:
> There could be different combinations, but for most of the use cases, I think `state-sync.snapshot-interval` can be simply set to the same as `memiavl.snapshot-interval` when enabled, so a state sync snapshot is created after each memiavl snapshot is created.
> We set `memiavl.snapshot-interval` to `10000`, it can even be larger if you can accept a little bit longer worst-case start-up time, `10000` translates to a few seconds in our case.
> I'd also recommend setting the `memiavl.snapshot-interval` much larger during catching-up, but that needs adjusting manually later on.
> The old `pruning-*` options have no effect on memiavl.

Sharing a problem encountered for snapshots using memiavl + evmosd on mainnet


ERR failed to create state snapshot err="failed to generate snapshot chunk 0: snapshot is not created yet: height: 19266752" height=192667

Looking at snapshots directory, initial restore height of 19072000 and the next snapshot is available.  The rest appear to fail
```
28G     19072000
28G     19080000
1.5K    19150000
1.5K    19160000
1.5K    19170000
1.5K    19180000
1.5K    19190000
1.5K    19217745
1.5K    19224746
1.5K    19231747
1.5K    19238748
1.5K    19245749
1.5K    19252750
1.5K    19259751
1.5K    19266752
```

data/memiavl.db has data as expected
```
memiavl.db
512     current
512     LOCK
171G    snapshot-00000000000019246541
171G    snapshot-00000000000019251544
171G    snapshot-00000000000019256547
171G    snapshot-00000000000019261550
171G    snapshot-00000000000019266553
171G    snapshot-00000000000019271556
```

evmosd snapshots list only shows the 2 successful data/snapshot heights
```
% evmosd snapshots list
height: 19080000 format: 3 chunks: 2923
height: 19072000 format: 3 chunks: 2924
```

This is the current app.toml settings
```
[memiavl]
  async-commit-buffer = 0
  cache-size = 1000
  enable = true
  snapshot-interval = 5003
  snapshot-keep-recent = 5
  zero-copy = false

[state-sync]
  snapshot-interval = 7001
  snapshot-keep-recent = 2
```

Are you able to see the same issue on your side?  Any alternate settings that you recommend checking?

yihuang:
> Sharing a problem encountered for snapshots using memiavl + evmosd on mainnet
> 
> ERR failed to create state snapshot err="failed to generate snapshot chunk 0: snapshot is not created yet: height: 19266752" height=192667
> 
> Looking at snapshots directory, initial restore height of 19072000 and the next snapshot is available. The rest appear to fail

The fifth parameter in [memiavlstore.SetupMemIAVL](https://github.com/evmos/evmos/blob/main/app/app.go#L343) is `supportExportNonSnapshotVersion bool`, where `false` means don't support taking a snapshot on non-snapshot versions.
The reason is there's a specialized snapshot algorithm when taking state-sync snapshots from the memiavl snapshots directly, which is more efficient than the generic algorithm that traverses the tree (although in practice, the difference is actually marginal), so you have too options here:
1. Set the `supportExportNonSnapshotVersion` to `true`, and it'll support exporting snapshots at any version.
2. Set `state-sync.snapshot-interval` to be multiples of `memiavl.snapshot-interval`, so when you take a state-sync snapshot, there's always a memiavl snapshot for it.

chillyvee:
> The fifth parameter in [memiavlstore.SetupMemIAVL](https://github.com/evmos/evmos/blob/main/app/app.go#L343)

Cosmos code has many unnamed variables.  This can make adding a configuration variable difficult (find if a parameter exists, determine if parameter can be passed, add parameter in calling, edit function call in called library)  

Is there any "guidance" against code structure such as:
```
       baseAppOptions = memiavlstore.SetupMemIAVL(common, options)
       baseAppOptions.setSupportExportNonSnapshotVersion(true)
       baseAppOptions.FinishSetup()
```
    

> personally I'm totally fine and supportive if we maintain it under cosmos org.

Would you prefer to start that yourself? Or could we prepare a repository and move that to cosmos org once it makes sense?


Do you think it is generally "safe" to add memiavl to any chain of any sdk version with this pattern of code?
```
	baseAppOptions = memiavlstore.SetupMemIAVL(logger, homePath, appOpts, false, false, baseAppOptions)

	// Setup Mempool and Proposal Handlers
	baseAppOptions = append(baseAppOptions, func(app *baseapp.BaseApp) { 
```

yihuang:
> Cosmos code has many unnamed variables. This can make adding a configuration variable difficult (find if a parameter exists, determine if parameter can be passed, add parameter in calling, edit function call in called library)

yeah, agree on this specific option, but it's usually static, there are several other options that are integrated into `app.toml`, we can maybe add this one to config file as well.

chillyvee:
> we have compatible versions for sdk 0.46 and 0.47, 0.50 is on draft PR.

Are the compatible version branches on https://github.com/crypto-org-chain/cronos/tree/main/memiavl ?

yihuang:
> > we have compatible versions for sdk 0.46 and 0.47, 0.50 is on draft PR.
> 
> Are the compatible version branches on https://github.com/crypto-org-chain/cronos/tree/main/memiavl ?

Yeah, currently, you need to follow cronos release branch, `v1.0.x` for SDK 0.46, `v1.1.x` for SDK 0.47.

I'm preparing a standalone repo at https://github.com/crypto-org-chain/memiavl.

chillyvee:
> 0.50 is on draft PR.

Please share when available.  Would like to run on a v0.50 testnet

chillyvee:
> `main` for SDK 0.47.

memiavl works well for gaiad in theta testnet

statesync node over p2p then put into validator mode. snapshots creation also works.  No issues so far.

chillyvee:
@yihuang Do you know if the statesync snapshot/restore works properly for memiavl + cosmwasm module chains?

chillyvee:
deleting data/memiavl.db/wal causes this error
```
panic: async wal writing goroutine quit unexpectedly: out of order

goroutine 36 [running]:
github.com/crypto-org-chain/cronos/store/rootmulti.(*Store).Commit(0xc0010d1040)
        github.com/crypto-org-chain/cronos/store@v0.0.5-0.20240301012710-e9bb006c8d14/rootmulti/store.go:120 +0x558
```

Is it possible to "rollback" to a memiavl snapshot and create a valid empty wal?

chillyvee:
> follow cronos release branch, v1.0.x for SDK 0.46,

Was not able to locate memiavl in: 
https://github.com/crypto-org-chain/cronos/tree/release/v1.0.x

Found this with a cosmos-sdk v0.46 reference.  Is this the right branch for v0.46?
https://github.com/crypto-org-chain/cronos/blob/memiavl/v0.0.5/memiavl/go.mod

yihuang:
> @yihuang Do you know if the statesync snapshot/restore works properly for memiavl + cosmwasm module chains?
> 
> 
> 
> 

I think so, it only replaces rms, the other part is still managed by sdk

yihuang:
> Was not able to locate memiavl in: https://github.com/crypto-org-chain/cronos/tree/release/v1.0.x
> 
> Found this with a cosmos-sdk v0.46 reference. Is this the right branch for v0.46? https://github.com/crypto-org-chain/cronos/blob/memiavl/v0.0.5/memiavl/go.mod

It's a bit hard to manage the release, it'll be easier when using a standalone repo, I suggest following the Cronos's `go.mod` itself, for example, right now it uses:

https://github.com/crypto-org-chain/cronos/blob/release/v1.0.x/go.mod#L222

> 
> Also is it difficult to modify for cosmos-sdk v0.45?

chillyvee:
> I think not difficult, since the store interfaces don't change much from 0.45 to 0.46.

Attempting to create a v0.45 branch for memiavl based on Jackal Chain

Here is the current progress code with current errors.  Rootmulti is not connecting properly yet.

https://github.com/chillyvee/cronos/tree/cv045wip

https://github.com/chillyvee/canine-chain/tree/cv3.1.3memiavl

With app.toml [state-sync] configured the error is:
```
failed to load latest version: state sync snapshots require a rootmulti store
```

With app.toml [state-sync] removed the error is:
```
panic: uncached KVStore is not supposed to be accessed directly
```

chillyvee:
wasm InitializedPinnedCodes calls cosmos-sdk which calls rootmulti GetKVStore
```
github.com/crypto-org-chain/cronos/store/rootmulti.(*Store).GetKVStore(0x0?, {0x0?, 0x0?})
        github.com/crypto-org-chain/cronos/store@v0.0.0-00010101000000-000000000000/rootmulti/store.go:231 +0x25
github.com/cosmos/cosmos-sdk/types.Context.KVStore({{0x2aa3b20, 0x3c0f940}, {0x7f419807b578, 0xc000dfcd00}, {{0x0, 0x0}, {0x0, 0x0}, 0x67ca50, {0x0, ...}, ...}, ...}, ...)
        github.com/cosmos/cosmos-sdk@v0.45.17/types/context.go:253 +0x92
github.com/CosmWasm/wasmd/x/wasm/keeper.Keeper.InitializePinnedCodes({{0x2a8a160, 0xc000ef85d0}, {0x2aba7e0, 0xc0002d4340}, {0x2a8e690, 0xc0005ff320}, {0x2a84b60, 0xc00037c020}, {0x2a842a0, 0xc00022d378}, ...}, ...)
        github.com/CosmWasm/wasmd@v0.32.0/x/wasm/keeper/keeper.go:890 +0x79
```

In this case cronos/store/rootmulti.(*Store).GetKVStore(0x0?, {0x0?, 0x0?}) immediately panics


Adding cosmwasm seems to require a way to directly get a named store for wasmd keeper to read and write to.

yihuang:
> uncached KVStore is not supposed to be accessed directly

You can try remove this check, it should be fine especially for read-only cases

chillyvee:
> > uncached KVStore is not supposed to be accessed directly
> 
> You can try remove this check, it should be fine especially for read-only cases

For statesync cosmwasm restore, write access is required.  Can you recommend a different object to return?

chillyvee:
> It's a bit hard to manage the release, it'll be easier when using a standalone repo, I suggest following the Cronos's `go.mod` itself, for example, right now it uses `v0.0.0-20240130060137-c12016e4052e`:

Attempted to use the recommended branch/commit for v0.46 target using gitopia chain without cosmwasm

Followed the replace to utilize memiavl within the same branch
https://github.com/crypto-org-chain/cronos/blob/release/memiavl/v0.0.x/store/go.mod#L135
```
	github.com/crypto-org-chain/cronos/memiavl => ../memiavl
```


This branch downgrades the commit above to match cosmos-sdk v0.46.13
https://github.com/chillyvee/cronos/tree/gitopia/cv0.46.13store

This branch of gitopia uses the branch above
https://github.com/chillyvee/gitopia/commit/3a4f8fe37b5fbfdca59ebc69f19e4d17e97775e1

On statesync restore, there is a **apphash mismatch error**

yihuang:
> > It's a bit hard to manage the release, it'll be easier when using a standalone repo, I suggest following the Cronos's `go.mod` itself, for example, right now it uses `v0.0.0-20240130060137-c12016e4052e`:
> 
> Attempted to use the recommended branch/commit for v0.46 target using gitopia chain without cosmwasm
> 
> Without memiavl the chain statesync restore is successful
> 
> Followed the replace to utilize memiavl within the same branch https://github.com/crypto-org-chain/cronos/blob/release/memiavl/v0.0.x/store/go.mod#L135
> 
> ```
> 	github.com/crypto-org-chain/cronos/memiavl => ../memiavl
> ```
> 
> This branch downgrades the commit above to match cosmos-sdk v0.46.13 https://github.com/chillyvee/cronos/tree/gitopia/cv0.46.13store
> 
> This branch of gitopia uses the branch above [chillyvee/gitopia@3a4f8fe](https://github.com/chillyvee/gitopia/commit/3a4f8fe37b5fbfdca59ebc69f19e4d17e97775e1)
> 
> On statesync restore, there is a **apphash mismatch error**

with sdk 0.46, the option `memiavlSdk46Compact` should be set to `true`, can you give that a try?

chillyvee:
> with sdk 0.46/0.45, the option `memiavlSdk46Compact` should be set to `true`, can you give that a try?

Success on statesync for jackal, memiavlSdk46Compact = true 
Success on statesync for gitopia, memiavlSdk46Compact = true 

> Just tagged a new version, please update to memiavl v0.0.6, store v0.0.6, (versiondb v0.0.6 if used).

Had some trouble during go mod tidy due to some cosmos-sdk dependencies.  Will feedback later after resolved.

Tagging with matching cosmos-sdk versions can be helpful for using the right version.  `memiavl/cosmos-sdk0.46.13` for example.   You can use the idea after memiavl is split into a dedicated repo.

yihuang:
> > with sdk 0.46/0.45, the option `memiavlSdk46Compact` should be set to `true`, can you give that a try?
> 
> Success on statesync for jackal, memiavlSdk46Compact = true Success on statesync for gitopia, memiavlSdk46Compact = true

great!

> 
> > Just tagged a new version, please update to memiavl v0.0.6, store v0.0.6, (versiondb v0.0.6 if used).
> 
> Had some trouble during go mod tidy due to some cosmos-sdk dependencies. Will feedback later after resolved.
> 
> Tagging with matching cosmos-sdk versions can be helpful for using the right version. `memiavl/cosmos-sdk0.46.13` for example. You can use the idea after memiavl is split into a dedicated repo.

you'll need a little bit change in the versiondb setup code, please check the patch here: https://github.com/crypto-org-chain/cronos/pull/1339

chillyvee:
> you'll need a little bit change in the versiondb

Thank you.  Will try again when there is time and share the result

PR #880, opened 2024-02-08
fix: pass initial version at immutable tree creation

beer-1:
In previous iavl, the version is not affecting to working hash. but in current iavl, the version is a part of working hash, so the `InitialVersion` should be used when you make working hash at newly added moules. 

https://github.com/cosmos/iavl/blob/af7ae13f47a477f11c18a4856d4a0c4e4a344e45/node.go#L435

The problem is happening when you adding an new module [`packetforward`](https://github.com/cosmos/ibc-apps/tree/main/middleware/packet-forward-middleware), you can easily reproduce the consensus breaking with software upgrade proposal. You should run more than two nodes and after upgrade restart validator node and see other nodes.

We deeply debugged it and found the working hash for the stores, which are added via upgrade, are same whatever the height you did upgrade (` [164 128 39 101 194 146 82 214 50 145 131 89 83 171 212 207 142 176 239 6 10 8 254 1 95 237 121 132 93 24 213 34]` for `packetforward`). but it should be differ in current iavl because it is using version to make working hash. and also found it is using `tree.version`, which is 0, when a new module firstly added by upgrade module. This wrong store hash is kept only in runtime. When you restart the node, it makes different `WorkingHash` and then consensus broken.

### DIFF
--- a/mutable_tree.go
+++ b/mutable_tree.go
@@ -56,7 +56,7 @@ func NewMutableTree(db dbm.DB, cacheSize int, skipFastStorageUpgrade bool, lg lo
 	}
 
 	ndb := newNodeDB(db, cacheSize, opts, lg)
-	head := &ImmutableTree{ndb: ndb, skipFastStorageUpgrade: skipFastStorageUpgrade, version: int64(opts.InitialVersion)}
+	head := &ImmutableTree{ndb: ndb, skipFastStorageUpgrade: skipFastStorageUpgrade}
 
 	return &MutableTree{
 		logger:                   lg,
@@ -460,6 +460,8 @@ func (tree *MutableTree) LoadVersion(targetVersion int64) (int64, error) {
 
 	if firstVersion == 0 {
 		if targetVersion <= 0 {
+			tree.version = int64(tree.ndb.opts.InitialVersion)
+
 			if !tree.skipFastStorageUpgrade {
 				tree.mtx.Lock()
 				defer tree.mtx.Unlock()
--- a/mutable_tree.go
+++ b/mutable_tree.go
@@ -56,7 +56,7 @@ func NewMutableTree(db dbm.DB, cacheSize int, skipFastStorageUpgrade bool, lg lo
 	}
 
 	ndb := newNodeDB(db, cacheSize, opts, lg)
-	head := &ImmutableTree{ndb: ndb, skipFastStorageUpgrade: skipFastStorageUpgrade}
+	head := &ImmutableTree{ndb: ndb, skipFastStorageUpgrade: skipFastStorageUpgrade, version: int64(opts.InitialVersion)}
 
 	return &MutableTree{
 		logger:                   lg,


beer-1:
If there is any side effect, then happy to move this to under `WorkingHash` function like

```go
// WorkingHash returns the hash of the current working tree.
func (tree *MutableTree) WorkingHash() []byte {
	_version := tree.ImmutableTree.version
	tree.ImmutableTree.version = tree.WorkingVersion()
	hash := tree.ImmutableTree.Hash()
	tree.ImmutableTree.version = _version

	return hash
}
```

alexanderbez (approved):

cool-develope:
Thanks for reporting this issue, it seems like the `InitialVersion` option is not used properly. `Tree.version` should be determined in `LoadVersion` so in this case, we should consider the `InitialVersion` while `LoadVersion`

beer-1:
> Thanks for reporting this issue, it seems like the `InitialVersion` option is not used properly. `Tree.version` should be determined in `LoadVersion` so in this case, we should consider the `InitialVersion` while `LoadVersion`.

If you like this more, I can fix the PR 

maybe result should be same but I think we can fix this by adding this line to `LoadVersion`
```diff
	if firstVersion == 0 {
		if targetVersion <= 0 {
			if !tree.skipFastStorageUpgrade {
				tree.mtx.Lock()
				defer tree.mtx.Unlock()
				_, err := tree.enableFastStorageAndCommitIfNotEnabled()
				return 0, err
			}

+			tree.version = int64(tree.ndb.opts.InitialVersion)
			return 0, nil
		}
		return 0, fmt.Errorf("no versions found while trying to load %v", targetVersion)
	}
```
https://github.com/cosmos/iavl/blob/af7ae13f47a477f11c18a4856d4a0c4e4a344e45/mutable_tree.go#L460-L475

mutable_tree.go
```diff
@@ -56,7 +56,7 @@ func NewMutableTree(db dbm.DB, cacheSize int, skipFastStorageUpgrade bool, lg lo
 	}
 
 	ndb := newNodeDB(db, cacheSize, opts, lg)
-	head := &ImmutableTree{ndb: ndb, skipFastStorageUpgrade: skipFastStorageUpgrade}
+	head := &ImmutableTree{ndb: ndb, skipFastStorageUpgrade: skipFastStorageUpgrade, version: int64(opts.InitialVersion)}
```
cool-develope, line 59: I know the result is same.
The happy case is to load the version while constructing the tree in `LoadVersion`, so we don't need to set the `InitialVersion` option. The `InitialVersion` is just optional considering the above case like new module.
But if we force set the version here, it can be misleading the meaning of `InitialVersion`.

beer-1: yea sure, I don't have strong opinion about this.
I will update PR after testing :)

beer-1: This fix is also working well 👍

beer-1: I don't have strong opinion on this, so happy to follow your way. 

Is the above solution (in comment) is good? or should we make other changes on `LoadVersion` function?

cool-develope: no, we should move this part to the `LoadVersion`, your commit will be not working when set the initial version by `SetInitialVersion` without opts.

beer-1: okay, I moved the fix to LoadVersion and tested it. It's also working well :)

cool-develope (approved):

PR #883, opened 2024-02-11
chore: docs proofreading

Pan-chao:
Came across few typos while reading the docs, hope I could be of any help!


tac0turtle (approved):
thank you

PR #889, opened 2024-02-25
perf: Parallelize saveNewNodes' DB writes with figuring out "what to write"

ValarDragon:
This PR parallelizes to SaveNodes, and figuring out what it is we have to write. ("saveNewNodes").

We can improve this in the future to process eveything but "Set" in another goroutine, and then keep a buffered queue for "Set" that completes asynchronously.

If this is not useful with the IAVL v2 work, can I just put this in the IAVL v1 line?  

This PR as is feels like a pretty straightforward improvement, that should give a 7% sync improvement on Osmosis for IAVL v1 today. I don't think theres any tests to add here, I don't see any edge case here thats not covered by existing tests.

Benchmark for 2000 blocks on IAVL v1 on osmosis mainnet for context: 
image

This PR will drop the latency of this from 42 seconds to 24 seconds. However

We should be able to with subsequent work:
- (No async commit) Parallelize latency for this function time to the longest of:
  - num nodes to hash * time to sha256 / num cores
  - DB writing all new nodes
- And async commit removes the DB writing part

Better parallelism would make it:
- No async commit be max(4 seconds, 18 seconds / num cores)
- Async commit (18 seconds / num cores)

### DIFF
--- a/mutable_tree.go
+++ b/mutable_tree.go
@@ -1028,51 +1028,55 @@ func (tree *MutableTree) balance(node *Node) (newSelf *Node, err error) {
 // saveNewNodes save new created nodes by the changes of the working tree.
 // NOTE: This function clears leftNode/rigthNode recursively and
 // calls _hash() on the given node.
+// TODO: Come back and figure out how to better parallelize this code.
 func (tree *MutableTree) saveNewNodes(version int64) error {
 	nonce := uint32(0)
-	newNodes := make([]*Node, 0)
-	var recursiveAssignKey func(*Node) ([]byte, error)
-	recursiveAssignKey = func(node *Node) ([]byte, error) {
+
+	nodeChan := make(chan *Node, 64) // Buffered channel to avoid blocking.
+	doneChan := make(chan error, 1)  // Channel to signal completion and return any errors.
+
+	// Start a goroutine to save nodes.
+	go func() {
+		var err error
+		for node := range nodeChan {
+			if saveErr := tree.ndb.SaveNode(node); saveErr != nil {
+				err = saveErr
+				break
+			}
+			node.leftNode, node.rightNode = nil, nil // Detach children after saving.
+		}
+		doneChan <- err // Send any error encountered or nil if none.
+	}()
+
+	var recursiveAssignKey func(*Node) []byte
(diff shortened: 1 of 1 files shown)


ValarDragon:
We've tested this gave a speedup on IAVL v1 on Osmosis!

mutable_tree.go
```diff
@@ -1028,53 +1028,57 @@ func (tree *MutableTree) balance(node *Node) (newSelf *Node, err error) {
...
 // calls _hash() on the given node.
+// TODO: Come back and figure out how to better parallelize this code.
 func (tree *MutableTree) saveNewNodes(version int64) error {
 	nonce := uint32(0)
-	newNodes := make([]*Node, 0)
-	var recursiveAssignKey func(*Node) ([]byte, error)
-	recursiveAssignKey = func(node *Node) ([]byte, error) {
+
+	nodeChan := make(chan *Node, 64) // Buffered channel to avoid blocking.
+	doneChan := make(chan error, 1)  // Channel to signal completion and return any errors.
```
kocubinski, line 1036: There is some overhead with creating channels and goroutine, although likely not so big, it might be better to couple channel and goroutine creation with the lifecycle of `MutableTree` instead of `saveNewNodes`.

Since I don't know off hand how much overhead that actually is maybe it's fine as-is too.

mutable_tree.go
```diff
@@ -1028,53 +1028,57 @@ func (tree *MutableTree) balance(node *Node) (newSelf *Node, err error) {
...
+				return node.nodeKey.GetKey()
 			}
-			return node.hash, nil
+			return node.hash
 		}
 		nonce++
 		node.nodeKey = &NodeKey{
 			version: version,
-			nonce:   nonce,
+			nonce:   nonce, // Example nonce calculation; adjust as needed.
```
cool-develope, line 1062: just realized the original codebase is wrong, not a bug but just a performance issue.
`newNodes = append(newNodes, node)` should come after this line, the main idea of `newNodes` slice is to save the node in a sorted manner by the nodekey (here the nonce) to reduce the compaction.

Just worried we can't assume this sorted manner after refactoring ...

PR #890, opened 2024-02-25
perf: Remove unneeded GetKey calls to the LRU cache

ValarDragon:
Remove some extra `GetKey` calls within the LRU cache code. GetKey calls took 1.5s on an Osmosis 2000 block long block sync. This PR should shave off .45s of that.

Though ideally we get the LRU code rethought / adapted more, to make it generally lower overhead. (Its currently 13.5 s on a 2000 block sync!)


kocubinski (approved):

cool-develope (approved):

tac0turtle:

odeke-em:
@ValarDragon this isn't fully the correct optimization despite eliminating the extra call to node.Key() while then also keep the original intended optimization or zero byteslice->string conversion in the case of already existing keys when looking im the map. In short it now it incurs the byteslice->string conversion in all cases, but could have been faster if it had been:

```go
key := node.GetKey()
if value, ok := c.dict[string(key)]; ok {
   ...
}

c.dict[string(key)] = value
```

cache/cache.go
```diff
@@ -96,8 +97,9 @@ func (c *lruCache) Len() int {
 }
 
 func (c *lruCache) Remove(key []byte) Node {
-	if elem, exists := c.dict[string(key)]; exists {
-		return c.remove(elem)
+	keyS := string(key)
+	if elem, exists := c.dict[keyS]; exists {
```
odeke-em, line 101: This undoes the zero byteslice->string Go map lookup optimization.

ValarDragon: Oh wow, didn't know about this compiler optimization.

Issue #885, opened 2024-02-15
Lazy Pruning


cool-develope:
## Background

Currently, the pruning process in IAVL is synchronous and blocks the `Commit` flow, leading to performance bottlenecks, especially when pruning legacy orphaned nodes in v1.

## Proposal

Implement `lazy pruning` in IAVL v1 to make the process asynchronous and improve performance. Lazy pruning will separate the pruning operation from the main `SaveVersion` flow, allowing it to run independently and not block the `Commit` flow.

## Plan

- Run the internal pruning loop process.
- Store the version in the `Prune` function to let the internal pruning process know about the up-to-date version. It allows to not break the current interface.
- Setup the channel between `SaveVersion` and the internal pruning process to allow blocking the pruning process in `SaveVersion`.

alexanderbez:
Nice! To be clear for those reading, by "lazy" we mean deferred execution.

ValarDragon:
Would really appreciate us doing this!

tac0turtle:
completed

PR #898, opened 2024-03-05
feat: lazy pruning

cool-develope:
## Implement the deferred pruning

Closes: #885

## Fix

When the root refers to the legacy root, it will reformat the legacy node as a new node key format. This node is removed while pruning but it is still being referred by the forward version.
Another issue with the reference node, it can refer to the internal node, not only the root. For example, if there are no adds, just removes, then the leaf or inner node can be the root.

- Store the `nodeKey` instead of only the version as a reference node value
- When reformat the legacy root, use (version, 0) as a `nodeKey` to distinguish with the root. It will also be applied for the case of when pruning the version, reformat the given root nodeKey to (version, 0).
- Fix the legacy node pruning

### DIFF
--- a/.mergify.yaml
+++ b/.mergify.yaml
@@ -4,26 +4,6 @@ queue_rules:
       - "#approved-reviews-by>1"
 
 pull_request_rules:
-  - name: automerge to main with label automerge and branch protection passing
-    conditions:
-      - "#approved-reviews-by>1"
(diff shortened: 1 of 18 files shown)


cool-develope:
The lazy pruning feature is not working properly since multi trees share the same db backend in store

close in favor of #909

PR #900, opened 2024-03-05
chore: fix mergify

tac0turtle:


.mergify.yaml
```diff
@@ -3,27 +3,6 @@ queue_rules:
     conditions:
       - "#approved-reviews-by>1"
 
-pull_request_rules:
```
julienrbrt, line 6: Shouldn't this header stay?

julienrbrt (approved):
ACK,  but I think the pull_request_rules are now queue_rules.

PR #902, opened 2024-03-07
chore: cleanup unnecessary `ndb.Commit` calls

cool-develope:
## Context

There are some unnecessary `ndb.Commit` and `resetBatch` calls which introduced to avoid the large batch commit originally. They are not necessary since we are using `BatchWithFlusher` now.

### DIFF
--- a/mutable_tree.go
+++ b/mutable_tree.go
@@ -15,9 +15,6 @@ import (
 )
 
 var (
-	// commitGap after upgrade/delete commitGap FastNodes when commit the batch
-	commitGap uint64 = 5000000
-
 	// ErrVersionDoesNotExist is returned if a requested version does not exist.
 	ErrVersionDoesNotExist = errors.New("version does not exist")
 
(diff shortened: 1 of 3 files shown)


ValarDragon (approved):
Awesome! Just need to get cosmos-db sorted out for batch with flusher on osmosis

tac0turtle (approved):

tac0turtle:
backport?

cool-develope:
> backport?

not sure, @czarcas7ic osmosis implemented the db wrapper with flusher fully?
I think we can merge it, and later backport when need

czarcas7ic:
@cool-develope I don't think we have batchWithFlusher implemented

czarcas7ic:
Thanks for implementing this for our sdk fork @cool-develope! Seems we are unblocked to merge so we can test

czarcas7ic:
Sorry maybe I am not fully understanding how implementing with flusher has to do with this PR?

In the PR you made @cool-develope, it seems like NewBatch and NewBatchWithSize still behave the same, which I thought that was what needed to be addressed https://github.com/osmosis-labs/cosmos-sdk/blob/6c1422877673277448bc14e2be415133cc6738e6/store/wrapper/tmdb.go#L52-L57

My apologies if my understanding is off here.

cool-develope:
> Sorry maybe I am not fully understanding how implementing with flusher has to do with this PR?
> 
> In the PR you made @cool-develope, it seems like NewBatch and NewBatchWithSize still behave the same, which I thought that was what needed to be addressed https://github.com/osmosis-labs/cosmos-sdk/blob/6c1422877673277448bc14e2be415133cc6738e6/store/wrapper/tmdb.go#L52-L57
> 
> My apologies if my understanding is off here.

`NewBatchSize` is aimed to allocate the memory when create the batch, it will allow to avoid memory grow in batch insertion. It is not directly related to the flusher. The purpose of flusher is to avoid the large batch writing, it writes periodically if the batch size is greater than the threshold. 
Ofc, `NewBatchSize` will increase the performance, unfortunately I have no idea of how to refactor the prefixdb of comet-db. However, we can merge it without `NewBatchSize` still.

czarcas7ic:
Thanks for the explanation!

cool-develope:
@Mergifyio backport release/v1.1.x

PR #903, opened 2024-03-07
fix: revert the children node key refer

cool-develope:
## Context

We cannot assume the `leftNodeKey` and `rigthNodeKey` is finalized at the hash calculation stage, they will be re-assigned in `saveNewNodes`. IAVL V1 doesn’t ensure leftNodeKey and rightNodeKey updates when rotating, only update leftNode and rightNode relations in-memory, and children nodeKey will be resolved at the commit stage.


tac0turtle (approved):
could you make a pr to v1.1.x as well

czarcas7ic (approved):

PR #906, opened 2024-03-08
chore: backport lazy pruning to v1.x.x

czarcas7ic:
Should be merged to v1.1.x as well.

### DIFF
--- a/nodedb.go
+++ b/nodedb.go
@@ -82,8 +82,11 @@ type nodeDB struct {
 	firstVersion        int64            // First version of nodeDB.
 	latestVersion       int64            // Latest version of nodeDB.
 	legacyLatestVersion int64            // Latest version of nodeDB in legacy format.
+	pruneVersion        int64            // Version to prune up to.
 	nodeCache           cache.Cache      // Cache for nodes in the regular tree that consists of key-value pairs at any version.
 	fastNodeCache       cache.Cache      // Cache for nodes in the fast index that represents only key-value pairs at the latest version.
+	isCommitting        bool             // Flag to indicate that the nodeDB is committing.
+	chCommitting        chan struct{}    // Channel to signal that the committing is done.
 }
 
 func newNodeDB(db dbm.DB, cacheSize int, opts Options, lg log.Logger) *nodeDB {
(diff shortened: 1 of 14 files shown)


nodedb.go
```diff
@@ -1036,6 +1178,24 @@ func (ndb *nodeDB) nodes() ([]*Node, error) {
 	return nodes, nil
 }
 
+func (ndb *nodeDB) legacyNodes() ([]*Node, error) {
+	nodes := []*Node{}
+
+	err := ndb.traversePrefix(legacyNodeKeyFormat.Prefix(), func(key, value []byte) error {
```
czarcas7ic, line 1184: This was the only part I had questions on. The PR to main has legacyNodeKeyFormat as a KeyFormat type instead of a FastPrefixFormatter type. Key() in FastPrefixFormatter requires a byte array input. I think Prefix() is what we should use here but would like confirmation.

cool-develope: we can use both, Key() = Prefix()

cool-develope (approved):
LGTM

czarcas7ic:
@cool-develope test failure looks concerning though, having trouble running locally as well, looking into it

czarcas7ic:
@cool-develope I see you reran and it worked, is this non determinism expected / existed prior to this PR?

cool-develope:
> @cool-develope I see you reran and it worked, is this non determinism expected / existed prior to this PR?

yes, that one is non-deterministic which is related to random sets, I will fix it later

czarcas7ic:
Great, I will test this on Osmosis then and report back

nodedb.go
```diff
@@ -385,14 +457,56 @@ func (ndb *nodeDB) deleteVersion(version int64) error {
...
 		return err
 	}
-	if rootKey == nil || GetNodeKey(rootKey).version < version {
-		if err := ndb.batch.Delete(ndb.nodeKey(GetRootKey(version))); err != nil {
+
+	literalRootKey := GetRootKey(version)
+	if rootKey == nil || !bytes.Equal(rootKey, literalRootKey) {
+		// if the root key is not matched with the literal root key, it means the given root
+		// is a reference root to the previous version.
+		if err := ndb.deleteFromPruning(ndb.nodeKey(literalRootKey)); err != nil {
```
roy-dydx, line 465: This use of batch deletes is unsafe. For example, we let's say we batch delete the root key for the target version. If the main routine saves a tree, it will write the batch to disk. Then `traverseOrphans` will fail because it assumes that the target version is still on disk

czarcas7ic:
Closing in favor of https://github.com/cosmos/iavl/pull/911

PR #913, opened 2024-03-13
perf: fix accidentally undone Go map[string(byteslice)] lookup optimization

odeke-em:
This change brings back an optimization that go innocently removed while trying to reduce the number of invocations to node.GetKey() in PR #890.

The Go compiler considers the special case of lookups per:

    value, ok := map[string(byteslice)]
or
    value := map[string(byteslice)]

and performs a zero byteslice->string conversion because it knows that there is no storing in the map hence a read-only usage of the byteslice's memory.

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -61,8 +61,8 @@ func New(maxElementCount int) Cache {
 }
 
 func (c *lruCache) Add(node Node) Node {
-	key := string(node.GetKey())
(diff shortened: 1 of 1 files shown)


ValarDragon (approved):
Thank you, didn't know about this optimization!

likhita-809 (approved):

tac0turtle (approved):

PR #917, opened 2024-03-15
perf(encoding): speed up EncodeVarint with io.ByteWriter+hand rolled varintEncode

### LINKED ISSUE #891: Change IAVL node.writeBytes to take in bytes.Buffer
Every usage of node.WriteBytes we use takes in an io.Writer, not a bytes.Buffer.

We currently are spending 2.1 seconds out of a 2000 block sync on getting and returning buffer pools in node.WriteBytes. (Due to PutVarint calls) I think we should be able to remove this time entirely, by making PutVarint methods that just call bytes.Buffer.WriteByte directly, rather than using a slice from the sync.Pool and then using the generic io.Writer(WriteBytes) which is a second memmove.

odeke-em:
This change speeds up EncodeVarint by testing if the input writer implements io.ByteWriter and if so, goes to use our hand-rolled varint encoder, instead of using the awkward standard libary encoding/binary.PutVarint that requires a byteslice, which we also retrofitted using a bytearray pool.
While here, added parity tests to ensure that we get the exact same results as with the Go standard library's encoding/binary package with caution from https://cyber.orijtech.com/advisory/varint-decode-limitless and also added benchmarks whose results reflect the change in just the benchmark initially

```shell
name            old time/op    new time/op    delta
EncodeVarint-8     360ns ± 3%     245ns ± 3%  -31.80%  (p=0.000 n=10+10)

name            old alloc/op   new alloc/op   delta
EncodeVarint-8     0.00B          0.00B          ~     (all equal)

name            old allocs/op  new allocs/op  delta
EncodeVarint-8      0.00           0.00          ~     (all equal)
```

Fixes #891

### DIFF
--- a/internal/encoding/bench_test.go
+++ b/internal/encoding/bench_test.go
@@ -1,0 +1,59 @@
+package encoding
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"math"
+	"testing"
+)
+
+var encValues = []int64{
+	-1, -100, -1 << 32,
+	0, 1, 100, 1 << 32,
+	-1 << 52, 1 << 52, 17,
+	19, 28, 37, 388888888,
+	-99999999999, 99999999999,
+	math.MaxInt64, math.MinInt64,
+}
+
+// This tests that the results from directly invoking binary.PutVarint match
+// exactly those that we get from invoking EncodeVarint and its internals.
+func TestEncodeVarintParity(t *testing.T) {
+	buf := new(bytes.Buffer)
+	var board [binary.MaxVarintLen64]byte
+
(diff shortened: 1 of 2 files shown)


odeke-em:
Kindly cc-ing @ValarDragon

ValarDragon (approved):
Ah amazing, nice job!

tac0turtle (approved):

tac0turtle:
@mergifyio backport release/v1.x

tac0turtle:
@mergifyio backport release/v1.1.x

tac0turtle:

odeke-em:
Thanks for the reviews Dev & Marko!

PR #925, opened 2024-04-04
feat: deferred pruning

cool-develope:
Closes: #885 

Added the `AsyncPruning` option, it needs to call `SetCommitting` and `UnsetCommitting` when committing in the store module in the async pruning mode.
It doesn't break the original APIs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### Improvements
+
+- [#925](https://github.com/cosmos/iavl/pull/925) Add the `AsyncPruning` option to the `MutableTree` constructor to enable async pruning.
+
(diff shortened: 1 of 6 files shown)


czarcas7ic:
I plan on testing this either this weekend, or this coming week. Thanks so much for making this!

czarcas7ic (approved):
Worked with @cool-develope and tested against Osmosis mainnet, no issues thus far!

tac0turtle:
before we tag id like to test on dydx. Ill aim to do it this week, but we can go ahead with the merge with one more approval

sontrinh16 (approved):

tac0turtle:
does this need to be backported?

PR #928, opened 2024-04-05
fix: legacy node key

cool-develope:
## Context

After reformatting of the pruned root node, it may be referred as a legacy node, since the nonce is 0.


czarcas7ic (approved):
Tested and it fixes our app hash issue

tac0turtle (commented):
can you add a test case please

tac0turtle:
tests pass locally, git actions are acting weird

PR #943, opened 2024-05-07
fix: working hash with initial version

cool-develope:
## Context

`InitialVersion` is not reflected in the `WorkingHash` calculation, only reflected in the `SaveVersion`. This will lead to the `AppHash` mismatch while upgrading, since the the hash of the root is already calculated in the `WorkingHash` and won't re-calculated in the `SaveVersion` which is wrong.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## v1.0.4 May 7, 2024
+
+### Bug Fixes
+
+- [#943](https://github.com/cosmos/iavl/pull/943) Fix the `WorkingHash` with the `InitialVersion` option.
+
 ## v1.0.3 April 8, 2024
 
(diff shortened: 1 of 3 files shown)


tac0turtle (approved):

yihuang:
This PR seems to be a major breaking change in how root hash is calculated with the initial version.
ref: https://github.com/cosmos/iavl/pull/660

yihuang:
> > This PR seems to be a major breaking change in how root hash is calculated with the initial version. ref: #660
> 
> This is modified in #676 , yes v1 is a consensus breaking in the edge case, [#956 (comment)](https://github.com/cosmos/iavl/issues/956#issuecomment-2192854803)

I see, what's interesting is we only notice the difference after the `WorkingHash` is fixed.

PR #952, opened 2024-05-24
feat: expose nodeDB's DeleteVersionsFrom method

pirtleshell:
`DeleteVersionsFrom` is the counterpart to `DeleteVersionsTo`. It writes deletions of a `MutableTree` to disk for all versions >= the provided `fromVersion`.

It exposes the method of the same name from the underlying `nodeDB`.

This method is useful for things like:
* rolling back an upgrade that adds a module (allows for the complete removal of a KVStore from disk)
* custom implementations of sharded data drives (allows one to pare down a data drive to a specific set of versions)

Please let me know if this should first be proposed as a Feature Request via the repo issues.

## Summary by CodeRabbit

- **New Features**
  - Introduced the ability to delete versions starting from a specified version with the new `DeleteVersionsFrom` API.
  
- **Tests**
  - Added new tests to verify the functionality of the `DeleteVersionsFrom` API.
  - Renamed existing test function `TestDelete` to `TestDeleteVersionsTo` for clarity.


pirtleshell:
@cool-develope sure thing.

## rollback case
First, suppose you are adding a new sdk module during an upgrade. Suppose there is a bug in that module addition such that the root hash of the new module is non-deterministic & results in an app hash mismatch when the upgrade is run in a network. The goal is to roll back the bad version & re-run the upgrade with a patched binary.

The SDK's `rollback` cmd calls `LoadVersionForOverwriting`, which works by loading the previous version of the data. If an upgrade handler adds a module via `StoreUpgrades`, the rollback of the upgrade block will panic with an error that looks like:
```
Error: failed to rollback to version: version does not exist
```
This occurs when `LoadVersionForOverwriting` is called on the newly added module's store with the previous height, which does not exist because the module was just added.

AFAIK, to get around this and successfully roll back, you have two options:
* use the previous binary to rollback, it doesn't know about the new module so won't attempt to load it
* manually de-register the module from the root multistore so `LoadVersionForOverwriting` is never called for it

However, either of those options still leaves the upgrade version of the added module's store on disk.

When the upgrade is re-attemtped, the new data will not be committed because of the version that already exists on disk. The relevant sdk code is here:
https://github.com/cosmos/cosmos-sdk/blob/c4d9a495052b78a02723ed1df1d336efe83a4e8d/store/rootmulti/store.go#L1188-L1197

The code is necessary to recover from interruptions, but means that new data is never committed & written to disk for the patched binary's upgrade block. If it were to manage calling commit, the IAVL Commit() method would likely error [here](https://github.com/cosmos/iavl/blob/master/mutable_tree.go#L696-L702) due to mismatched data (the patched binary's version will differ from the bad one that was committed in the failed upgrade).

To truly rollback & re-run the upgrade, the data must be removed from disk.

## sharding case
Less important with the awesome space-saving y'all have been doing with iavl v1 & v2, but `DeleteVersionsFrom` could also be used to create dbs for shards; ie. one server has blocks 1 to 2.5M, another has blocks 2.5M to 5M. You could create these dbs from a full archive node using `DeleteVersionsTo` and `DeleteVersionsFrom`

---

I recognize that much of what I describe about `rollback` could/should be accounted for higher up in the sdk (I've opened an issue: https://github.com/cosmos/cosmos-sdk/issues/20472). Exposing `DeleteVersionsFrom` is a just a simple non-breaking change that gives developers more control over the disk's data. It helped me resolve an app hash mismatch by rolling back an upgrade that added modules, so I figured it might help others as well, and is worth adding to this package.

Happy to make any changes, explain anything further, or open an issue/discussion.

pirtleshell:
> @pirtleshell do you think we also need to update cosmos-sdk for the upgrade rollback case?

yes, i wanted to get this into iavl first but i do have a rough implementation of an update to the rollback command that fully removes the KVStore on disk for listed modules. i linked it in the issue i opened on https://github.com/cosmos/cosmos-sdk/issues/20472

probably needs cleanup but it is here: https://github.com/Kava-Labs/cosmos-sdk/pull/546
currently that code uses a forked iavl v1.2 that has the patch from this PR applied.

thinking through it right now, the cosmos-sdk versions that use iavl v0.20 can & probably should use `DeleteVersionRange` to accomplish the same.

pirtleshell:
@cool-develope anything else needed for this PR? i think i may have confused the CI by rebasing on master to fix CHANGELOG conflict. 🙂

pirtleshell:
bump @cool-develope @tac0turtle
could someone trigger the Test CI so this can be auto-merged?
thanks!

pirtleshell:
bump @cool-develope @tac0turtle 

is anything further needed for this to be included?

tac0turtle (approved):

tac0turtle:
@pirtleshell could merge main into your branch then we can merge it

pirtleshell:
bumping again, hoping to get this merged @tac0turtle @cool-develope 🙏

PR #954, opened 2024-05-28
refactor: use cosmossdk.io/core/log 

tac0turtle:
This pr uses cosmossdk.io/core/log instead of cosmossdk.io/log to avoid the import of zero log and implementation specific items. 

This is a proposal for dep cleanup

## Summary by CodeRabbit

- **Chores**
  - Updated logging import paths across various files for better organization.
  - Replaced `log.NewTestLogger(t)` with `log.NewNopLogger()` in `migrate_test.go` for logging consistency.


go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/cosmos/iavl
 go 1.21
 
 require (
-	cosmossdk.io/log v1.3.1
+	cosmossdk.io/core v0.12.1-0.20240514205955-97c9bbb0341b
```
kocubinski, line 6: This does enshrine protobuf a bit more (dependency of core) https://github.com/cosmos/cosmos-sdk/blob/dac569c3a2a5277cec1b5ecba978074ab45def80/core/go.mod#L8-L9.
But it looks like that's already a part of iavl anyway.

kocubinski (approved):
LGTM, question about protubuf dep from core.

go.mod
tac0turtle: was thinking the same when I was working on it and tried to remove it but seems like a large battle for not much benefit

PR #955, opened 2024-05-28
refactor: remove cosmos-db as a dep

tac0turtle:


tac0turtle:
> overall, lgtm! small concern of RocksDB users won't be compatible

what do you mean? since db is passed from above would it not work out of the box?

tac0turtle:
> > > overall, lgtm! small concern of RocksDB users won't be compatible
> > 
> > 
> > what do you mean? since db is passed from above would it not work out of the box?
> 
> no, I'd like to add the small godoc or description of how to use `wrapper` for pebbledb or rocksdb users

in the cmd go.mod right?

PR #963, opened 2024-07-16
refactor: remove fastnode system

tac0turtle:
this pr is a proposal to remove the fast node system. 

we will run benchmarks on main and this branch to test the differences 

v1 works better without the fast node system due to OS and db caching


tac0turtle:
after talking with osmosis this is needed

PR #965, opened 2024-07-25
refactor: use expected logger interface in iavl 1.2

julienrbrt:
Replicates https://github.com/cosmos/iavl/pull/964, but for IAVL v1.2.x


tac0turtle (approved):

julienrbrt:
> can we include #962 in the next release? I am also preparing one small PR for #966 , that would be great if we can include that.

Okay, I won't tag afterwards, I'll just use a commit in the SDK then 👍🏾

PR #983, opened 2024-08-19
fix: batch write race on commit

corverroos:
Wait for previous batch before writing batch on `importer.Commit`

### DIFF
--- a/import.go
+++ b/import.go
@@ -213,7 +213,17 @@ func (i *Importer) Commit() error {
 			len(i.stack))
 	}
 
-	err := i.batch.WriteSync()
+	// Wait for previous batch.
+	var err error
+	if i.inflightCommit != nil {
+		err = <-i.inflightCommit
+		i.inflightCommit = nil
+	}
+	if err != nil {
+		return err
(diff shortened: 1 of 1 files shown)


corverroos:
For more context, see this [issue](https://github.com/omni-network/omni/issues/1524) on omni that resulted in this PR.

import.go
```diff
@@ -213,7 +213,17 @@ func (i *Importer) Commit() error {
 			len(i.stack))
 	}
 
-	err := i.batch.WriteSync()
+	// Wait for previous batch.
+	var err error
+	if i.inflightCommit != nil {
+		err = <-i.inflightCommit
```
corverroos, line 219: maybe using a `sync.Mutex` around batch.Write(Sync)` and `batch.Commit` would be easier to reason about?

kocubinski (approved):
nice catch, lgtm

corverroos:
Hi gents, is there anything more I need to do from my side?

tac0turtle:
@mergifyio backport master

PR #1002, opened 2024-10-31
feat: allow saving version 0

kocubinski:
This patch makes `SetInitialVersion(0)` calls valid, which previously would error during `SaveVersion`.  This is particularly useful for the server/v2 which processes and saves a Genesis block at version `initialHeight -1`, which is 0 is the default case.


tac0turtle:
this needs to be backported to v1.3

tac0turtle:
@mergifyio backport release/v1.3.x

PR #1006, opened 2024-11-19
bug: root node get pruned when it's still referenced

yihuang:
Wrote a unit test to reproduce.

thanks @mmsqe 

the problem is [here](https://github.com/cosmos/iavl/blob/master/nodedb.go#L456), we can't simply delete the old root node because it might be referenced by other versions down the road, but I guess we also can't leave it there, because it might create a gap in the versions, which violates the assumption in the design, right? @cool-develope

### DIFF
--- a/mutable_tree_test.go
+++ b/mutable_tree_test.go
@@ -1494,3 +1494,27 @@ func TestMutableTree_InitialVersionZero(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, int64(0), version)
 }
+
+func TestReferenceRootPruning(t *testing.T) {
+	memDB := dbm.NewMemDB()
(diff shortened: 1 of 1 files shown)


yihuang:
> > Wrote a unit test to reproduce.
> > thanks @mmsqe
> > the problem is [here](https://github.com/cosmos/iavl/blob/master/nodedb.go#L456), we can't simply delete the old root node because it might be referenced by other versions down the road, but I guess we also can't leave it there, because it might create a gap in the versions, which violates the assumption in the design, right? @cool-develope
> 
> thanks for this report, fixed in https://github.com/cosmos/iavl/pull/1007/files#diff-6df49839605c94e5caeb7d4549afe697ad97a00144ef26ec6746d9b5e8a3019cR162-R175

great, closing this one.

PR #1023, opened 2024-12-17
fix: Better lock handling on close

### LINKED ISSUE #1017: Intermittent deadlock when Closing a tree
I am frequently encountering this in tests involving multiple SDK apps in the same process, on iavl tag v1.3.2. I am working with the latest cosmos-sdk commit, [where async pruning is fixed to true](https://github.com/cosmos/cosmos-sdk/blob/256ec4f90277cadc0f976e89e10d3100cc820e50/store/v2/commitment/iavl/tree.go#L28).

The relevant code snippets are:

https://github.com/cosmos/iavl/blob/d89d5d22030e8fea42ce5406743ce45b75ad86ea/nodedb.go#L1122-L1129

and

https://github.com/cosmos/iavl/blob/d89d5d22030e8fea42ce5406743ce45b75ad86ea/nodedb.go#L599-L608

`(*nodeDB).startPruning` runs in its own goroutine, created during `newNodeDB`. `(*nodeDB).Close` is called on a separate goroutine, e.g. from closing an SDK commitment store. Flow during the deadlock happens as follows:

1. The `Close` goroutine acquires the lock on `ndb.mtx`
2. Concurrently, the `startPruning` goroutine enters the default case and attempts to call `ndb.mtx.Lock()`, but it cannot acquire the lock until the `Close` goroutine releases it
3. Therefore, the `Close` goroutine is blocked reading from `ndb.done` because the `startPruning` goroutine cannot advance past acquiring the lock

alpe:
Resolves #1017

I have tested this via SDK sims.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## v1.3.4 unreleased 
+## v1.3.4, December 17, 2024 
 
 - [#1023](https://github.com/cosmos/iavl/pull/1023) Better lock handling on close
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## v1.3.4 unreleased 
+
+- [#1023](https://github.com/cosmos/iavl/pull/1023) Better lock handling on close
+
+
 ## v1.3.3, December 16, 2024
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## v1.3.4 unreleased
```
julienrbrt, line 3: Can we make this tagging ready?

tac0turtle (approved):

julienrbrt (approved):

PR #1048, opened 2025-02-21
Handle deleting dangling refence root node

iboss-ptk:
We found a case in Osmosis node where there is a root key that is points to a node that doesn't exists and it hangs the pruning process because fails at get root key (returns `ErrVersionDoesNotExist`).

There is already a code to clean the dangling ref node up, but it just never get there because it early returns `ErrVersionDoesNotExist` before getting there.


ValarDragon (approved):

aljo242:
@iboss-ptk we have some conflict - would be good to resolve so we can merge

PR #1050, opened 2025-02-23
fix: importer to wait db close

beer-1:
Here is context: https://github.com/cosmos/cosmos-sdk/issues/23740

When we use statesync, it is using this IAVL importer but sometimes the IAVL db is empty even there is no notable errors. And after digging the issue, I found import is not waiting the `batch.Close()`.

This PR is to fix to wait importer to wait previous batch `WriteSync()` and `Close()`.


tac0turtle (approved):

aljo242 (approved):

aljo242 (changes requested):
@beer-1 can we get a test for this case?

beer-1:
Hey @aljo242 
I found `(goleveldb)Batch.Write()` is already handling this issue, by calling `.Commit()` [here](https://github.com/cosmos/cosmos-db/blob/58ab43bc265b9985c8fd6ff78e1a8be1dfd4f4dc/goleveldb_batch.go#L75).

so this was not the cause, but it only happened in `v1.2.x` problem
this commit should be back ported to `v1.2.x` https://github.com/cosmos/iavl/commit/34895bf30d2a74f54a64f58af0e17ac5ac9fc893

aljo242:
@beer-1 lgtm when we fix the lints (just can add //nolint comment)

aljo242 (approved):

PR #1067, opened 2025-03-27
fix: get latest version in store after legacy pruning fails

PaddyMc:
### Description

Currently if legacy pruning fails it sets the `firstVersion` to `latestLegacyVersion + 1`

That works under the assumption that if legacy pruning fails that the non-legacy node is the first version is available in the store, this is not the case, and the code will then linearly iterate until it finds the next version, which could be a long wait.

### Context
please read https://github.com/cosmos/iavl/pull/1063 for context on this issue

### What this function does

- A binary search to find the latest non-legacy version in the store
- This is a monkey patch and should only run only once when a node start pruning then never again, then the cached first version should be kept per node pruning session 

### What should actually happen

- Pruning should succeed for legacy node
- We should still search for the latest none-legacy node

## Summary by CodeRabbit

- **New Features**
  - Enhanced the process for cleaning up outdated versions, ensuring the system retains only valid data for improved reliability.

- **Tests**
  - Added comprehensive tests that validate the improved version cleanup under various scenarios.

### DIFF
--- a/nodedb.go
+++ b/nodedb.go
@@ -724,7 +724,15 @@ func (ndb *nodeDB) deleteVersionsTo(toVersion int64) error {
 		if err := ndb.deleteLegacyVersions(legacyLatestVersion); err != nil {
 			ndb.logger.Error("Error deleting legacy versions", "err", err)
 		}
-		first = legacyLatestVersion + 1
+		// NOTE: When pruning is broken for legacy versions we need to find the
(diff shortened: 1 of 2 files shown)


aljo242:
@Mergify backport release/v1.2.x

aljo242:

aljo242:
@Mergifyio backport release/v1.3.x

PR #1090, opened 2025-04-29
fix: properly store errors in Iterator.Next() method

GarmashAlex:
This commit fixes the error handling in the Iterator.Next() method 
by explicitly storing errors from traversal.next() in the iterator's 
error field. Previously, errors were silently dropped, causing them to 
be unavailable through the Error() method.

The fix:
1. Separates the condition checking for errors and nil nodes
2. Properly stores any returned error in iter.err
3. Maintains the same behavior of marking the iterator as invalid 
   in both error and end-of-traversal cases

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling during iteration to better capture and report errors separately from normal completion.  
  - Enhanced stability by ensuring iteration stops and reports errors when encountering missing data during traversal.

### DIFF
--- a/iterator_test.go
+++ b/iterator_test.go
@@ -1,7 +1,6 @@
 package iavl
 
 import (
-	"errors"
 	"math/rand"
 	"sort"
 	"sync"
@@ -381,13 +380,6 @@ func syncMapCount(m *sync.Map) int {
 		return true
 	})
 	return count
-}
-
-// mockNodeDB always returns an error for GetNode
-type mockNodeDB struct{}
-
-func (m *mockNodeDB) GetNode(_ []byte) (*Node, error) {
-	return nil, errors.New("mock nodeDB error")
 }
 
 func TestIterator_Next_ErrorHandling(t *testing.T) {


yihuang (approved):

iterator.go
```diff
@@ -228,8 +228,15 @@ func (iter *Iterator) Next() {
...
-	if node == nil || err != nil {
+	// If an error occurred or no more nodes, update iterator state accordingly
+	if err != nil {
+		iter.err = err
+		iter.t = nil
+		iter.valid = false
+		return
+	}
+
+	if node == nil {
```
yihuang, line 239:
```suggestion
	// TODO: double-check if this error is correctly handled.
	if node == nil || err != nil {
	    if err != nil {
	        iter.err = err
	    }
```

The change is equivalent to this.

yihuang: thanks for taking the suggestion, but it seems messed up with the indent, 4 spaces vs 1 tab.

yihuang (approved):
Can you fix the code format, I think my code suggestion messed up the spaces and tabs.

yihuang (approved):

aljo242:
This looks good @GarmashAlex could you add a test to verify that this is working?

aljo242:

aljo242:
@Mergifyio backport release/v1.3.x

aljo242 (approved):

PR #1107, opened 2025-07-07
fix: nodedb find partial latest version in edge cases

yihuang:
`BatchWithFlusher` could separate the commit of the same version into multiple batches, when the batches are not persisted atomically, the current latest version method will return an incomplete version and eventually return error in `MutableTree.Load(0)` method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Bug Fixes
 
 - [#1007](https://github.com/cosmos/iavl/pull/1007) Add the extra check for the reformatted root node in `GetNode`
+- [#]() nodedb latest version should not set to a partial committed version.
 
 ### Improvements
 
--- a/batch_flusher_test.go
+++ b/batch_flusher_test.go
@@ -1,0 +1,85 @@
+package iavl
+
+import (
+	"fmt"
+	"math/rand"
+	"testing"
+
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Bug Fixes
 
 - [#1007](https://github.com/cosmos/iavl/pull/1007) Add the extra check for the reformatted root node in `GetNode`
+- [#]() nodedb latest version should not set to a partial committed version.
```
yihuang, line 8:
```suggestion
- [#1107](https://github.com/cosmos/iavl/pull/1107) nodedb latest version should not set to a partial committed version.
```

batch_flusher_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"fmt"
+	"math/rand"
+	"testing"
+
+	dbm "github.com/cosmos/iavl/db"
+	"github.com/stretchr/testify/require"
+)
+
+type MockDBBatch struct {
+	dbm.Batch
```
mmsqe, line 13: seems they removed `dbm.Batch` in https://github.com/cosmos/iavl/pull/955/files#diff-367b87ccd372ceca8c6f2e7157b827b4d469318a6571c0cacc99d54c1bb19089

yihuang: changed to corestore

PR #1159, opened 2026-06-19
fix(keyformat): validate KeyBytes segment count

### LINKED ISSUE #760: keyformat: (*KeyFormat).KeyBytes presumes that .prefix will always be non-empty
This code https://github.com/cosmos/iavl/blob/d0241db89d211a17e9dbd249587cba257d36a4f9/keyformat/key_format.go#L70 assumes that .prefix will always have a length > 0. I found this issue by code auditing but also by fuzzing

```go
package keyformat

import (
	"encoding/json"
	"fmt"
	"strings"
	"testing"
)

func FuzzKeyFormatKeyBytes(f *testing.F) {
	if testing.Short() {
		f.Skip("in -short mode")
	}

	// 1. Create some seeds.
	seeds := []*KeyFormat{
		NewKeyFormat('a', 1, 2, 3, 5, 6, 0),
		NewKeyFormat('b', 1, 2, 3, 5, 6, 1),
		NewKeyFormat('b', 9, 21),
		NewKeyFormat(byte('e'), 8, 8, 8),
		NewKeyFormat(byte('e'), 8, 0),
	}

	type envelope struct {
		KF *KeyFormat `json:"kf"`
		BS [][]byte   `json:"bs"`
	}

	for _, kf := range seeds {
		bsL := [][][]byte{
			[][]byte{[]byte(""), []byte("abcdefgh")},
			[][]byte{{1, 2, 3, 4, 5, 6, 7, 8}, {1, 2, 3, 4, 5, 6, 7, 8, 9}},
			[][]byte{{1, 2, 3, 4, 5, 6, 7, 8}, []byte("hellohello")},
			[][]byte{{1, 2, 3, 4, 5, 6, 7, 8}, {1, 2, 3, 4, 5, 6, 7, 8}, {1, 1, 2, 2, 3, 3}},
		}
		for _, bs := range bsL {
			blob, err := json.Marshal(&envelope{KF: kf, BS: bs})
			if err != nil {
				f.Fatal(err)
			}
			f.Add(blob)
		}
	}

	// 2. Fuzz it now.
	f.Fuzz(func(t *testing.T, inputJSON []byte) {
		defer func() {
			if r := recover(); r != nil {
				str := fmt.Sprintf("%s", r)
				if !strings.Contains(str, "provided to KeyFormat.KeyBytes() is longer than the") {
					// Re-raise the panic as it panicked in an unexpected place.
					panic(r)
				}
			}
		}()

		env := new(envelope)
		if err := json.Unmarshal(inputJSON, env); err != nil {
			return
		}

		kf := env.KF
		if kf == nil {
			return
		}
		_ = kf.KeyBytes(env.BS...)
	})
}
```

/cc @elias-orijtech

haoshengzhen:
## Summary

- Validate that `KeyFormat.KeyBytes` is not called with more segments than the key format layout supports.
- Return a clear `KeyFormat.KeyBytes` panic instead of a runtime index-out-of-range panic.
- Add a regression test for the zero-layout case.
- Add an Unreleased changelog entry.

## Motivation

`KeyFormat.Key` already checks for too many arguments before formatting. `KeyBytes` did not have the equivalent guard, so calling it with more byte segments than the layout length could panic with a low-level bounds error.

Fixes #760.

## Testing

- `go test ./keyformat -count=1`
- `make legacydump && go test ./... -count=1`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Bug Fixes
 
+- Return a clear `KeyFormat.KeyBytes` panic when too many segments are provided.
+
 ### Improvements
 
 ## v1.3.6, April 6, 2026
--- a/keyformat/key_format.go
+++ b/keyformat/key_format.go
@@ -52,6 +52,11 @@ func NewKeyFormat(prefix byte, layout ...int) *KeyFormat {
 
 // Format the byte segments into the key format - will panic if the segment lengths do not match the layout.
 func (kf *KeyFormat) KeyBytes(segments ...[]byte) []byte {
+	if len(segments) > len(kf.layout) {
+		panic(fmt.Errorf("keyFormat.KeyBytes() is provided with %d segments but format only has %d segments",
+			len(segments), len(kf.layout)))
+	}
+
 	keyLen := kf.length
 	// In case segments length is less than layouts length,
 	// we don't have to allocate the whole kf.length, just
--- a/keyformat/key_format_test.go
+++ b/keyformat/key_format_test.go
@@ -62,6 +62,18 @@ func TestKeyFormatBytes(t *testing.T) {
 	}
 }
 
+func TestKeyFormatBytesPanicsWithTooManySegments(t *testing.T) {
+	kf := NewKeyFormat(byte('e'))
+
+	assert.PanicsWithError(
+		t,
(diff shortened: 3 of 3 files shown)


haoshengzhen:
@mattac21 @vladjdk @Eric-Warehime Hi, Could you please review this PR at your convenience? Thank you very much.

